PR #8, opened 2018-03-05
@unbox Pandas.Series -> StringArray

fschlimb:
@ehsantn

### DIFF
--- a/hpat/_str_ext.cpp
+++ b/hpat/_str_ext.cpp
@@ -10,10 +10,10 @@ using std::regex_search;
 // #include <boost/regex.hpp>
 // using boost::regex;
 // using boost::regex_search;
-
+extern "C" {
 struct str_arr_payload {
     int64_t size;
-    int32_t *offsets;
+    uint32_t *offsets;
     char* data;
 };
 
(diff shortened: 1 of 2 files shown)


hpat/_str_ext.cpp
```diff
@@ -311,6 +316,68 @@ void* str_from_float64(double in)
...
+
+    uint32_t * offsets = new uint32_t[n];
+    std::vector<const char *> tmp_store(n);
+    size_t len = 0;
+    for(Py_ssize_t i = 0; i < n; ++i) {
+        offsets[i] = len;
+        PyObject * s = PySequence_GetItem(obj, i);
+        CHECK(s, "getting element failed");
+        CHECK(PyString_Check(s), "expecting a string");
+        tmp_store[i] = PyString_AsString(s);
```
ehsantn: @fschlimb I think we need to deallocate the strings returned from `PyUnicode_AsUTF8`

https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize

ehsantn:
@fschlimb could you add a unittest for this?

hpat/_str_ext.cpp
fschlimb: the doc says: "The caller is not responsible for deallocating the buffer."
We release the reference to the PyObject in 363.

ehsantn: Sorry, I misread the page :)

fschlimb:
a meaningful unittest is not that straight-forward as we need a PyObject and it returns a C-StringArray.

I can provide a small python example which tests it, though.

ehsantn:
Let's add unittests with the next PR. merging.

PR #11, opened 2018-03-27
Fix compiler warnings; remove a redundant comment.

agayev:


ehsantn:
The quantile change seems to cause issue. That file needs more work.

PR #12, opened 2018-03-27
Support np.random.permutation with int argument.

agayev:
We currently replicate the result of the call to np.random.permutation with
integer argument.  Ideally, this should be parallelized, so that each processor
generates a chunk of the permutation array.  But this patch is mostly an exercise on
how to call a C++ extension, and we leave the true parallel implementation of
permutation for future work.

### DIFF
--- a/hpat/_distributed.cpp
+++ b/hpat/_distributed.cpp
@@ -44,6 +44,9 @@ void oneD_reshape_shuffle(char* output,
                           int64_t old_0dim_global_len,
                           int64_t out_lower_dims_size,
                           int64_t in_lower_dims_size);
+
+void permutation_int(int64_t* output, int n);
+
 int hpat_finalize();
 int hpat_dummy_ptr[64];
 void* hpat_get_dummy_ptr() {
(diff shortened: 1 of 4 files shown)


hpat/distributed_analysis.py
```diff
@@ -306,6 +306,17 @@ def _analyze_call(self, lhs, func_var, args, array_dists):
             self._analyze_call_np_concatenate(lhs, args, array_dists)
             return
 
+        if call_list == ['permutation', 'random', np]:
+            call_def = guard(get_definition, self.func_ir, func_var)
+            if isinstance(call_def, ir.Expr) and call_def.op == 'getattr':
```
ehsantn: call definition check is not required since the full call name is checked in call_list.

hpat/distributed_analysis.py
```diff
@@ -306,6 +306,17 @@ def _analyze_call(self, lhs, func_var, args, array_dists):
             self._analyze_call_np_concatenate(lhs, args, array_dists)
             return
 
+        if call_list == ['permutation', 'random', np]:
```
ehsantn: This piece of code shouldn't be part of this patch because shuffling of arrays with permutation is not implemented here.

hpat/_distributed.cpp
```diff
@@ -455,6 +461,16 @@ int hpat_finalize()
     return 0;
 }
 
+void permutation_int(int64_t* output, int n)
+{
+     if (hpat_dist_get_rank() == 0) {
```
ehsantn: There is no need to generate permutation in C because `output` already has a permutation array from Python.

PR #15, opened 2018-03-29
Support np.random.permutation with integer argument.

agayev:
For now, we generate a permutation index in one processor and broadcast it to
all.  Our assumption is that the index is much smaller than the data items we
will shuffle, therefore, the cost of broadcasting an array will be negligible
compared to data shuffling.  We will support fully parallel permutation in the
future.


agayev:
Do not merge this yet.  It's not working properly.

ehsantn:
ok.

agayev:
You can merge this.

PR #17, opened 2018-04-02
 Support array shuffling with an index array generated with np.random.permutation

agayev:
Currently, it assumes the array that is being indexed is a one-dimensional array of ints and appears to work.  Please see if this makes sense.  If it does, then I will extend it to support multi-dimensional arrays.

### DIFF
--- a/hpat/_distributed.cpp
+++ b/hpat/_distributed.cpp
@@ -577,6 +577,10 @@ void permutation_array_index(unsigned char *lhs, int64_t len, int64_t elem_size,
         return;
     }
 
+    MPI_Datatype element_t;
+    MPI_Type_contiguous(elem_size, MPI_UNSIGNED_CHAR, &element_t);
+    MPI_Type_commit(&element_t);
+
     auto num_ranks = hpat_dist_get_size();
     auto rank = hpat_dist_get_rank();
     auto dest_ranks = find_dest_ranks(rank, num_ranks, p, p_len);
@@ -595,13 +599,15 @@ void permutation_array_index(unsigned char *lhs, int64_t len, int64_t elem_size,
     }
 
     MPI_Alltoallv(send_buf.data(), send_counts.data(), send_disps.data(),
-                  MPI_INT64_T, lhs, recv_counts.data(), recv_disps.data(),
-                  MPI_INT64_T, MPI_COMM_WORLD);
+                  element_t, lhs, recv_counts.data(), recv_disps.data(),
+                  element_t, MPI_COMM_WORLD);
 
     auto begin = p + hpat_dist_get_start(p_len, num_ranks, rank);
     auto p1 = arg_sort(begin, dest_ranks.size());
     auto p2 = arg_sort(p1.data(), dest_ranks.size());
     apply_permutation(lhs, elem_size, p2);
+
+    MPI_Type_free(&element_t);
 }
 
 void oneD_reshape_shuffle(char* output,
--- a/hpat/_distributed.cpp
+++ b/hpat/_distributed.cpp
@@ -526,6 +526,8 @@ std::vector<int> find_recv_counts(int64_t rank, int64_t num_ranks,
     return recv_counts;
 }
 
+// Returns an |index_array| which would sort the array |v| of size |len| when
+// applied to it.  Identical to numpy.argsort.
 template<class T>
 std::vector<size_t> arg_sort(T *v, int64_t len) {
     std::vector<size_t> index_array(len);
@@ -535,6 +537,8 @@ std::vector<size_t> arg_sort(T *v, int64_t len) {
     return index_array;
 }
 
+// |v| is an array of elements of size |elem_size|.  This function swaps
+// elements located at indices |i1| and |i2|.
 void elem_swap(unsigned char *v, int64_t elem_size, size_t i1, size_t i2)
 {
     std::vector<unsigned char> tmp(elem_size);
@@ -545,6 +549,8 @@ void elem_swap(unsigned char *v, int64_t elem_size, size_t i1, size_t i2)
     std::copy(std::begin(tmp), std::end(tmp), i2_offset);
 }
 
+// Applies the permutation represented by |p| to the array |v| whose elements
+// are of size |elem_size|.
 void apply_permutation(unsigned char *v, int64_t elem_size,
                        std::vector<size_t>& p)
 {
@@ -561,6 +567,8 @@ void apply_permutation(unsigned char *v, int64_t elem_size,
     }
 }
 
+// Applies the permutation represented by |p| of size |p_len| to the array |rhs|
+// of elements of size |elem_size| and stores the result in |lhs|.
 void permutation_array_index(unsigned char *lhs, int64_t len, int64_t elem_size,
                              unsigned char *rhs, int64_t *p, int64_t p_len)
 {
--- a/hpat/_distributed.cpp
+++ b/hpat/_distributed.cpp
@@ -481,6 +481,11 @@ void permutation_int(int64_t* output, int n)
      MPI_Bcast(output, n, MPI_INT64_T, 0, MPI_COMM_WORLD);
 }
 
+// Given the permutation index |idx| and |rank|, and the number of ranks
+// |num_ranks|, finds the destination ranks of indices of the |rank|.  For
+// example, if |rank| is 1, |num_ranks| is 3, |idx_len| is 12 and |idx| is the
+// following array [ 9, 8, 6, 4, 11, 7, 2, 3, 5, 0, 1, 10], the function returns
+// [2, 0, 1, 0].
 std::vector<int> find_dest_ranks(int rank, int num_ranks,
                                  int64_t *idx, int idx_len)
 {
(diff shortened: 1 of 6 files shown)


hpat/distributed_analysis.py
```diff
@@ -518,6 +518,14 @@ def _analyze_getitem(self, inst, lhs, rhs, array_dists):
                                                 new_dist.value))
             return
 
+        # array selection with permutation array index
+        if is_np_array(self.typemap, index_var.name):
+            arr_def = guard(get_definition, self.func_ir, index_var)
+            if (arr_def.op == 'call' and
+                    self._call_table[arr_def.func.name] == ['permutation', 'random', np]):
+                self._meet_array_dists(lhs, rhs.value.name, array_dists)
```
ehsantn: `arr_def` could be None or any node (not necessarily Expr). This `if` needs more checks otherwise it fails.

ehsantn: We need to check `rhs.value` to be multi-dimensional.

ehsantn: call_table is gone now. need to use find_callname.

hpat/distributed.py
```diff
@@ -1099,25 +1124,33 @@ def f(A, val, start, stop, chunk_start, chunk_count):  # pragma: no cover
...
+            lhs = full_node.target
+
             # we only consider 1st dimension for multi-dim arrays
             if index_var.name in self._tuple_table:
                 inds = self._tuple_table[index_var.name]
                 index_var = inds[0]
                 is_multi_dim = True
 
+            arr_def = guard(get_definition, self.func_ir, index_var)
+            if (arr_def.op == 'call' and
```
ehsantn: same issue with `arr_def` check before.

hpat/distributed_analysis.py
```diff
@@ -231,7 +230,7 @@ def _analyze_call(self, lhs, rhs, func_var, args, array_dists):
             self._analyze_call_hpat_dist(lhs, func_name, args, array_dists)
             return
 
-        if fdef == ('permutation', 'np.random'):
+        if fdef == ('permutation', 'numpy.random'):
             assert len(args) == 1
             assert self.typemap[args[0].name] == types.int64
```
ehsantn: this needs removed.

hpat/distributed.py
```diff
@@ -676,6 +677,30 @@ def f(lhs, n):
         f_block.body = [assign] + f_block.body
         return f_block.body[:-3]
 
+    def _run_permutation_array_index(self, lhs, rhs, idx):
+        scope, loc = lhs.scope, lhs.loc
+        alloc = mk_alloc(self.typemap, self.calltypes, lhs,
+                         tuple(self._array_sizes[lhs.name]),
```
ehsantn: chunk size of first dimention: self._array_counts[lhs.name][0].

agayev:
Fixed the chunk size.  You were wrong about the code path not being exercised.  It was being exercised, except when you execute your shuffle algorithm with one node, you end up sending all of the elements to yourself in sorted order.  The missing thing was shuffle after MPI_Alltoallv.  I added a hack to receive make up a PRNG and shuffle with that, but the proper way is to obtain the seed from numba and use that.

agayev:
Dead code eliminator removes np.random.seed() call, which makes it hard to write test for this.  Fixed it in #2873.

hpat/_distributed.cpp
```diff
@@ -470,6 +481,80 @@ void permutation_int(int64_t* output, int n)
      MPI_Bcast(output, n, MPI_INT64_T, 0, MPI_COMM_WORLD);
 }
 
+std::vector<int> find_dest_ranks(int rank, int num_ranks,
+                                 int64_t *idx, int idx_len)
```
ehsantn: needs doc.

hpat/_distributed.cpp
```diff
@@ -470,6 +481,80 @@ void permutation_int(int64_t* output, int n)
...
+    std::vector<int64_t> send_buf(dest_ranks.size());
+    for (size_t i = 0; i < dest_ranks.size(); ++i)
+        send_buf[offsets[dest_ranks[i]]++] = irhs[i];
+
+    MPI_Alltoallv(send_buf.data(), send_counts.data(), send_disps.data(),
+                  MPI_INT64_T, ilhs, recv_counts.data(), recv_disps.data(),
+                  MPI_INT64_T, MPI_COMM_WORLD);
+
+    std::mt19937 rng(0);
+    std::shuffle(ilhs, ilhs + dest_ranks.size(), rng);
```
ehsantn: shuffle? idx should determine the order.

hpat/tests/test_basic.py
```diff
@@ -264,5 +264,16 @@ def test_impl(n):
         self.assertEqual(count_array_REPs(), 0)
         self.assertEqual(count_parfor_REPs(), 0)
 
+    def test_permuted_array_indexing(self):
+        def test_impl(n):
+            A = np.arange(n)
+            B = np.random.permutation(n)
+            A = A[B]
+            return A.sum()
+
```
ehsantn: Needs a test where two arrays are permuted with the same index array, to make sure they are permuted in the same way.

hpat/_distributed.cpp
```diff
@@ -470,6 +481,80 @@ void permutation_int(int64_t* output, int n)
...
+    auto recv_disps = find_disps(recv_counts);
+
+    auto offsets = send_disps;
+    std::vector<int64_t> send_buf(dest_ranks.size());
+    for (size_t i = 0; i < dest_ranks.size(); ++i)
+        send_buf[offsets[dest_ranks[i]]++] = irhs[i];
+
+    MPI_Alltoallv(send_buf.data(), send_counts.data(), send_disps.data(),
+                  MPI_INT64_T, ilhs, recv_counts.data(), recv_disps.data(),
+                  MPI_INT64_T, MPI_COMM_WORLD);
```
ehsantn: Needs support for non-int types.

agayev:
The last patch addresses all the issues, except the unit tests.  You can review this, while I work on the unit tests.

hpat/_distributed.cpp
```diff
@@ -470,6 +480,130 @@ void permutation_int(int64_t* output, int n)
      MPI_Bcast(output, n, MPI_INT64_T, 0, MPI_COMM_WORLD);
 }
 
+// Given the permutation index |p| and |rank|, and the number of ranks
+// |num_ranks|, finds the destination ranks of indices of the |rank|.  For
+// example, if |rank| is 1, |num_ranks| is 3, |p_len| is 12, and |p| is the
+// following array [ 9, 8, 6, 4, 11, 7, 2, 3, 5, 0, 1, 10], the function returns
+// [2, 0, 1, 0].
```
ehsantn: Shouldn't this be `[0,2,0,1]`?

hpat/_distributed.cpp
```diff
@@ -470,6 +480,130 @@ void permutation_int(int64_t* output, int n)
...
+    std::copy(i2_offset, i2_offset + elem_size, i1_offset);
+    std::copy(std::begin(tmp), std::end(tmp), i2_offset);
+}
+
+// Applies the permutation represented by |p| to the array |v| whose elements
+// are of size |elem_size|.
+void apply_permutation(unsigned char *v, int64_t elem_size,
+                       std::vector<size_t>& p)
+{
+    std::vector<unsigned char> tmp(elem_size);
```
ehsantn: tmp not used.

hpat/_distributed.cpp
```diff
@@ -470,6 +480,130 @@ void permutation_int(int64_t* output, int n)
...
+    for (size_t i = 0; i < p.size(); ++i) {
+        auto current = i;
+        while (i != p[current]) {
```
ehsantn: This loop needs doc/reference.

hpat/_distributed.cpp
```diff
@@ -470,6 +480,130 @@ void permutation_int(int64_t* output, int n)
...
+    std::vector<unsigned char> send_buf(dest_ranks.size() * elem_size);
+    for (size_t i = 0; i < dest_ranks.size(); ++i) {
+        auto send_buf_offset = offsets[dest_ranks[i]]++ * elem_size;
+        auto *send_buf_begin = send_buf.data() + send_buf_offset;
+        auto *rhs_begin = rhs + i * elem_size;
+        std::copy(rhs_begin, rhs_begin + elem_size, send_buf_begin);
+    }
+
+    MPI_Alltoallv(send_buf.data(), send_counts.data(), send_disps.data(),
+                  MPI_INT64_T, lhs, recv_counts.data(), recv_disps.data(),
```
ehsantn: Needs MPI data type of size `elem_size`.

hpat/_distributed.cpp
```diff
@@ -470,6 +480,130 @@ void permutation_int(int64_t* output, int n)
...
+                  MPI_INT64_T, MPI_COMM_WORLD);
+
+    auto begin = p + hpat_dist_get_start(p_len, num_ranks, rank);
+    auto p1 = arg_sort(begin, dest_ranks.size());
+    auto p2 = arg_sort(p1.data(), dest_ranks.size());
```
ehsantn: Needs explanation.

hpat/_distributed.cpp
agayev: Correct.  Fixed.

hpat/_distributed.cpp
agayev: Removed.

hpat/_distributed.cpp
agayev: Added.

hpat/_distributed.cpp
agayev: Fixed.

hpat/_distributed.cpp
agayev: Added.

ehsantn:
Looks good. Merging.

PR #18, opened 2018-04-06
Fix comments and a memory leak.

agayev:
I reviewed the comment I wrote in a rush before I left and found many grammatical mistakes and a nonsensical sentence, which this patch fixes.  While doing that, noticed a flurry of delete[]s at the end of the file and thought you may have missed some.


ehsantn:
Makse sense. Thanks.

PR #26, opened 2018-04-30
Building with Intel(R) distribution for python and boost regex

fschlimb:
- simplifying building parquet extension 
- fixing arguments to setuptools extension specs
- adding unit test for regex
- using boost regex

### DIFF
--- a/buildscripts/hpat-conda-recipe/meta.yaml
+++ b/buildscripts/hpat-conda-recipe/meta.yaml
@@ -21,12 +21,10 @@ requirements:
     - pandas
     - mpich # [not win]
     - pyarrow 0.9.*
-    - boost
+    - boost-cpp
     - hdf5
     - h5py
-    - daal-devel
     - opencv 3.3.1
-    - hpat-parquet-reader # [not win]
 
   run:
     - python 3.6.*
@@ -34,11 +32,10 @@ requirements:
     - pandas
     - mpich # [not win]
     - pyarrow 0.9.*
-    - boost
+    - boost-cpp
     - numba 0.38.*
     - hdf5
     - h5py
-    - hpat-parquet-reader # [not win]
 
 test:
   imports:
--- a/setup.py
+++ b/setup.py
@@ -76,16 +76,16 @@ if 'HPAT_XE_SUPPORT' in os.environ and  os.environ['HPAT_XE_SUPPORT'] != "0":
 
 ind = [PREFIX_DIR+'/include',]
 lid = [PREFIX_DIR+'/lib',]
-eca = ['-std=c++11', '-DUSE_BOOST_REGEX',]
+eca = ['-std=c++11',]  # '-g', '-O0']
 ela = ['-std=c++11',]
 
 MPI_LIBS = ['mpi']
-H5_COMPILE_FLAGS = []
+H5_CPP_FLAGS = []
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -135,71 +150,73 @@ def readme():
...
+                         library_dirs = lid,
+)
 
-pq_libs = MPI_LIBS + ['boost_filesystem']
 
-if is_win:
-    pq_libs += ['arrow', 'parquet']
-else:
-    # seperate parquet reader used due to ABI incompatibility of arrow
-    pq_libs += ['hpat_parquet_reader']
```
ehsantn: If we remove hpat_parquet_reader as a separate package, we can't build HPAT with default Anaconda environment anymore. Can we have both default Anaconda and Intel builds? maybe if macros?

setup.py
```diff
@@ -209,7 +226,7 @@ def readme():
     _ext_mods.append(ext_xenon_wrapper)
 
 setup(name='hpat',
-      version='0.2.0',
+      version='0.3.0',
```
ehsantn: We should figure out our versioning policy.

buildscripts/hpat-conda-recipe/meta.yaml
```diff
@@ -21,24 +21,21 @@ requirements:
     - pandas
     - mpich # [not win]
     - pyarrow 0.9.*
-    - boost
+    - boost-cpp
     - hdf5
     - h5py
-    - daal-devel
     - opencv 3.3.1
-    - hpat-parquet-reader # [not win]
```
ehsantn: We need hpat-parquet-reader for building with default Anaconda.

buildscripts/hpat-conda-recipe/meta.yaml
```diff
@@ -21,24 +21,21 @@ requirements:
     - pandas
     - mpich # [not win]
     - pyarrow 0.9.*
-    - boost
+    - boost-cpp
```
ehsantn: what is the difference of boost and boost-cpp?

setup.py
fschlimb: What happens with anaconda?

buildscripts/hpat-conda-recipe/meta.yaml
fschlimb: Why?

setup.py
ehsantn: Anaconda is not compatible with conda-forge (we use parquet/arrow) because of compiler version differences (gcc 7.2 vs 4.8). hpat_parquet_reader provides a C interface (without std::string) to avoid the issue.

buildscripts/hpat-conda-recipe/meta.yaml
fschlimb: boost is a meta package from anconda, which depends on libboost and boost-py (or so). The conda-forge naming is different. The arrow packages we use from conda-forge use the ones from conda-forge. anaconda is behind with the version, we cannot mix.

setup.py
fschlimb: I guess the bigger problem is boost, which we use from conda-forge.

ehsantn: It's easier to go around boost problems.

ehsantn:
Closing in favor of #27.

PR #30, opened 2018-07-05
added dynamic home using conda command

petacube:
instead of always explicitly specifying prefix


shssf:
@fschlimb Do you think we need to drop this PR or slightly change it and accept?

fschlimb:
I see no issue with this PR. I am fine with merging.

PR #39, opened 2018-08-14
Fs/daal4py

fschlimb:
Adding "generic" support for daal4py. 
Using built-in specification from daal4py to generate necessary typing and lowering code.
Supports virtually all arguments from daal4py and distinguishes between optional and required args.


hpat/distributed_analysis.py
```diff
@@ -298,10 +299,51 @@ def _analyze_call(self, lhs, rhs, func_var, args, array_dists):
...
+                    # handle all input arguments and set their distribution as given by the spec
+                    for i in range(len(args)):
+                        aname = args[i].name
+                        adist = algo.spec.input_types[i][2]
+                        if aname not in array_dists:
+                            array_dists[aname] = adist
+                        else:
+                            # bail out if there is a distribution conflict with some other use of the argument
+                            # FIXME: handle Distribution.Thread and Disribution.REP as equivalent
+                            assert array_dists[aname] == adist,\
```
ehsantn: I think a better way than this assertion here is probably "(meet of array_dists[aname] and adist) == adist". This enables converting argument distribution to a lower one without problem. For example, if an argument is 1D but adist is REP, the argument can be converted to REP.

hpat/ml/d4p.py
```diff
@@ -0,0 +1,390 @@
...
+                           input_types = [(x[0], self.all_nbtypes[x[1].rstrip('*')], d4p_dtypes[x[2]]) for x in spec['input_types']])
+        elif 'attrs' in spec:
+            # filter out (do not support) properties for which we do not know the numba type
+            attrs = []
+            for x in spec['attrs']:
+                typ = x[1].rstrip('*')
+                if typ in self.all_nbtypes:
+                    attrs.append((x[0], self.all_nbtypes[typ]))
+                else:
+                    print("Warning: couldn't find numba type for '" + x[1] +"'. Ignored.")
```
ehsantn: `warnings.warn` is better than print.

ehsantn:
@fschlimb Could you make daal4py dependency optional for now?

hpat/distributed_analysis.py
fschlimb: Not sure what this would imply. Right now, it assumes the data is OneD and act accordingly. 
I don't know which of the algorithms are sensitive to data duplication and which are not.
It requires explicit configuration to switch to proper REP processing in a distributed setup. Doable, but should probably be postponed.

fschlimb:
done.

PR #40, opened 2018-08-22
supporting proper dlls

fschlimb:


ehsantn:
Build error ([link](https://travis-ci.org/IntelLabs/hpat/jobs/419213895)):
`hpat/_distributed.cpp:12:10: fatal error: _distributed.h: No such file or directory`

ehsantn:
This can cut down a lot of boilerplate code, but I'm a bit concerned about the reliability of this approach. Let's see if we can get it working on Linux/Mac/Windows.

fschlimb:
The problem here is that the test script is executed in the checkout directory. When you import hpat, it actually imports the "source" files from '.' and not the ones from  PREFIX/lib/site-packages/hpat. 

I made the build script install a "develop" build which should make this work for now, but the test-script needs fixing. It's not testing an actual install but rather a weird mix of local files and installed libs.

fschlimb:
No idea why mac can't resolve _hpat_barrier (and why it prepends '_' in the first place) and why the issue does not occur with other symbols.

fschlimb:
We will copy shared code to header files and accept the code-size over head. For now.

PR #44, opened 2018-09-07
report complex128 result type for Series corr and cov methods

Quasilyte:
For complex input arguments, return complex type.
For all other types, return float64.

Added test cases that were previously unsupported due to typecheck errors from numba.
There are precision problems for some other input data, but this is unrelated issue and should be
addressed separately (if at all, I don't know the requirements).

### DIFF
--- a/hpat/pd_series_ext.py
+++ b/hpat/pd_series_ext.py
@@ -417,23 +417,22 @@ class SeriesAttribute(AttributeTemplate):
         ret_typ = if_arr_to_series_type(ret_typ)
         return signature(ret_typ, *args)
 
-    @bound_function("series.cov")
-    def resolve_cov(self, ary, args, kws):
+    def _resolve_cov_func(self, ary, args, kws):
         # array is valid since hiframes_typed calls this after type replacement
         assert len(args) == 1 and isinstance(args[0], (SeriesType, types.Array))
         assert isinstance(ary.dtype, types.Number)
         assert isinstance(args[0].dtype, types.Number)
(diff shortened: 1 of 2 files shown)


Quasilyte:
СС @TocarIP.

hpat/pd_series_ext.py
```diff
@@ -423,17 +423,17 @@ def resolve_cov(self, ary, args, kws):
         assert len(args) == 1 and isinstance(args[0], (SeriesType, types.Array))
         assert isinstance(ary.dtype, types.Number)
         assert isinstance(args[0].dtype, types.Number)
-        # TODO: complex numbers return complex
-        return signature(types.float64, *args)
+        ret_typ = types.complex128 if isinstance(args[0].dtype, types.Complex) else types.float64
```
ehsantn: I think `ary.dtype` should also be checked since the output is complex if either input is complex.

Quasilyte: Ah, I see.
So, the `ary` is method receiver and `args[0]` is a first argument, so we have:
```python
cov(ary, args[0]) # Re-written as a func call
```
Is that correct?

ehsantn:
@Quasilyte thanks for the patch. Merging.

Quasilyte:
Hmm. I'll do `--amend` commits the next time to avoid multi-commit merging.
Or you can do a squash-merge. :)

ehsantn:
Multi-commit is fine (to preserve history). We don't do squash-merge yet but maybe we should at some point.

PR #46, opened 2018-09-07
Handle |= reductions

TocarIP:


Quasilyte:
Both C++ and Py sources in this commit have some formatting issues (missing space after `,` inside lists).
Turning on auto formatting is almost impossible, since all files seem to have such issues, there will be far too much noise.

ehsantn:
Since everything was written for sum and product reductions, I'm kind of surprised that this works but it does!
Can be merged after missing space after commas are fixed. Thanks.

ehsantn:
Actually, there is an error. `MPI_LOR` is logical _or_ but `|` performs bitwise _or_ operation (needs `MPI_BOR` I think).
The tests should check for this as well instead of silently passing (use `np.arange` and arithmetic to generate different numbers in parallel).

ehsantn:
What are your errors when running the whole suite?
I suggest making sure `A` and `B` are not identical in the tests (e.g. `np.arange(n)+1` for one of them).
Also, force push prevents seeing history and reviewing only latest changes (squash is a better option if number of commits became an issue).

TocarIP:
Sorry for force push. I've updated test to use different values.
When running  `mpiexec -n 4 python -m unittest`  without my changes, I  see:
```.....
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 31526 RUNNING AT antsatel03
=   EXIT CODE: 139
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

hpat/tests/test_basic.py
```diff
@@ -204,18 +204,25 @@ def test_reduce(self):
...
-            B = np.ones(3);
-            for i in numba.prange(N):
-                A += B
-            return A
+        binops = ['+=', '*=', '+=', '*=', '|=', '|=']
+        dtypes = ['float32', 'float32', 'float64', 'float64', 'int32', 'int64']
+        for (op,typ) in zip(binops,dtypes):
+            func_text = """def f(n):
+                  A = np.arange(np.{}(10));
+                  B = np.arange(np.{}(3), np.{}(3+10));
```
ehsantn: Please remove semicolons.
Also, I don't think the type of the generated arrays are correct. For example, `np.arange(np.float32(10))` will generate array of float64s instead of float32. The right way is to use `np.arange(0, 10, 1, np.float32)` (not using the dtype keyword argument since Numba doesn't support it).

ehsantn:
@TocarIP the crash is possible since we only test on 1-3 processors on Travis. Let's open a new ticket for it since it's not related to this PR. Also, could you investigate? You can use verbose mode of unittest to see which test is crashing.

ehsantn:
Thanks, merging.

PR #53, opened 2018-09-19
test travis rolling variable apply 2 3-4s

ehsantn:


shssf:
@fschlimb I think we need to drop this PR, needn't we?

fschlimb:
Looks like it. @ehsantn?

PR #54, opened 2018-09-19
WIP: max, min, timedelta

DrTodd13:


ehsantn:
Thanks, Todd. There is a lot to do here but this is a good start. There are test errors related to aggregate but they are minor; I'll fix them after merge.

PR #56, opened 2018-09-26
support series.combine

TocarIP:
Implementation is based on series.apply.

### DIFF
--- a/hpat/hiframes_typed.py
+++ b/hpat/hiframes_typed.py
@@ -757,6 +757,11 @@ class HiFramesTyped(object):
     def _handle_series_combine(self, assign, lhs, rhs, series_var):
         """translate s1.combine(s2,lambda x1,x2 :...) to prange()
         """
+        # error checking: make sure there is function input only
+        if len(rhs.args) < 2:
+            raise ValueError("not enough arguments in call to combine")
+        if len(rhs.args) > 3:
+            raise ValueError("too many arguments in call to combine")
         func = guard(get_definition, self.func_ir, rhs.args[1])
         if func is None or not (isinstance(func, ir.Expr)
                                 and func.op == 'make_function'):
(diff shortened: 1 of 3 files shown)


hpat/hiframes_typed.py
```diff
@@ -751,6 +754,96 @@ def f(arr, on_arr, w, center):  # pragma: no cover
             return nodes
         return [assign]
 
+    def _handle_series_combine(self, assign, lhs, rhs, series_var):
+        """translate s1.combine(s2,lambda x1,x2 :...) to prange()
+        """
+        func = guard(get_definition, self.func_ir, rhs.args[1])
```
ehsantn: The number of arguments and the type of the first argument should be checked here as well.

hpat/hiframes_typed.py
```diff
@@ -751,6 +754,96 @@ def f(arr, on_arr, w, center):  # pragma: no cover
...
+        # prange func to inline
+        if use_nan:
+            func_text = "def f(A, B):\n"
+        else:
+            func_text = "def f(A, B, C):\n"
+        func_text += "  numba.parfor.init_prange()\n"
+        func_text += "  n1 = len(A)\n"
+        func_text += "  n2 = len(B)\n"
+        func_text += "  n = n1\n"
+        func_text += "  if n2 > n:\n"
```
ehsantn: Control flow currently prevents the `init_prange` optimizations. Using `max(n1, n2)` is better here.

hpat/hiframes_typed.py
```diff
@@ -751,6 +754,96 @@ def f(arr, on_arr, w, center):  # pragma: no cover
...
+                        and stmt.value.op == 'call'):
+                    fdef = guard(get_definition, f_ir, stmt.value.func)
+                    if isinstance(fdef, ir.Global) and fdef.name == 'map_func':
+                        inline_closure_call(f_ir, _globals, block, i, func)
+                        break
+
+        # remove sentinel global to avoid type inference issues
+        ir_utils.remove_dead(f_ir.blocks, f_ir.arg_names, f_ir)
+        f_ir._definitions = build_definitions(f_ir.blocks)
+        if use_nan:
```
ehsantn: `arg_types` uses `series_var` for all argument types. It should use the types of `rhs.args[0]` and `rhs.args[2]`.

hpat/pd_series_ext.py
```diff
@@ -469,6 +469,21 @@ def resolve_apply(self, ary, args, kws):
...
         return self._resolve_map_func(ary, args, kws)
 
+    def _resolve_combine_func(self, ary, args, kws):
+        dtype = ary.dtype
+        # getitem returns Timestamp for dt_index and series(dt64)
+        if dtype == types.NPDatetime('ns'):
+            dtype = pandas_timestamp_type
+        code = args[1].value.code
+        f_ir = numba.ir_utils.get_ir_of_code({'np': np}, code)
+        f_typemap, f_return_type, f_calltypes = numba.compiler.type_inference_stage(
```
ehsantn: This assumes that both series have the same `dtype` which is not necessarily the case.

hpat/tests/test_series.py
```diff
@@ -611,6 +611,42 @@ def test_impl(S):
         S = pd.Series([1.0, 2., 3., 4., 5.])
         pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
 
+    def test_series_combine(self):
```
ehsantn: Tests for data types other than `float64` would be useful. Also, input series with different types would be good. For example, S1 with integers and S2 with floats can make sure output type is not just looking at S1's type.

hpat/hiframes_typed.py
```diff
@@ -751,6 +754,96 @@ def f(arr, on_arr, w, center):  # pragma: no cover
...
+        func_text += "  numba.parfor.init_prange()\n"
+        func_text += "  n1 = len(A)\n"
+        func_text += "  n2 = len(B)\n"
+        func_text += "  n = n1\n"
+        func_text += "  if n2 > n:\n"
+        func_text += "      n = n2\n"
+        func_text += "  S = numba.unsafe.ndarray.empty_inferred((n,))\n"
+        func_text += "  for i in numba.parfor.internal_prange(n):\n"
+        if use_nan:
+            func_text += "    t1 = np.nan\n"
```
ehsantn: `np.nan` cannot be stored in integer arrays. I think we could add a runtime assertion for integer output cases that prevents different series sizes (with a message asking for explicit float type annotation). This assertion can be after `init_prange` not to prevent optimizations.

ehsantn: Only the dtype of the output matters for this NaN handling. The output can be float with even both inputs being integer (e.g. a formula with division).

ehsantn: Pandas doesn't check type of inputs for conversion: https://github.com/pandas-dev/pandas/blob/master/pandas/core/series.py#L2303

TocarIP: In that case, I'm not sure why we should check series length in combine handler at all. map_func can return valid integer even if one of the inputs is Nan.

ehsantn: When sizes don't match, NaN values need to be assigned. Pandas throws an error when it can't assign NaN to integer output array:
```python
S1 = pd.Series([1,2,3,4])
S2 = pd.Series([1,2,3,4,5])
S1.combine(S2, lambda a,b: a+b)
```
`ValueError: cannot convert float NaN to integer`

TocarIP: I understand that, but map_func can return integer for Nan input, as a trivial example consider:
```
S1 = pd.Series([1,2,3,4])
S2 = pd.Series([1,2,3,4,5])
S1.combine(S2, lambda a,b: b)  # works fine with pandas
```

ehsantn: Ok, I see what you mean, it's not necessarily related to length. Sounds like a much harder check to always examine the output of the user function, which could have performance implications. I think we can skip the check. Thanks.

TocarIP: After thinking about it some more, we probably need an assert, because otherwise inputs of map_func are not type stable if input series is not float (input has either float type due to nan or input series type). I've tried removing assertions and as expected it broke integer combine testes.

ehsantn: Ok. makes sense.

ehsantn:
Looks good. Merging. Thanks.

Issue #63, opened 2018-10-05
Invalid result for HO function with bound instancemethod argument


Quasilyte:
Numba does partially support high-order functions and local closures, but the boundaries are not clearly defined as well as the inlining effect on it.

For HPAT, code below actually results in invalid results and no JIT compilation time error.
We should fix this bug and properly document what kinds of callable values HPAT supports.
After that, it would be simpler to give proper error messages to the user.
Right now it's unclear what are should be compiled and what should result in a compilation error.

Reproducer:
```python
import hpat
import pandas as pd

# Higher-order function
@hpat.jit
def indirect(call, fn):
    if call:
        return fn()
    return 0

@hpat.jit
def weird():
    df = pd.DataFrame({'A': [1]})
    return indirect(False, df.head)


# This one works fine
@hpat.jit
def indirect_ok(fn):
    return fn()

@hpat.jit
def works_as_expected():
    df = pd.DataFrame({'A': [1]})
    return indirect_ok(df.head)

print(weird())
print(works_as_expected())
```

Results for code without `@hpat.hit` decorator:
```
0
   A
0  1
```
Results for code with JIT:
```
Empty DataFrame
Columns: [A]
Index: []
   A
0  1
```

Quasilyte:
Numba fails to compile this code:
```python
import pandas as pd
import numpy as np
from numba import jit

@jit(nopython=True)
def indirect_ok(fn):
    return fn()

@jit(nopython=True)
def works_as_expected():
    a = np.array([1])
    return indirect_ok(a.sum)

print(works_as_expected())
```

Quasilyte:
Ah, I see.
With disabled inlining, HPAT gives same typing error for bound instancemethod passed as a function argument. Thanks.

ehsantn:
I looked at these examples some more. Looks like inlining is fine. For the top example, the `weird` function has a [type unification issue](http://numba.pydata.org/numba-doc/latest/user/troubleshoot.html#my-code-has-a-type-unification-problem), since either a dataframe (from `df.head`) or 0 is assigned to a variable based on a conditional. This will be caught when we make the hiframes pass a typed pass (the current pass has no way of knowing).

I think the `a.sum` example triggers a bug in Numba's parfor preprocessing stage where `a.sum` should be converted to `np.sum(a)` but looks like the call doesn't have the array argument.

PR #64, opened 2018-10-05
hiframes: better error reporting for getattr

Quasilyte:
- Report bound instancemethod references that were not
  immediately invoked (see #63)
- Report undefined (unknown) data frame attributes access

TODO: fill "unimplemented" lists, so HPAT can report
"unimplemented yet" instead of "unknown attribute".

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>

### DIFF
--- a/hpat/hiframes.py
+++ b/hpat/hiframes.py
@@ -152,12 +152,6 @@ class HiFrames(object):
             func_ir, typingctx, args, _locals, self.reverse_copies)
         self.h5_handler = pio.PIO(self.func_ir, _locals, self.reverse_copies)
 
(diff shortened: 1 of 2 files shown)


ehsantn:
I don't think we should add this instance method restriction. Dead code elimination removes unused ones easily, and we can catch other invalid cases (e.g. returned ones). #63 seems like an issue with inlining.

ehsantn:
I see there are still TODOs. Let me know when this is ready for review and merge.

Quasilyte:
Not working on HPAT anymore, sorry.

PR #70, opened 2018-11-01
Support latest version of pyarrow.

TocarIP:
Pyarrow has recently updated to 0.11.1, which doesn't have BitUtil::BitNotSet.
Which was removed in https://github.com/apache/arrow/pull/2355.
Use !BitUtil::GetBit instead, as suggested in PR above.


ehsantn:
Thanks, merging. CI is broken due to other issues. We have quite a bit of maintenance work to do.

PR #72, opened 2018-11-05
adding dict bytes_vec_int64, cleaning up

fschlimb:
* adding byte_vec_int64 dictionary
* replacing some macros with templates for better maintanability/debugability/readability
* removing left-overs from previous versions

test_hiframes.TestHiFrames.unique and others still failing, don't know why.

Still requires some clean-up in python part of dictionary handling (using generic path only).


ehsantn:
They might be failing because of string type (it had special handling).

PR #73, opened 2018-11-07
adding dict bytes_vec_int64, cleaning up

fschlimb:
* adding byte_vec_int64 dictionary
* replacing some macros with templates for better maintanability/debugability/readability
* removing left-overs from previous versions
test_hiframes.TestHiFrames.unique and others still failing, don't know why.

Still requires some clean-up in python part of dictionary handling (using generic path only).

supersedes #72

### DIFF
--- a/hpat/_dict_ext.cpp
+++ b/hpat/_dict_ext.cpp
@@ -1,114 +1,366 @@
+/*
+ * Implementation of dictionaries using std::unordered_map.
+ *
+ * Provides most common maps of simple data types:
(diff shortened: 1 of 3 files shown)


fschlimb:
About [UnorderedAssociativeContainers:](https://en.cppreference.com/w/cpp/named_req/UnorderedAssociativeContainer)
"
_Rehashing invalidates iterator and might cause the elements to be re-arranged in different buckets but it doesn't invalidate references to the elements._
"

PR #79, opened 2019-03-01
D4p

fschlimb:
@ehsantn Moving daal4py functionality to daal4py. Adding generic mechanism to add call-analysis from external. We might want to generalize and extend this mechanism. Right now we need an extra import statement, it'd be nice to auto-detect when both are used. For that we need to somehow break the import cycle and add an extra init-step to HPAT, like in hpat.jit.

See also https://github.com/IntelPython/daal4py/pull/68


examples/d4p_kmeans.py
```diff
@@ -1,19 +1,16 @@
 import daal4py
+import daal4py.hpat
 import hpat
 import numpy as np
 
-daal4py.daalinit(spmd=True)
-
-@hpat.jit
+@hpat.jit(nopython=True)
```
ehsantn: `(nopython=True)` is not necessary since it is on by default.

fschlimb: Yea, I know. Just a left-over from my desperate attempts to get the error output of actually failing features...I will remove the examples anyway...

fschlimb:
@ehsantn if you are ok with the mechanism of registering d4p analysis, I suggest we merge now. I will merge the d4p counterpart even though it's incomplete. It's sufficient to make the tutorial work.

ehsantn:
Similar to what Numba does, I think a dictionary of implementations is better than a list and a for loop for every program function. Keys will be function names or variable type and method name.

fschlimb:
@ehsantn Like now?

PR #81, opened 2019-03-29
tutorial

fschlimb:
tutorial as held in London 20190328

Before merging we need
* d4p branch merged to HPAT
* hpat branch merged to daal4py
* binary packages available with the above changes
* updated README with instructions of how to setup an operating env
  * a conda-env file would probably be the easiest
  * must work on Windows

The sooner we get this done, the better.

@ehsantn


ehsantn:
merging this to avoid dependency on me.

PR #90, opened 2019-07-01
Move to use Intel IMPI conda package for Windows

Vyacheslav-Smirnov:
Use impi-devel and impi_rt conda packages for
Windows instead of Intel(R) MPI.
This will remove requirement to have already
installed MPI for HPAT build on Windows.

PR #91, opened 2019-07-15
Skipped failed tests

anmyachev:
I propose so far to skip tests that fall. Thus, it will be possible to work in parallel to correct the tests and, for example, to do implementation of new features.

@fschlimb @Alexander-Makaryev @Vyacheslav-Smirnov what you think about it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,6 @@ matrix:
         - env: CONDA_ENV=travisci NUM_PES=2
         - env: CONDA_ENV=travisci NUM_PES=3
 
-branches:
-    only:
-        - master
-
 before_install:
     - buildscripts/setup_conda.sh
     - export PATH=$HOME/miniconda3/bin:$PATH
--- a/hpat/tests/test_basic.py
+++ b/hpat/tests/test_basic.py
@@ -51,6 +51,8 @@ class BaseTest(unittest.TestCase):
 
 
 class TestBasic(BaseTest):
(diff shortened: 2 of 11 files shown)


fschlimb:
Are there any features being worked on that depend on each other or any other non-merged PR?
If not, I do not see how this would help. We can develop features and accompanying tests independently.

anmyachev:
@fschlimb you have, for example, two PRs: #89, #90. It would be great to know if these changes break any tests. In the case when there are already tests in the master that do not pass, it is difficult to track the impact of new changes that will be pushed on the master.

fschlimb:
Our priority right now is to get things stabilized, e.g. fix issues in existing stuff. New features, improvements and optimizations are secondary (or less).

OK, let's be practical. It's probably easier to start from a green state. Let's merge this now. 

However, from then on, we should merge only bug-fixes until all these tests are corrected an enabled.

anmyachev:
> "fix needed" might be better than "needed fix", though.

I'll fix it now

anmyachev:
> Our priority right now is to get things stabilized, e.g. fix issues in existing stuff. New features, improvements and optimizations are secondary (or less).
> 
> OK, let's be practical. It's probably easier to start from a green state. Let's merge this now.
> 
> However, from then on, we should merge only bug-fixes until all these tests are corrected an enabled.

@fschlimb OK I understood

PR #93, opened 2019-07-17
[BUG] Fixed problems with generation parquet files

anmyachev:
Some tests were skipped by me, because I did not think that I needed to separately generate data.

Besides I think that unit tests should be run without additional run of the data generation script.

### DIFF
--- a/hpat/tests/gen_test_data.py
+++ b/hpat/tests/gen_test_data.py
@@ -5,6 +5,56 @@ import pyarrow as pa
 import pandas as pd
 
 
+class ParquetGenerator:
+    GEN_KDE_PQ_CALLED = False
+    GEN_PQ_TEST_CALLED = False
+
+    @classmethod
+    def gen_kde_pq(cls, file_name='kde.parquet', N=101):
+        if not cls.GEN_KDE_PQ_CALLED:
+            df = pd.DataFrame({'points': np.random.random(N)})
+            table = pa.Table.from_pandas(df)
+            row_group_size = 128
+            pq.write_table(table, file_name, row_group_size)
+            cls.GEN_KDE_PQ_CALLED = True
+
+    @classmethod
+    def gen_pq_test(cls):
+        if not cls.GEN_PQ_TEST_CALLED:
+            df = pd.DataFrame(
+                {
+                    'one': [-1, np.nan, 2.5, 3., 4., 6., 10.0],
+                    'two': ['foo', 'bar', 'baz', 'foo', 'bar', 'baz', 'foo'],
+                    'three': [True, False, True, True, True, False, False],
+                    # float without NA
+                    'four': [-1, 5.1, 2.5, 3., 4., 6., 11.0],
+                    # str with NA
+                    'five': ['foo', 'bar', 'baz', None, 'bar', 'baz', 'foo'],
+                }
+            )
+            table = pa.Table.from_pandas(df)
+            pq.write_table(table, 'example.parquet')
+            pq.write_table(table, 'example2.parquet', row_group_size=2)
+            cls.GEN_PQ_TEST_CALLED = True
+
+
+class SparkGenerator:
+
+    @staticmethod
+    def generate():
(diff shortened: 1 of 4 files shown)


fschlimb:
I like the idea of generating data on the fly.

Could you explain in more detail why you disabled tests?

anmyachev:
Specifically, I mistakenly turned off these tests, because I tested it locally on a machine and (which was not very obvious to me) missed the data generation phase (`python hpat/tests/gen_test_data.py`), which runs when testing occurs in the Trave CI.

For example, a comment below says that data is generated only for `test_io.py` which isn't True.
(file - hpat/buildscripts/test.sh)
```python
# generate test data for test_io
python hpat/tests/gen_test_data.py
```

hpat/tests/gen_test_data.py
```diff
@@ -5,102 +5,132 @@
 import pandas as pd
 
 
+class ParquetGenerator:
```
shssf: Are you sure we need a class here (same for `SparkGenerator`)? It looks like it can be just a functions. it is ok, but I think it is better to make things simpler.

hpat/tests/gen_test_data.py
```diff
@@ -5,102 +5,132 @@
...
+                             StructField('DATE', TimestampType(), True)])
+        sdf = spark.createDataFrame(df, schema)
+        sdf.write.parquet('sdf_dt.pq', 'overwrite')
+
+        spark.stop()
+
+
 def gen_lr(file_name, N, D):
-    points = np.random.random((N,D))
+    points = np.random.random((N, D))
```
shssf: Please, next time do not mix code style changes and algorithm changes.

anmyachev: OK

hpat/tests/gen_test_data.py
anmyachev: `class ParquetGenerator` stores information about the generated data. If replace methods with functions, will have to use global variables to control generated data state. Using global variables is not good and I did not think of a third option :)

The class for another generator(`class SparkGenerator`) is really not needed yet, so I replaced it with function.

hpat/tests/gen_test_data.py
```diff
@@ -5,102 +5,129 @@
...
+    dt1 = pd.DatetimeIndex(['2017-03-03 03:23',
+                            '1990-10-23', '1993-07-02 10:33:01'])
+    df = pd.DataFrame({'DT64': dt1, 'DATE': dt1.copy()})
+    df.to_parquet('pandas_dt.pq')
+
+    spark = SparkSession.builder.appName("GenSparkData").getOrCreate()
+    schema = StructType([StructField('DT64', DateType(), True),
+                         StructField('DATE', TimestampType(), True)])
+    sdf = spark.createDataFrame(df, schema)
+    sdf.write.parquet('sdf_dt.pq', 'overwrite')
```
shssf: Is it possible to move this directory (and other generated data) under "build" directories tree? I don't think we need it on top sources level.

anmyachev: The location of the test folder is quite good, as far as I can see, it is used in many large projects (for example, pandas).

The generated data is really out of place.

I suggest leaving the test folder in place and transferring the generated data into it.

anmyachev: I am ready to do it, but probably better in separate PR

anmyachev: @shssf This PR can be merged? There is no blocker?

shssf: @anmyachev yes, but lets wait @fschlimb approval.

anmyachev: @shssf ok, thanks

fschlimb: Yes, this is standard practice to have test folders to each source folder.
I am not sure I follow the discussion about "generated data". We should not generate any data or source at build or test time into source directories. Build-time and run-time files should always be separate.

PR #95, opened 2019-07-17
Fixing issue with named series handling in fillna

kozlov-alexey:


hpat/hiframes/api.py
```diff
@@ -361,7 +361,7 @@ def generic(self, args, kws):
 class FillNaStrType(AbstractTemplate):
     def generic(self, args, kws):
         assert not kws
-        assert len(args) == 2
+        assert len(args) == 3
```
anmyachev: It is necessary to correct the comment under this line.

fschlimb: Please do so.

hpat/tests/test_series.py
```diff
@@ -627,12 +620,16 @@ def test_impl(A):
         pd.testing.assert_series_equal(hpat_func(df.A), test_impl(df.A),
             check_names=False)
 
-    @unittest.skip('numba.errors.LoweringError - fix needed\n'
-                   'Failed in hpat mode pipeline '
-                   '(step: nopython mode backend)\n'
-                   'expecting {{i64, i64, i32*, i8*, i8*, i8*}, i8*, '
-                   '{i8*, i64, i32, i32, i64, i8*, i8*}} \n'
-                   'but got {{i64, i64, i32*, i8*, i8*, i8*}, i8*, i8*}\n')
+    def test_series_fillna_inplace1(self):
```
fschlimb: Please add a comment what this function is testing.

kozlov-alexey: I added some comment (there are no comments for other tests though).
This new test just adds missed check for attempting to jit inplace fillna on series with numeric data. There was a test for string series, but not numeric one.

hpat/hiframes/api.py
kozlov-alexey: It's done.

hpat/tests/test_series.py
fschlimb: That's fine. We want to improve code style and docu. So anything new should be well documented.

PR #97, opened 2019-07-20
[STL] PEP8 code style for 'test_strings.py', 'test_utils.py', 'test_series.py'

anmyachev:


anmyachev:
@shssf, @fschlimb can the PR be merged into master?

shssf:
Before doing code style changing we need to implement the procedure for this. It should be a command line way to allow all developers to change and check code style in the same way.
So, we need `python ./setup.sh codestyle-check` and `python ./setup.sh codestyle-apply` command before merging any code style changes.

shssf:
Please see #103

PR #98, opened 2019-07-24
[BUG] tests folder should be omitted for coverage run

anmyachev:


shssf:
Did you mean "should not be"?

anmyachev:
> Did you mean "should not be"?

I mean that the `hpat/tests` folder should not be taken into account when calculating the test coverage. That is, must be specifically omitted for `coverage run` command (for example: 'coverage run --source ./hpat --omit ./hpat/tests/*')

fschlimb:
why?

anmyachev:
@fschlimb code coverage metric must show the percentage of the program **source** code that was executed during the testing process. Unless you explicitly exclude the testing folder, then this metric will be affected by code changes in the tests themselves.

For example, you can look at #97 , in which the coverage metric is reduced simply because I replaced a few lines in tests. It's false positive.

fschlimb:
I guess it's a matter of what you want to see. Coverage of the test-code is also an interesting metric.

anmyachev:
@shssf @fschlimb "Travis CI - Branch Expected — Waiting for status to be reported" it looks like a bug.
Maybe this will help? https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154

shssf:
@anmyachev I see this issue in every PR in this repository

anmyachev:
@shssf @fschlimb  FYI.
This fix can most likely be made **only by the owner** and should not take more than 10 minutes I think.

anmyachev:
@shssf can the PR be merged?

shssf:
@fschlimb Yes, I think yes.

PR #99, opened 2019-07-26
Fix for Pandas merge overload wrong handling of string literals

kozlov-alexey:


hpat/hiframes/pd_dataframe_ext.py
```diff
@@ -625,38 +625,63 @@ def generic(self, args, kws):
...
+                  right_by=None, suffixes=('_x', '_y'), tolerance=None,
+                  allow_exact_matches=True, direction='backward'):
+            if on is not None:
+                left_on = right_on = on
 
-        return hpat.hiframes.api.join_dummy(
-            left, right, left_on, right_on, 'asof')
+            return hpat.hiframes.api.join_dummy(left, right, left_on, right_on, 'asof')
 
     return _impl
```
shssf: I'm not sure but I think these functions (including `merge_overload`) might be implemented easier. I mean, use `if\else` clause to calculate proper `left_on` and `right_on` variables and return the same `_impl` function.

kozlov-alexey: I tried that at first (i.e. assign new_right/left_on to proper values in context of merge_asof_overload() and remove lines 677 and 678 at all, hence using same body for _impl()), but for some reason it breaks currently working test_join1_seq_key_change1 (the only test we have currently to test "on=['A']" case). In other words, it looks like assignment:
     left_on = right_on = on
only works for lists if it's in the context of _impl(). The error was as below:
     KeyError: "Failed in hpat mode pipeline (step: convert to distributed)\nlist[('A',)](unicode_type)""
probably because the type of 'on' in context of merge_asof_overload() is not the same as type of 'on' in _impl(), hence providing one instead of the other causes failure.
So I ended up using two different overloads - legacy one for lists and new one taking literal values for strings.
Maybe there's a way to copy Numba's ConstList value 'literally' to new_left/right_on variables, but I don't know what it is.

fschlimb:
I don't think I follow. I guess @shssf suggested something like the below. Why doesn't it work?

```python
@overload(pd.merge_asof)
def merge_asof_overload(left, right, on=None, left_on=None, right_on=None,
                        left_index=False, right_index=False, by=None, left_by=None,
                        right_by=None, suffixes=('_x', '_y'), tolerance=None,
                        allow_exact_matches=True, direction='backward'):

    # 'on' can be a list or a string literal
    if isinstance(on, types.StringLiteral):
        # use literal strings value to fetch them as consts from IR later
        new_right_on = on.literal_value
        new_left_on = on.literal_value
    elif on is not None:
        new_right_on = on
        new_left_on = on

        def _impl(left, right, on=None, left_on=None, right_on=None,
                  left_index=False, right_index=False, by=None, left_by=None,
                  right_by=None, suffixes=('_x', '_y'), tolerance=None,
                  allow_exact_matches=True, direction='backward'):

            return hpat.hiframes.api.join_dummy(left, right, new_left_on, new_right_on, 'asof')

    return _impl
```

fschlimb:
Ignore my previous (deleted) comment. 
What is the error of the original code?

kozlov-alexey:
@fschlimb
The original error was "ValueError: Failed in hpat mode pipeline (step: typed dataframe pass)" in
get_const_or_list() in dataframe_pass.py. I'm attaching the file with full traceback here: https://github.com/IntelPython/hpat/files/3441340/full_traceback_for_test_join.txt

fschlimb:
1. You should make sure you're using a compatible version of numpy
2. This means you have to understand why `get_const_or_list()` fails. Its name suggests it is intended to address exactly the issue you observe with literals/consts.

kozlov-alexey:
@fschlimb
1. OK (it goes away with downgrading to scikit-learn=0.19.1 and numpy=1.16.1)
2. I believe the problem is in how Numba treats 'on' variable in the original merge implementation.
That is, during Numba type inference 'on' becomes an instance of TypeVar class, so assigning later this value to left_on and right_on will not make them the same constant (we need to assign obj.literal_value for that). Maybe Numba fails to add const names 'left_on' and 'right_on' in IR when it copies Literal[Str] object? 

And in _get_const_or_list() we merely try to find const (or lists) objects by name in self.func_ir, which fails due to above (no such const names exist in IR). I have no explanation why it works for lists though, probably, because they are created differently (via build sequence).

@overload_method(DataFrameType, 'merge')
@overload(pd.merge)
def merge_overload(left, right, how='inner', on=None, left_on=None,
        right_on=None, left_index=False, right_index=False, sort=False,
        suffixes=('_x', '_y'), copy=True, indicator=False, validate=None):

    def _impl(left, right, how='inner', on=None, left_on=None,
            right_on=None, left_index=False, right_index=False, sort=False,
            suffixes=('_x', '_y'), copy=True, indicator=False, validate=None):
        if on is not None:
            left_on = right_on = on       # <-------- Even though 'on' was string literal, here we assign 
                                          # some Numba object, not object's literal value, hence left_on and 
                                          # right_on are not found as const objects in self.func_ir later
        return hpat.hiframes.api.join_dummy(
            left, right, left_on, right_on, how)

    return _impl

kozlov-alexey:
@fschlimb 
Thank you for helping me sort that out and suggesting a better solution!
I had to modify the check a bit though, i.e. instead of 
    if isinstance(on, types.NoneType)
use 
   if isinstance(numba.typeof(on), types.NoneType)

because, with the first version other tests that do not provide 'on' argument fail.
It looks like that check always returns True. It may be because Numba's NoneType is not inherited from python's None:
https://github.com/numba/numba/issues/3590

So we can either compare to NoneType but use typeof(on), as in this commit, or we can use the following negated check:
    onHasValidType = isinstance(on, types.StringLiteral, types.List)
both of which appear to work.

kozlov-alexey:
I had to skip back one of the tests - test_merge_asof_parallel1 - which still fails on NUM_PES=3 build now with a different error and only on the second test run.  It can be seen in Travis logfile that it is passed on the first run though, so probably it's due to the problem with 
test_series_head_index_parallel1
which is not skipped at the moment, but was identified as a root cause of corrupting memory and causing other test failures. 

I'm going to deal with skipping test_series_head_index_parallel1 and unskipping other tests in PR#100 anyway, so I hope it's not a big issue.

PR #100, opened 2019-07-26
Unskipping unit tests from test_basic.py that currently pass

kozlov-alexey:
Tested locally by running "python -u -m unittest hpat.tests" and "mpiexec -n 3 python -u -m unittest -v hpat.tests" several times - uncommented tests in test_basic.py are passed, so I'm proposing to unskip them.


fschlimb:
@kozlov-alexey Some test in PR fail.

kozlov-alexey:
It seems tests are launched differently in travis CI, i.e. I ran "mpiexec -n 3 python -u -m unittest -v hpat.tests" and buildscripts/test.sh used by travis runs "mpiexec -n $NUM_PES python -u -m unittest -v", and the latter actually results in each test being run twice (see total number of tests is 948 in green jobs). Not clear why it causes segfault though, continue debugging.

shssf:
@kozlov-alexey Do you plan to make this PR to work or we have to wait some time?

kozlov-alexey:
@shssf 
Yes, but I want to unskip as many tests as possible, not just tests from test_basic.py. It turns out that most of them pass if launched as a standalone test, and also pass if we skip certain unstable parallel tests (such as test_series_head_index_parallel1). The latter always passes if executed once, and always fails if executed twice. Moreover it causes failures/crashes of other test executed afterwards. I'm going to push final version later today.

kozlov-alexey:
@shssf
Please don't merge it yet - even though tests are passed here, now they are failing locally.

kozlov-alexey:
I think this can be merged now, as I am no longer able to reproduce the fails I had (tried with and w/o clearing pycache, -B, mpi) and they are not see in CI, maybe it was some environment issue. Sorry for messing this up.

PR #102, opened 2019-07-31
Remove spark dependency

Vyacheslav-Smirnov:
Spark is used for only 1 df generation for tests execution.
There is a problem to get spark works on Windows.
So remove spark dependency from HPAT and add generated
sdf_dt.pq data frame to HPAT repo instead.


shssf:
1. Do not put binary data on git
2. I see here is algorithm changed. Previously DF generated by interfaces but in new version it just unpack binary data from archive.
I don't think this is correct approach.

fschlimb:
I had suggested using a small archive with input data rather than creating a dependency on spark. The archive is really small so not a big deal. We didn't see a good way to quickly generate spark-data without spark. Sugegestions are welcome.

What is wrong what is wrong about pre-generated input data (except maybe that larger binaries are problematic in git if they change frequently) ?

shssf:
@fschlimb How many tests we have which are depends on this Spark data? I mean, may be better to put them into separate "module/directory" and run them if "spark requirements" satisfied (+ generate data for them "on the fly" instead "fixed file with data")?

I would propose to avoid using pre-generated binary data because in long-term it might brings more pain than algorithm to generate it.

At least, please put that archive into subfolder. For example, "data" or "static data".

In case of "binary in git":
The problems begin when git needs to generate diffs and merges: git cannot generate meaningful diffs, or merge binary files in any way that could make sense. So all merges, rebases or cherrypicks involving a change to a binary file will involve you making a manual conflict resolution on that binary file.
Also, you need to decide whether the binary file changes are rare enough that you can live with the extra manual work they cause in the normal git workflow involving merges, rebases, cherrypicks.

fschlimb:
@shssf Yes, I agree with all you said, I only think in this particular case risk for causing issues is very low, not worth skipping a test. Both is fine with me, though.

PR #103, opened 2019-07-31
HPAT Build: Code style check for C and Python sources

shssf:
- Python sources can not be fully fixed automatically for all rules required by pep8
- Excluded `E265,W503` rules from style check due to reason above. Can be get them back in future.
- Algorithm improvement might be addressed in other PRs
- Only `build` and `.git` directories and sub directory excluded from search paths

To see documentation please use `python ./setup.py --help-commands` for new build option and `python ./setup.py style --help` for its parameter


setup.py
```diff
@@ -288,6 +315,124 @@ def readme():
...
+                command_cout, command_cerr = command_output.communicate()
+                if not self.apply:
+                    if command_cout.find(b'<replacement ') > 0:
+                        bad_style_file_names.append(f)
+        except BaseException:
+            print("%s is not installed.\nPlease use: %s" % (self._c_formatter, self._c_formatter_install_msg))
+            exit(1)
+
+        # Python files handling
+        py_files = self._get_file_list(root_dir, self._py_file_extensions)
```
anmyachev: Maybe you should use `pycodestyle` itself? For example, `pycodestyle hpat -q` command will return a list of files from `hpat` dir that do not satisfy for PEP8 style.

setup.py
```diff
@@ -288,6 +315,124 @@ def readme():
...
+    _c_formatter_command_line = [_c_formatter, '-style=file']
+    _c_file_extensions = ['.h', '.c', '.hpp', '.cpp']
+
+    _py_checker = 'pycodestyle'
+    _py_formatter = 'autopep8'
+    _py_formatter_install_msg = 'pip install --upgrade autopep8\npip install --upgrade pycodestyle'
+    # E265 and W503 excluded because I didn't find a way to apply changes automatically
+    _py_checker_command_line = [_py_checker, '--ignore=E265,W503']
+    _py_formatter_command_line = [_py_formatter, '--in-place', '--aggressive', '--aggressive']
+    _py_file_extensions = ['.py']
```
fschlimb: What about using yapf? Seems to produce nicer code.

setup.py
```diff
@@ -288,6 +315,124 @@ def readme():
...
+                if not self.apply:
+                    command_output = subprocess.Popen(
+                        self._py_checker_command_line + [f])
+                    returncode = command_output.wait()
+                    if returncode != 0:
+                        bad_style_file_names.append(f)
+                else:
+                    command_output = subprocess.Popen(
+                        self._py_formatter_command_line + [f])
+                    command_output.wait()
```
fschlimb: maybe autopep8 also allows this, but yapf would allow to simplify this: just call `yapf.yapflib.yapf_api.FormatFile(f)` instead of using a pipe.

setup.py
```diff
@@ -288,6 +315,124 @@ def readme():
...
+        except BaseException:
+            print("%s is not installed.\nPlease use: %s" % (self._py_formatter, self._py_formatter_install_msg))
+            exit(1)
```
fschlimb: this is a rather generic `except`, but the message is very specific. This can be very confusing if something else goes wrong. This should at least print the actual exception (message).
Using a python module rather than a pipe/process will also help with better error messages.

setup.py
```diff
@@ -288,6 +315,124 @@ def readme():
...
+        ('apply', 'a', 'Apply codestyle changes to sources.')
+    ]
+    description = "Code style check and apply (with -a)"
+    boolean_options = []
+
+    _result_marker = "Result:"
+    _project_directory_excluded = ['build', '.git']
+
+    _c_formatter = 'clang-format-6.0'
+    _c_formatter_install_msg = 'pip install clang'
```
anmyachev: As I looked [here,](https://pypi.org/project/clang/) this command does not install `clang` itself. Where is the code that installs `clang`?

fschlimb: You should not pip install anything in the first place. Just import what you need. setup.py should not modify the user's python env. It is possible to import only when really running 'style' so you're not dependent on it if you don't. But I don't see where this is using the clang package anyway.

setup.py
shssf: Do you know a way to exclude `build` directory from file list provided by command `pycodestyle -q`? Because `pycodestyle hpat -q` gives the list in subdirectory only.

setup.py
shssf: Did you try it? I looked to yapf briefly and stay with pep8 (autopep8 and pystyle as successors). I have no objections to use yapf for pyton part of sources.

setup.py
shssf: It require a check for yapf module unavailability. If it is ok I will implement.

setup.py
anmyachev: Something like that: `pycodestyle -q --exclude build,examples,docs`

setup.py
shssf: > As I looked [here,](https://pypi.org/project/clang/) this command does not install `clang` itself. Where is the code that installs `clang`?

I had no task to install `clang`. I just looked for clang-format tool installation.

setup.py
shssf: Do you have an experience with `yapf`? @fschlimb requested to move to this tool. Do you think it is better to stay with `pycodestyle`?

setup.py
fschlimb: Not sure I understand what you mean. In this python else branch you can just import it. If it's not available you will get a meaningful error. 
As mentioned elsewhere, installing packages in setup.py should be avoided. So we'd have the same issue with autopep8.

setup.py
fschlimb: No I didn't. I like their approach, though. Can autopep8 be used a python library?
Also, do the clang bindings for python allow using the code-formatting tool (without using a command-line tool through opening a pipe)?

setup.py
shssf: > you can just import

What is no `yapf` package installed in the system? In this case the user take long, confused python error message. I think we need to provide simple, readable message about this. To do this I need to check the package availability first. it is not big code (several lines) but i think it should be.

fschlimb: ok, I don't think the error message is hard to read. But you can just wrap the import in a try/except and print an error of your liking.

setup.py
anmyachev: I don't have any experience with `yapf`.

Probably better to choose the tool with which one of you worked the longest and has significant experience. Moreover, in case the tool needs to be changed for some reason  I do not see the problem of switching to a new one.

I suggest to stay on `yapf` and move on to solving other problems :)

shssf: > and move on to solving other problems

Completely agree!!!

setup.py
shssf: @fschlimb I tried `yapf -i ./setup.py` and found:
- it has same limitations as `autopep8` (E501, E265)
- it formats code worse than `autopep8 -i ./setup.py` (look into Extension lines formatting)
I didn't find python module for `clang-format`.

setup.py
shssf: > setup.py should not modify the user's python env

It is not implemented in this PR. This line is just help message.

fschlimb: Oops...

setup.py
fschlimb: ok, let's get something done that's reasonable and improve as needed.

shssf:
I addressed part of comments.
- I did not change `autopep8` to `yapf` due to `yapf` provide worse formatting.
- I didn't implement Python module based style check/adjustment. Currently it works consistently for C and Pyton sources. If somebody want to do this, please address this in separate PR.

PR #104, opened 2019-08-01
explicitly adding data-file

fschlimb:
@shssf @Vyacheslav-Smirnov 
Non-python files need to be explicitly declared to bet copied. Not sure why this didn't trigger on linux.
This now builds on Windows and runs almost all tests - 9 StreamReader tests fail.


Vyacheslav-Smirnov:
Thank you, Frank!
I have found this problem yesterday too, but did not find a solution.

PR #114, opened 2019-08-08
Add tests for Series arithmetic and comparison methods and fix Series.div

kozlov-alexey:


hpat/tests/test_series.py
```diff
@@ -417,76 +417,106 @@ def test_impl(A):
...
+            n = 11
+            df = pd.DataFrame({'A': np.arange(1, n)})
+            pd.testing.assert_series_equal(hpat_func(df.A, 1), test_impl(df.A, 1), check_names=False)
 
     def test_series_op3(self):
-        def test_impl(A, i):
-            A += i
-            return A
-        hpat_func = hpat.jit(test_impl)
+        arithmetic_binops = ('+', '-', '*', '/', '//', '%', '**')
```
shssf: Please extract duplicated code into separate function.

hpat/tests/test_series.py
```diff
@@ -417,76 +417,106 @@ def test_impl(A):
...
 
-        n = 11
-        df = pd.DataFrame({'A': np.arange(n)})
-        pd.testing.assert_series_equal(hpat_func(df.A, 1),
-                                       test_impl(df.A, 1), check_names=False)
+        for operator in arithmetic_binops:
+            func_text = "def test_impl(S, i):\n"
+            func_text += "   return S {} i\n".format(operator)
```
shssf: Is it possible to avoid autogeneration of the code and `exec()` method?

hpat/hiframes/pd_series_ext.py
```diff
@@ -974,28 +974,27 @@ def array_attribute_attachment(self, ary):
...
-    operator.floordiv: 'floordiv',
-    operator.mod: 'mod',
-    operator.pow: 'pow',
-    operator.lt: 'lt',
-    operator.gt: 'gt',
-    operator.le: 'le',
-    operator.ge: 'ge',
-    operator.ne: 'ne',
-    operator.eq: 'eq',
+    'add': operator.add,
```
shssf: Do you think it will faster? Or why you changed the order?

hpat/tests/test_series.py
```diff
@@ -417,76 +417,106 @@ def test_impl(A):
...
     def test_series_op6(self):
-        def test_impl(A, B):
-            return A.eq(B)
-        hpat_func = hpat.jit(test_impl)
-
-        n = 11
-        A = pd.Series(np.arange(n))
-        B = pd.Series(np.arange(n)**2)
-        pd.testing.assert_series_equal(hpat_func(A, B), test_impl(A, B))
+        comparison_binops = ('<', '>', '<=', '>=', '!=', '==')
```
shssf: Do we want to check the case `if a=b` and evaluate properly throw exception?

hpat/tests/test_series.py
```diff
@@ -499,10 +529,15 @@ def test_impl(A, B):
...
-    @unittest.skip('AssertionError - fix needed\n'
-                   'Attribute "dtype" are different\n'
-                   '[left]:  int64\n'
-                   '[right]: int32\n')
+    def test_series_abs(self):
+        def test_impl(A):
+            return A.abs()
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series([1, 2.2, -5.8, 3, 2, -1.2, -4, 4.7])
```
shssf: Does it supports bin, oct and hex numbers? Scientific notation (0.5E-01)?

hpat/tests/test_series.py
kozlov-alexey: @shssf Sorry, I don't understand what duplication has to be removed, do you mean combine all tests for arithmetic operations (i.e. op1 - op3) into one single tests?

hpat/tests/test_series.py
kozlov-alexey: Not sure if decorator-solution works here, as we use operators in the function body.
Will check, but I don't really think it's a problem, because it's not in functional code, but in tests and they seem to be simpler with this form.

hpat/hiframes/pd_series_ext.py
kozlov-alexey: No, the order was changes because there were two values with the same key in original explicit_binop_funcs:
    operator.truediv: 'div',
    operator.truediv: 'truediv',

Hence iterating over (key, value) pairs didn't work and only one (truediv) Series method was generated, so Series.div didn't work at all - it failed with:
numba.errors.TypingError: Failed in hpat mode pipeline (step: nopython frontend)
Unknown attribute 'div' of type series(int64, array(int64, 1d, C), none, False)

hpat/tests/test_series.py
kozlov-alexey: Hmm, I didn't think of it. But it as far as I understand it won't test HPAT anyway, it will test Python, because SyntaxError will be raised before we attempt to jit-compile anything. Am I wrong?

hpat/tests/test_series.py
kozlov-alexey: @shssf I overlooked it, there's no need in this test at all - there is another one test_series_abs1, which also checks NaN. Anyway, it works, yes. I will add hex/dec/oct literals to it.

hpat/tests/test_series.py
shssf: Not in one test but these tests have many similar code. The idea is to make this code as separate function which could be called from these tests.

hpat/tests/test_series.py
shssf: I don't know. I just propose the idea to implement additional checks. It is ok if you think it is not applicable here.

Vyacheslav-Smirnov:
For me this auto generation looks too complicated.
Maybe organize each test like this:
def test_impl()
    return [A + B,
               A += B,
              A.add(B)]
and then check each result from list?
Or create separate test for each case?

PR #119, opened 2019-08-12
Add gettatr for Series index

akharche:


hpat/tests/test_series.py
```diff
@@ -1429,6 +1429,31 @@ def test_impl(S):
...
+
+    def test_series_index2(self):
+        def test_impl():
+            A = pd.Series([1, 2, 3], index=[0, 1, 2])
+            return A.index
+
+        hpat_func = hpat.jit(test_impl)
+        np.testing.assert_array_equal(hpat_func(), test_impl())
+
+    @unittest.skip("Enabel after fixing distributed for get_series_index")
```
densmirn: type: enable

shssf:
@akharche Is this PR solves the issue with `len(None)`? Should this PR enable more tests?

akharche:
@shssf  It does not solve this problem, PR adds small feature to get Series.index. This functionality appears in workloads and was commented as TODO.
Problem with `len(None)` is more complicated, it is in progress.

PR #120, opened 2019-08-12
CI on Windows

Alexander-Makaryev:
Created a configuration for CI on Windows with Azure Pipeline.
Example of run can be seen on this branch.

### DIFF
--- a/hpat/tests/test_io.py
+++ b/hpat/tests/test_io.py
@@ -1,5 +1,5 @@
 import unittest
-import sys
+import platform
 import pandas as pd
 from pandas.api.types import CategoricalDtype
 import numpy as np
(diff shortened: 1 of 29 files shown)


Alexander-Makaryev:
Last commit - it is not new fail, I just missed one while marked them.

hpat/tests/test_io.py
```diff
@@ -1,4 +1,5 @@
 import unittest
+import sys
```
dmitrii-zagornyi: Instead of sys.platform use platform.system(). This value is more predictable.

shssf: @dmitrii-zagornyi Good catch!
"The output of sys.platform and os.name are determined at compile time. platform.system() determines the system type at run time."
https://stackoverflow.com/a/11674977/4428424

Alexander-Makaryev: done

PR #125, opened 2019-08-16
Implemented iterator for pandas.Series

akharche:
The first version of iterator for pandas.Series type:
 - Created separate SeriesIterator class
 - Implemented methods getiter and setiter depending on array type storing in Series
 - Created unit tests for this functionality


hpat/tests/test_series.py
```diff
@@ -1485,6 +1485,30 @@ def test_impl():
         hpat_func = hpat.jit(test_impl)
         np.testing.assert_array_equal(hpat_func(), test_impl())
 
+    def test_series_iterator_int(self):
+        def test_impl():
+            A = pd.Series([1, 2, 3, 4, 5])
```
densmirn: What do you think about constructing Series with unsorted list?

hpat/tests/test_series.py
```diff
@@ -1485,6 +1485,30 @@ def test_impl():
...
+
+    def test_series_iterator_string(self):
+        def test_impl():
+            A = pd.Series(['a', 'ab', 'abc', '', 'dddd'])
+            return [i for i in A]
+
+        hpat_func = hpat.jit(test_impl)
+        np.testing.assert_array_equal(hpat_func(), test_impl())
+
+    def test_series_iterator_one_value(self):
```
shssf: Does it make sense to add something like:
```
    def test_series_iterator_one_value(self):
        def test_impl(A):
            return [i for i in A]

        A = pd.Series([5])
        np.testing.assert_array_equal(hpat_func(A), test_impl(A))
```

shssf:
@akharche Did you look for a code style for this by `python ./setup.py style`?

hpat/tests/test_series.py
kozlov-alexey: @akharche Should we add tests with other types, like float and boolean as well? Will they be handled correctly now?

hpat/tests/test_series.py
kozlov-alexey: @akharche We may also check that empty series is handled correctly, by supplying 
A = pd.Series([np.int64(x) for x in range(0)])

akharche:
@shssf Code style checks are ok for the changes

hpat/tests/test_series.py
akharche: @densmirn I think, it doesn't matter in this case

akharche: @kozlov-alexey Yes, I am going to add by the next commit

hpat/tests/test_series.py
akharche: @shssf I think yes, it is also important to check iteration over Series as a parameter. I will add this.

akharche: @kozlov-alexey Yes, I will add

PR #126, opened 2019-08-26
groupby.median

Hardcode84:
My attempt to fix `groupby.median`.

`groupby` uses functions already defined for `series` but run additional pass `agg_distributed_run` to parallelize(?) them. This pass parses original python function and expects very specific function structure (see `_column_mean_impl` for supported example). `series.median` just calls `hpat.hiframes.api.median` and pass cannot process it.

I tried to write separate path in `agg_distributed_run` to skip all parallelization logic and just call provided function but it is incomplete and doesn't work yet.


Hardcode84:
@shssf @kozlov-alexey FYI

shssf:
@Hardcode84 Is this PR is still in work in progress stage? Is it ready for merge?

fschlimb:
Before merge, please trigger CI with an empty commit or alike.

Hardcode84:
still not ready, no updates so far. I will close it for now

PR #127, opened 2019-08-27
Fix for head_index_parallel test failing if executed twice and with MPI

kozlov-alexey:


kozlov-alexey:
@shssf OK, next time I won't. I just started fixing newly introduced issues and forgot to stop. :)

hpat/distributed_api.py
```diff
@@ -195,15 +196,14 @@ def gatherv_str_arr_impl(data):
 
             # displacements
             all_data = StringArray([''])  # dummy arrays on non-root PEs
-            displs = np.empty(1, np.int32)
-            displs_char = np.empty(1, np.int32)
+            displs = np.empty(n_loc, np.int32)
+            displs_char = np.empty(n_loc, np.int32)
```
fschlimb: `displs` and `displs_char` are needed on root only and will be set properly in the following if-condition.
This is just a dummy for all other ranks to keep the code simple.

A proper solution would make `c_gatherv` accept `None` or `0` for non-root ranks. This might require some extra work like casting...

BTW: `n_loc` is definitely not the right value anyway.

hpat/distributed_api.py
```diff
@@ -330,30 +330,36 @@ def const_slice_getitem(arr, slice_index, start, count):
 
 @overload(const_slice_getitem)
 def const_slice_getitem_overload(arr, slice_index, start, count):
+    # slice.index start/stop contain absolute indexes of the slice in input array
+    # start/count define part of the array processed by this processor
+
+    # TODO: should this also handle slices not staring from zero?
```
fschlimb: Could you convert his into a proper doc-string?
What is this functions supposed to do? ...

hpat/tests/test_series.py
```diff
@@ -872,14 +859,38 @@ def test_impl(S):
         self.assertEqual(count_array_REPs(), 0)
         self.assertEqual(count_parfor_REPs(), 0)
 
+    def test_series_dist_input2(self):
+        def test_impl(S):
```
fschlimb: Pls add a doc-string saying what it's testing.

hpat/tests/test_series.py
```diff
@@ -872,14 +859,38 @@ def test_impl(S):
...
+        hpat_func = hpat.jit(distributed={'S'})(test_impl)
+
+        n = 111
+        S = pd.Series(np.arange(n), 1 + np.arange(n))
+        start, end = get_start_end(n)
+        self.assertEqual(hpat_func(S[start:end]), test_impl(S))
+        self.assertEqual(count_array_REPs(), 0)
+        self.assertEqual(count_parfor_REPs(), 0)
+
+    def test_series_dist_input3(self):
```
fschlimb: Pls add a doc-string saying what it's testing.

hpat/distributed_api.py
kozlov-alexey: @fschlimb Yes, I left it accidentally, it's not needed at all. I can change it to 0 size array, since the size doesn't really matter, or we can even remove these two lines as redundant. I'm not sure I got how to update c_gatherv. It accepts void*, but in gatherv_overload we have to keep the type consistent, so if we assign something to displs it it should have the same type as in the root branch. Shall I remove these lines?

hpat/distributed_api.py
```diff
@@ -330,30 +330,46 @@ def const_slice_getitem(arr, slice_index, start, count):
...
         reduce_op = Reduce_Type.Sum.value
 
         def getitem_str_impl(arr, slice_index, start, count):
             rank = hpat.distributed_api.get_rank()
             k = slice_index.stop
+
             # get total characters for allocation
             n_chars = np.uint64(0)
-            if k > count:
+            if k > start:
```
densmirn: Looks like the code from if-statement is also valid for else-statement. I mean we can avoid if-else branching here and leave only code from the if-statement.

kozlov-alexey: @densmirn That's true, thank you. But I think keeping if statement simplifies code. I will remove the redundant "max(k - start, 0)" in the first branch though, i.e. will make it this:

            if k > start:
                # if slice end is beyond the start of this subset we have to send our elements
                my_end = min(count, k - start)
                my_arr = arr[:my_end]
            else:
                my_arr = arr[:0]

hpat/distributed_api.py
fschlimb: I don't think you can remove them because you need something type-stable to pass it to `c_gatherv`. If a zero-sized array works that'd be great.

If numba does not automatically cast 0 to a pointer when the C-function expects a void* you can write an intrinsic which returns a `((void*)0)` and use that in the function you @jit/@overload. Maybe like this
```
@intrinsic
def mk_int32_null(typingctx):
    def codegen(context, builder, sig, args):
        zero = context.get_constant(types.int32, 0)
        return builder.inttoptr(zero, llvmlite.ir.IntType(32))
    return numba.types.int32, codegen
```

It's probably not worth the effort, though. You will probably need yet another layer between the caller and the ExternFunction. In particular if passing an empty array works we should just use that.

kozlov-alexey: It looks like else branch can be remove (I tried and test still passes). Inserting "print(my_arr)" into it shows empty array on non root ranks (i.e. it's not even needed to assign something in the else branch), looks like Numba default constructs the object, once it's able to infer the type? I failed to find where the magic comes from in the IR. I suggest keeping latest commit with zero size array which works too.

fschlimb:
manually triggered azure succeeded: https://dev.azure.com/IntelPython/HPAT/_build/results?buildId=45

PR #128, opened 2019-08-29
Add 3 performance test cases for unicode strings

densmirn:
Add initial versions of 3 performance test cases for unicode strings.


fschlimb:
why is unicode performance testing in HPAT? I presume numba can operate on arrays of unicode strings.

PR #129, opened 2019-08-30
rearranged documentation

rdesai16:
This PR adds updated sphinx setup to enable automated documentation


fschlimb:
We need a way to view the result before accepting/merging, not only the sources. Any idea?

rdesai16:
@fschlimb I am not sure how to do that. Maybe downloading these specific changes from my branch can help. But I will try to find another way and will keep you posted.

PR #130, opened 2019-08-30
Add support for DataFrame.median method

kozlov-alexey:


hpat/hiframes/pd_dataframe_ext.py
```diff
@@ -1259,6 +1259,39 @@ def lower_mean_dummy(context, builder, sig, args):
...
+        return hpat.hiframes.pd_dataframe_ext.median_dummy(df)
+
+    return _impl
+
+
+def median_dummy(df, n):
+    return df
+
+
+@infer_global(median_dummy)
```
shssf: Are you sure it is needed? Typing should be done in first part of `overload_method` (lines 1264 in he PR)

hpat/hiframes/pd_dataframe_ext.py
```diff
@@ -1259,6 +1259,39 @@ def lower_mean_dummy(context, builder, sig, args):
...
+        # output is float64 series with column names as string index
+        out = SeriesType(
+            types.float64, types.Array(types.float64, 1, 'C'),
+            string_array_type)
+        return signature(out, *args)
+
+
+@lower_builtin(median_dummy, types.VarArg(types.Any))
+def lower_median_dummy(context, builder, sig, args):
+    out_obj = cgutils.create_struct_proxy(
```
shssf: Did you try to move this into line 1267?

hpat/hiframes/pd_dataframe_ext.py
fschlimb: Isn't this for the dummy while the above is for the actual median method?

hpat/hiframes/pd_dataframe_ext.py
fschlimb: See comment in line 1265: with https://github.com/numba/numba/pull/4208 we can now simplify all this and get rid of the dummy construct. Maybe this should be done in an extra PR converting all the dummies in this file.

shssf: @fschlimb Could you please point exact link to the material or just copy part of it here to help to find it?

fschlimb: @shssf What material are you asking for? I think all I know is this PR...
As far as I understand one issue with using `@overload` is that (before the above PR to numba) the code in the `@overload` could not be inlined by numba. This in turn inhibits optimizations that numba/llvm can do with the current code (like loop-fusion). With `@overload(inline=always)` (or so) numba will now inline the code from an `@overload` and so loops over several series or several loops over the same series can potentially get fused. Such compiler opts are a major source of HPAT's performance advantages.

shssf:
@kozlov-alexey Is this PR ready for merge or still in work in progress stage?

fschlimb:
@kozlov-alexey could you push an empty commit or alike to force CI to start again?

kozlov-alexey:
@shssf @fschlimb 
I just pushed it. Shall we merge it as is without removing dummy?

I tried to remove the dummy and was able to use the hpat.hiframes.api.median on the series inside _impl itself, but it requires hpat.hiframes.api.get_series_data to be used, and the latter has the same limitations as DataFrame.getitem, i.e. it can be used with literal column names only. Here's the snippet that compiles (but the test fails of course):

```
@overload_method(DataFrameType, 'median')
def median_overload(df, axis=None, skipna=None, level=None, numeric_only=None):
    # TODO: kwargs
    # TODO: avoid dummy and generate func here when inlining is possible

    saved_columns = df.columns

    def _impl(df, axis=None, skipna=None, level=None, numeric_only=None):

#         for col_name in saved_columns:
#             print(5 * "*", col_name, 5 * "*") # this works 
#             print(hpat.hiframes.api.get_series_data(df[col_name])) # this fails, due to DataFrame getitem limitations

        series_data = hpat.hiframes.api.get_series_data(df['A'])
        median_value = hpat.hiframes.api.median(series_data)
        data = np.array([median_value for _ in np.arange(len(saved_columns))])
        index = hpat.str_arr_ext.StringArray(saved_columns)
        return hpat.hiframes.api.init_series(data , index)

        # above would work if DataFrame getitem worked with non static values as column names

    return _impl
```

PR #131, opened 2019-09-03
Fix for Series.head method

akharche:
Series.head() fails in distribution when index is None (not explicitly set)

- As a fix added index setting by numpy.arange to head func call
- Enabled skipped tests


hpat/hiframes/hiframes_typed.py
```diff
@@ -2277,6 +2271,19 @@ def _get_series_index(self, series_var, nodes):
         nodes += f_block.body[:-2]
         return nodes[-1].target
 
+    def _get_index_values(self, series_data, nodes):
```
fschlimb: Pls add a docstring.

hpat/tests/test_series.py
```diff
@@ -1265,11 +1261,7 @@ def test_impl(S):
         S = pd.Series(np.random.randint(-30, 30, m))
         np.testing.assert_array_equal(hpat_func(S).values, test_impl(S).values)
 
-    @unittest.skip('numba.errors.TypingError - fix needed\n'
```
kozlov-alexey: Please unskip test_series_head_parallel1 and test_series_rename1 tests too - they were failing due to the same reason (or if they still fail change the skip message).

akharche: test_series_rename1 is not connected to head method, but problem is the same: distributing None index. The fix should be applied to all the functions where index is used somehow in order to unskip the rest of the tests. May be, there is a better solution to control the index value in common place.

PR #134, opened 2019-09-05
Add Numba decorator use option

shssf:
Currently controled by `HPAT_CONFIG_PIPELINE_HPAT=1`. Planned to use in testing purposes.

PR #135, opened 2019-09-05
Pandas Series operators implementation.

shssf:
PR #134 needs to merge first


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -0,0 +1,68 @@
...
+    Algorithm: result = series[idx]
+    Where:
+        series: pandas.series
+        idx: integer number
+    Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_static_getitem_series1
+    '''
+
+    if not isinstance(self, HPATPandasSeriesType):
+        raise numba_TypingError('The object must be a pandas.series')
+
```
fschlimb: In other places overloads return nothing when meeting types that they are not handling (like here: https://github.com/numba/numba/blob/60d2bddd0a9ddf84cf900a70e4b7e17bcbe5dc87/numba/unicode.py#L492). I think we should follow this practice and not throw exceptions. Numba will then take care of error reporting (exceptions thrown here will be lost anyway).

shssf: https://github.com/numba/numba/blob/60d2bddd0a9ddf84cf900a70e4b7e17bcbe5dc87/numba/unicode.py#L694

fschlimb: ljust is not a standard operator!

PR #136, opened 2019-09-06
Add tests create dataframe

akharche:
Tests on [issue](https://github.com/IntelPython/hpat/issues/132)


hpat/tests/test_dataframe.py
```diff
@@ -50,6 +50,15 @@ def test_impl(A, B, c):
         c = 2
         pd.testing.assert_series_equal(hpat_func(A, B, c), test_impl(A, B, c))
 
+    @unittest.skip('Implement feature to create DataFrame without column names')
```
shssf: Why it is skipped? Work is in progress?

PR #137, opened 2019-09-06
Add perf test on Series.str.len

densmirn:
Added 1 performance test case on Series.str.len.
Implemented generator of string series.
Fixed numba build for Windows in ASV config.


Alexander-Makaryev:
The random functionality is used in the data generation but I can't find where seed for the RNG is determined. Please determine the seed and ensure that results of work are reproducible.

PR #141, opened 2019-09-11
Make Travis use conda-build

Vyacheslav-Smirnov:
- Make Travis use conda-build
 - Add build for pure python without MPI (using HPAT_CONFIG_MPI var)
 - Add python version define
 - Move coverage to run_test.sh


buildscripts/hpat-conda-recipe/meta.yaml
```diff
@@ -1,12 +1,19 @@
+{% set version = "0.29.2" %}
+{% set buildnumber = 1 %}
+
 package:
   name: hpat
-  version: {{ GIT_DESCRIBE_TAG }}
+  version: {{version}}
```
fschlimb: what's wrong about the git-way?

Vyacheslav-Smirnov: For some reasons conda says that these vars are not defined

PR #145, opened 2019-09-12
bumping version to 0.30.0

fschlimb:
@Vyacheslav-Smirnov @shssf


shssf:
@fschlimb Is this PR is still relevant?

PR #151, opened 2019-09-13
[WIP] This PR shows an example of API docstring for distributed.py module

rdesai16:
This is just an example of how the module and the classes inside that module would look like with Numpy style docstrings


rdesai16:
@shssf In this PR I have submitted distributed.py example. I have added a standardized way of how modules, classes and fuctions should be documented for autodoc. Note that now anywhere in documentation if you would like to reference one of these function or classes you can do that by adding :func: or :class: in front of the func/class name. For example, if you want to refer to _find_first_print function from distributed.py in any of the .rst file you can use :func:`hpat.distributed._find_first_print` and it will link you to the documentation for this function.

rdesai16:
I am still working on doing the same for distributed_api.py. Meanwhile the outlook of the docstring here can be reviewed.

hpat/distributed.py
```diff
@@ -73,10 +79,38 @@
...
+        
+        Note
+        ----
+        Do not include the `self` paramter in the ``Parameters`` section
+        
+        Parameters
+        -----------
+        func_ir : str
+             description of `func_ir`.(optional) you can mention supported or not here
+        typingctx : TypeForThisParam
```
shssf: Can we add a link to external site? I mean, need HTML link example.

docs/source/index.rst
```diff
@@ -11,4 +11,4 @@ HPAT documentation
    DevelopersGuide
    Modules
 
-To download HPAT documentation PDF click here :download:`pdf <../_build/latex/HPAT.pdf>`
\ No newline at end of file
+To download HPAT documentation PDF click here :download:`pdf <../_build/latex/HPAT.pdf>`
```
shssf: it will not work on Linux (we have 'build' directory there instead '_build'

hpat/distributed.py
```diff
@@ -73,10 +79,38 @@
...
+    """
 
     def __init__(self, func_ir, typingctx, targetctx, typemap, calltypes,
                  metadata):
+        """ Here are moer details about the init method or the class 
+       
+        The details can be multiple lines. You can mention all the supported
+        and not supported args here or you can do it in ``Parameters`` section
+        
+        Note
```
shssf: Is it possible to extract `Note` and other sections into one documentation page and only `Note` to an other documentation page? I mean a differences between `user` and `developer` documentation we discussed off line.

docs/source/index.rst
fschlimb: `_build` is the folder that sphinx creates (not the one from setuptools). There is no difference between OSes.

hpat/distributed.py
rdesai16: Yes you can. The same way you do it in rst file.  For an example I tried to add link to Pandas documentation and it worked.

hpat/distributed.py
rdesai16: I couldn't find something like that in Sphinx. @fschlimb do you know if Sphinx has a way to do this? If Sphinx can't do that then I'll have to write a parsing script for that.

fschlimb: What about
https://sublime-and-sphinx-guide.readthedocs.io/en/latest/conditions.html
or
https://www.sphinx-doc.org/en/master/usage/extensions/ifconfig.html
?

rdesai16: This is interesting! I will try it out. Thanks @fschlimb!

docs/source/index.rst
shssf: @fschlimb In this case this directory should be under `build`. We don't need several temporary directories in top project directory.

rdesai16: We don't need to use the pdf in _build directory. We can copy the pdf in other directory and use that link. But it should be mentioned explicitly that if someone has added changes to documentation and built it then he/she needs to build the pdf and update it in the location mentioned in the link.

fschlimb: Hm, the building sphinx docs is a separate process, it typically happens in the `docs` directory, not top level. It is probably possible to change default behavior. I don't think it's important, though.

In any case, in this extra process, sphinx creates a directory (e.g. `_build`) where it puts its output. The `html` directory in there is then what gets 'manually' copied to wherever  the webserver expects to see the files. I would expect (but I don't know, so you should check) that at build time sphinx will copy files referenced in the output dir to the html dir. Otherwise our 'manual' copy procedure needs to include the (path to) the pdf so that the webserver finds it.

docs/devsource/conf.py
```diff
@@ -65,7 +67,7 @@
 
 #autosummary_generate = True
 autosectionlabel_prefix_document = True
-autodoc_mock_imports = ['numba', 
+'''autodoc_mock_imports = ['numba',
```
shssf: Is it commented out and the code is not needed any more? If it is not needed, maybe it is better to delete it?

hpat/Pandas/Series.py
```diff
@@ -34,38 +41,28 @@
 from hpat.hiframes.pd_series_ext import SeriesType
 
 
-'''
-Pandas Series (https://pandas.pydata.org/pandas-docs/stable/reference/series.html)
-functions and operators definition in HPAT
-Also, it contains Numba internal operators which are required for Series type handling
-
-Implemented operators:
-    at
```
shssf: Are you sure we don't need "implemented operators" list anymore?

docs/devsource/conf.py
rdesai16: Actually I left it commented because it prevented all those import errors I was getting from autodoc before building hpat. It had taken some time to create this list of mock imports and in future if I get these errors again it will be easier to debug with this list. But now that I think of I'll just remove this code and save this list somewhere else. We don't want to clutter up the actual file with all this.

hpat/Pandas/Series.py
rdesai16: As I understood from our last conversation with Sergey Maidanov it was implied that we should remove this list from here. Because we are providing a list of implemented operators below in the file along with the summary lines.

PR #154, opened 2019-09-16
Add class name as part of benchmark name

densmirn:


fschlimb:
@densmirn please investigate and fix CI failures

PR #157, opened 2019-09-17
series.take()

densmirn:


shssf:
@densmirn `Assignees` means people who will work on the PR.

hpat/tests/test_series.py
```diff
@@ -1595,5 +1595,32 @@ def test_impl():
         np.testing.assert_array_equal(hpat_func(), test_impl())
 
 
+class TestSeriesTake(unittest.TestCase):
```
shssf: Please follow common rules to create tests in this file

hpat/hiframes/pd_series_ext.py
```diff
@@ -482,13 +482,6 @@ def resolve_sort_values(self, ary, args, kws):
         out = SeriesType(ary.dtype, ary.data, out_index)
         return signature(out, *args)
 
-    @bound_function("array.take")
```
shssf: This is signature definition of the function (like declaration in C). Need to find and remove implementation of the algorithm (like definition in C)

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -514,3 +514,42 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+        raise TypingError('{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if self.index == types.none:
+        raise TypingError('{} The object must NOT be a None. Given self.index: {}'.format(_func_name, self.index))
+
+    if not isinstance(indices, types.List):
+        raise TypingError('{} The object must be a List. Given self: {}'.format(_func_name, self))
+
+    def hpat_pandas_series_take_impl(self, indices):
+        return pandas.Series(self._data.take(indices), index=[self._index[i] for i in indices])
```
shssf: Why we need new list `self._index[i] for i in indices` here? Could we use `self._index` instead?

densmirn: We need to take elements from `self._index` according to `indices`. E.g.:
`data = [1.0, 13.0, 9.0, -1.0, 7.0]`
`index = [0, 1, 2, 3, 4]`
`indices = [1, 3]`

The result of `take()`:
`data = [13.0, -1.0]`
`index = [1, 3]`

I can't imagine solution w/o creation of new list.

hpat/hiframes/pd_series_ext.py
densmirn: As I see implementation is provided by Numba.

hpat/tests/test_series.py
densmirn: No problem!

hpat/hiframes/pd_series_ext.py
shssf: Could you please point me Numba code there this method implemented?

hpat/datatypes/hpat_pandas_series_functions.py
shssf: Could you please move this loop out of the line with `return` statement? I mean, create temporal value with meaningfull name, initialize it by this loop and put this variable into the line with return statement to `index`

hpat/hiframes/pd_series_ext.py
densmirn: https://github.com/numba/numba/blob/c9b4355446e630301f2551fbd032dd56bec9ebce/numba/targets/arrayobj.py#L3616-L3675

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Of course!

hpat/hiframes/pd_series_ext.py
shssf: I see many other methods for type `array` in this file. Are you sure this type is NumPy array? I don't understand why we need this signature definition here. I see it converts output from `array.take` to `pd.Series`. Do you have an idea why? Do we have failed tests if this signature removed?

hpat/tests/test_series.py
```diff
@@ -1622,6 +1618,39 @@ def test_impl():
         hpat_func = hpat.jit(test_impl)
         np.testing.assert_array_equal(hpat_func(), test_impl())
 
+    def test_series_take_index_default(self):
+        def pyfunc():
+            series = pd.Series([1.0, 13.0, 9.0, -1.0, 7.0])
+            indices = [1, 3]
```
densmirn: Could you please tell why we don't want to test unboxing stage? I mean why don't we pass `series` and `indices` to the function as arguments?

shssf: Testing of the list boxing functionality is the purpose of other specialized tests (in general, in good test system). But at current stage it might be better to have both variants here. I changed tests to localize an issue in development.

shssf: I would vote for having functionality tests for both, in JITted and out of JITted regions

PR #158, opened 2019-09-18
[Tests] Introduce tests for Numba limitations

AlexanderKalistratov:
Introduce tests for current numba limitations:

* jitclass with Dict
* Infer class members from init for jitclass
* list.sort(key=...)
* sorted(a, key=...)
* iterate over tuple
* try/except/raise
* defaultdict

### DIFF
--- a/hpat/tests/test_hpat_jit.py
+++ b/hpat/tests/test_hpat_jit.py
@@ -1,0 +1,142 @@
+import unittest
+import hpat
+import numba
+from hpat import *
+from numba.typed import Dict
+from collections import defaultdict
+
+class TestHpatJitIssues(unittest.TestCase):
+
+    @unittest.skip("Dict is not supported as class member")
+    def test_class_with_dict(self):
+        @jitclass([('d', Dict)])
+        class ClassWithDict:
+            def __init__(self):
+                self.d = Dict.empty(key_type=int32, value_type=int32)
+
+        @numba.njit
+        def test_impl():
+            c = ClassWithDict()
+
+            c.d[0] = 1
+
+            return c.d[0]
+
+        test_impl()
+
+    @unittest.skip("Type infer from __init__ is not supported")
+    def test_class_from_init(self):
+        @jitclass()
+        class ClassWithInt:
+            def __init__(self):
+                self.i = 0
+
+        @numba.njit
+        def test_impl():
+            c = ClassWithInt()
+
+            print(c.i)
+
+        test_impl()
+
+    @unittest.skip("list.sort with lambda is not supported")
+    def test_list_sort_lambda(self):
+        @numba.njit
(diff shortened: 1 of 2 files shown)


hpat/tests/test_numba.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_list_sort(self):
+        @numba.njit
+        def sort_with_list():
+            l = [5, 4, 3, 2, 1, 0]
+
+            l.sort(key=lambda x: x)
+
+            return l
+
+        func_with_list()
```
shssf: Did you mean `sort_with_list`?

hpat/tests/test_numba.py
```diff
@@ -0,0 +1,101 @@
+import unittest
+import hpat
+import numba
+from hpat import *
+from numba.typed import Dict
+from collections import defaultdict
+
+class TestNumba(unittest.TestCase):
```
shssf: Could you please rename it to something like `HPAT_Jit_issues`?

hpat/tests/test_numba.py
AlexanderKalistratov: fixed

hpat/tests/test_numba.py
AlexanderKalistratov: Renamed

AlexanderKalistratov:
Added defaultdict test

hpat/tests/test_hpat_jit.py
```diff
@@ -0,0 +1,142 @@
...
+import numba
+from hpat import *
+from numba.typed import Dict
+from collections import defaultdict
+
+class TestHpatJitIssues(unittest.TestCase):
+
+    @unittest.skip("Dict is not supported as class member")
+    def test_class_with_dict(self):
+        @jitclass([('d', Dict)])
```
sklam: This is failing because `Dict` is not treated as a type.  It needs to be spelled as `@jitclass([('d', numba.types.DictType(types.intp, types.float64))])`.  The dictionary type has to be fully specialized. See example in https://gist.github.com/sklam/945c3304c4b132127df377fa317f5ae5.

hpat/tests/test_hpat_jit.py
```diff
@@ -0,0 +1,142 @@
...
+        func_with_try_except()
+
+    @unittest.skip("raise is not supported")
+    def test_with_raise(self):
+        @numba.njit
+        def func_with_raise(b):
+            if b:
+                return b
+            else:
+                raise "error"
```
sklam: I just noticed that this is an error in pure-python, which will show

```
TypeError: exceptions must derive from BaseException
```

A valid and numba-compatible version will be:

```python
raise ValueError("error")
```

PR #160, opened 2019-09-18
fixing unboxing of categoricals in DFs

fschlimb:
@anmyachev

### DIFF
--- a/hpat/tests/test_dataframe.py
+++ b/hpat/tests/test_dataframe.py
@@ -122,6 +122,18 @@ class TestDataFrame(unittest.TestCase):
         df = pd.DataFrame({'A': ['aa', 'bb', 'cc']})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_box_categorical(self):
+        def test_impl(df):
+            df['A'] = df['A'] + 1
+            return df
+
+        hpat_func = hpat.jit(test_impl)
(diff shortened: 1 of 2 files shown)


hpat/tests/test_dataframe.py
```diff
@@ -122,6 +122,18 @@ def test_impl(df):
         df = pd.DataFrame({'A': ['aa', 'bb', 'cc']})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_box_categorical(self):
+        def test_impl(df):
+            df['A'] = df['A'] + 1
+            return df
+
+        hpat_func = hpat.jit(test_impl)
+        from pandas.api.types import CategoricalDtype
```
shssf: Do we really need this import inside the function? If not, I would prefer to see it at the top of file.

fschlimb: We do not need this symbol anywhere else, so why expose it at global scope? Why import unless needed in the first place? I think this is cleaner (a potential minor performance impact if called multiple times can be ignore here).

shssf: @fschlimb Not importing files at top of the file is a PEP8 violation. We need visible reason ("why do it in regular way" is not acceptable) to do this.

fschlimb: Then no import...

PR #163, opened 2019-09-20
Added operators pow, mod, lt, gt, le, ge, eq

1e-to:
Added operators pow, mod, lt, gt, le, ge, eq and tests for comparison operators.


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -469,3 +476,332 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+            """
+            Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op5_integer_scalar
+            Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op5_float_scalar
+            """
+
+            return pandas.Series(lhs._data ** rhs)
+
+        return hpat_pandas_series_pow_impl
+
+    raise TypingError('{} The object must be a pandas.series or scalar. Given rhs: {}'.format(_func_name, rhs))
```
shssf: The object must be pandas.series and argiment must be a number. Given lhs:... rhs...

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -469,3 +476,332 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+            """
+            Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op5_integer_scalar
+            Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op5_float_scalar
+            """
+
+            return pandas.Series(lhs._data % rhs)
+
+        return hpat_pandas_series_mod_impl
+
+    raise TypingError('{} The object must be a pandas.series or scalar. Given rhs: {}'.format(_func_name, rhs))
```
shssf: Change this as proposed above

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -469,3 +476,332 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+        fill_value: unsupported
+              axis: unsupported
+            result: pandas.series
+
+    Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op8
+    """
+
+    _func_name = 'Method gt().'
+
+    if not isinstance(lhs, SeriesType):
```
shssf: Original function has many extra arguments `gt(self, other, level=None, fill_value=None, axis=0)`
Could you please implement a checks for them?
I mean is these unsupported parameters are not equals to defaults - throw exception on `Unsupported parameters`
This is not implemented before in this module.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -501,6 +501,9 @@ def hpat_pandas_series_pow(lhs, rhs):
     if not isinstance(lhs, SeriesType):
         raise TypingError('{} The object must be a pandas.series. Given lhs: {}'.format(_func_name, lhs))
 
+    if level is not types.none or fill_value is not types.none or axis is not 0:
+        raise TypingError('Unsupported parameters {}'.format(_func_name))
```
shssf: Unsupported parameters `in` {}

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -522,10 +525,10 @@ def hpat_pandas_series_pow_impl(lhs, rhs):
 
         return hpat_pandas_series_pow_impl
 
-    raise TypingError('{} The object must be a pandas.series or scalar. Given rhs: {}'.format(_func_name, rhs))
+    raise TypingError('{} The object must be a pandas.series and argument must be a number. Given rhs: {}'.format(_func_name, rhs))
```
shssf: Thanks but we still have only `rhs` type printed here. It would be good to add `lhs` type also to see full information

densmirn:
> As far as I remember PIP8 requires using `is not` in type comparisons instead `!=`
I think operator choice depends on comparable objects. Operator "is not" compares objects addresses in memory (id() in Python). Operator "!=" compares values of objects.

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: @shssf I realized check for unsupported parameters, but it has status of change requested. Should I write different checks for any arguments?

hpat/tests/test_series.py
```diff
@@ -556,6 +556,37 @@ def test_series_op8(self):
...
+            n = 11
+            operand_series = pd.Series(np.arange(1, n))
+            operand_scalar = 10
+            pd.testing.assert_series_equal(
+                hpat_func(operand_series, operand_scalar),
+                test_impl(operand_series, operand_scalar),
+                check_names=False)
+
+    def test_series_op8_float_scalar(self):
+        comparison_methods = ('lt', 'gt', 'le', 'ge', 'ne', 'eq')
```
akharche: Please, remove ne method from the test or implement functional to make 'ne' with scalars

shssf: @akharche What wrong with `ne` method here? Just for better understanding

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -40,6 +40,34 @@
 
 from hpat.hiframes.pd_series_ext import SeriesType
 
+'''
+Pandas Series (https://pandas.pydata.org/pandas-docs/stable/reference/series.html)
```
shssf: Sorry but we changed documentation style. This section is not presented anymore here.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -482,3 +510,389 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+@overload_method(SeriesType, 'pow')
+def hpat_pandas_series_pow(lhs, rhs, level=None, fill_value=None, axis=0):
+    """
+    Pandas Series method 'pow' implementation.
+        https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.pow.html#pandas.Series.pow
+
+    Algorithm: result = lhs.pow(other, level=None, fill_value=None, axis=0)
+
+    Where:
+               lhs: pandas.series
```
shssf: Sorry but we changed documentation style. We have better formatting for docstrings now. Could you please apply new format style for all documentation?

hpat/datatypes/hpat_pandas_series_functions.py
shssf: @1e-to I see you already implemented checks for all function parameters. Thank you.

hpat/tests/test_series.py
akharche: @shssf Test fails because we can compare only series and series (lhs._data != rhs._data). So, I suggest to implement the same functionality for numbers in 'ne' method (ex. lhs._data != rhs) or remove 'ne' from the test for this pr

hpat/tests/test_series.py
```diff
@@ -545,7 +545,7 @@ def test_series_op7(self):
             pd.testing.assert_series_equal(hpat_func(A, B), test_impl(A, B), check_names=False)
 
     def test_series_op8(self):
-        comparison_methods = ('lt', 'gt', 'le', 'ge', 'ne', 'eq')
```
shssf: I would propose to do not remove this test but add integers algo branch into `ne()` method in this PR

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -482,3 +506,424 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
...
+
+@overload_method(SeriesType, 'pow')
+def hpat_pandas_series_pow(lhs, rhs, level=None, fill_value=None, axis=0):
+    """
+    Pandas Series method :meth:`pandas.Series.pow` implementation.
+    .. only:: developer
+
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_op5
+
+    Parameters
```
densmirn: Please add description for parameter "rhs" here and for other implementations in this commit.

shssf: Nice catch!
Also, please use supported types list as it presented in example here https://github.com/IntelPython/hpat/blob/master/hpat/datatypes/hpat_pandas_series_functions.py#L57

PR #167, opened 2019-09-24
Change tests failed due to int64 vs int32 dtype mismatch

kozlov-alexey:


kozlov-alexey:
@shssf I agree but I don't see any solution on our side here. The issue is in difference between Numba and NumPy in a way they infer integer types. The tests will pass you we change int typing login in:

```
..\numba\typing\typeof.py:
def _typeof_int(val, c):
    # As in _typeof.c
    nbits = utils.bit_length(val)
    if nbits < 32:
        typ = types.intp                       #  <------     it converts to int64, change to types.int32 and there'll be no fails 
```

But as far as I understand from conversation with Numba guys, it's not an option, as it was intentional design. See:  http://numba.pydata.org/numba-doc/latest/proposals/integer-typing.html). What other options can we try?

kozlov-alexey:
@shssf @fschlimb @Hardcode84
As a short term solution can we maybe change tests in a different way, i.e. instead of using check_dtype=False, specify it (for affected test only) when creating Series/DFs, i.e.
```
-            df = pd.DataFrame({'A': np.arange(1, n)})
-            pd.testing.assert_series_equal(hpat_func(df.A, 1), test_impl(df.A, 1), check_dtype=False, check_names=False)
+            df = pd.DataFrame({'A': np.arange(1, n)}, dtype=USE_EXACT_DTYPE)
+            pd.testing.assert_series_equal(hpat_func(df.A, 1), test_impl(df.A, 1), check_names=False)
```
where
USE_EXACT_DTYPE = 'int64' if platform.system() == 'Windows'and not IS_32BITS else 'int32'

What do you think?

hpat/tests/test_dataframe.py
```diff
@@ -332,30 +333,32 @@ def test_impl(df, n):
...
+        if platform.system() == 'Windows' and not IS_32BITS:
+            # set existing column
+            def test_impl(n):
+                df = pd.DataFrame({'A': np.ones(n, np.int64), 'B': np.arange(n) + 3.0})
+                df['A'] = np.arange(n, np.int64)
+                return df
+        else:
+            # set existing column
+            def test_impl(n):
+                df = pd.DataFrame({'A': np.ones(n, np.int64), 'B': np.arange(n) + 3.0})
```
densmirn: I'm not sure, but is `np.int64` exactly needed here?

kozlov-alexey: @densmirn In which place? In dataframe creation it's not needed - yes, only in the assignment later. But it turns out that np.arange cannot be jitted with dtype arg anyway, so I'm re-writing this at the moment.

PR #168, opened 2019-09-24
Attribute values for pandas Series

akharche:


hpat/hiframes/pd_series_ext.py
```diff
@@ -403,6 +403,7 @@ def cast_series(context, builder, fromty, toty, val):
 class SeriesAttribute(AttributeTemplate):
     key = SeriesType
 
+    # PR135. This needs to be commented out
```
shssf: This PR is 168 :-)

PR #170, opened 2019-09-24
Attribute shape for pandas Series

akharche:
Attribute shape for pandas Series in new style


hpat/hiframes/pd_series_ext.py
```diff
@@ -102,6 +102,10 @@ def key(self):
         # needed?
         return self.dtype, self.data, self.index, self.is_named
 
+    @property
+    def ndim(self):
```
shssf: Sorry, didn't understand why this is needed.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -143,6 +147,31 @@ def hpat_pandas_series_iloc_impl(self):
...
+    Test:  python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_shape1
+    """
+
+    _func_name = 'Attribute shape.'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    def hpat_pandas_series_shape_impl(self):
+        return len(self._data),
```
shssf: I would say than it is better to return `self._data.shape` instead `len(self._data)` https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.shape.html

shssf:
@akharche Could you please change docstrings according to new pattern merged into sources recently?

PR #172, opened 2019-09-25
[WIP]adding fix for changing func name in display

rdesai16:


docs/rename_function.py
```diff
@@ -0,0 +1,74 @@
...
+        dst_file = os.path.join(cur_dir, "..", "API_Doc", f)
+        copyfile(src_file, dst_file)
+        
+os.chdir(src_copy_dir)
+
+# Change the function names in copied files
+i = 0
+for filename in srcfiles_srcfuncs:
+    func_list = srcfiles_srcfuncs[filename]
+    with open(filename, '+r') as fn:
```
shssf: please don't forget to close file

docs/rename_function.py
```diff
@@ -0,0 +1,74 @@
...
+# key : srcfile
+# values per key : function names that should be changed
+# If the file is already present, just add the func name in the respective values
+# Create new entry if the srcfile is not present
+srcfiles_srcfuncs = {
+    "hpat_pandas_series_functions.py" : ["hpat_pandas_series_append", "hpat_pandas_series_ne", "hpat_pandas_series_iloc"]
+}
+
+
+# Add the filenames and the parent directory
```
shssf: Maybe is better to use Python style for variable documentation? It looks particular the same as for function but below the variable.

docs/CleanRSTfiles.py
```diff
@@ -0,0 +1,15 @@
+import os
+import itertools
+from shutil import copyfile
+
+cur_dir = os.getcwd()
+api_dir = os.path.join(cur_dir, "usersource", "api")
+os.chdir(api_dir)
+for file in os.listdir(os.getcwd()):
+    with open(file, "r+") as rstfile:
```
shssf: it is better to close file explicitly

docs/rename_function.py
rdesai16: That is the advantage of using "with open". It will close the file automatically.

docs/rename_function.py
rdesai16: Do you mean the variable names should be changed here?

shssf: I meant just Python style for variable __doc__ string. It is not necessary but it might be better to get.

rdesai16: Yes sure. I will do that. I just wanted to clear that I understood your comment correctly.

docs/rename_function.py
fschlimb: I guess the 'usual' way to do this is to end a `with open('r')` clause after reading the file and open a new `with open('w')` clause for writing.

fschlimb:
What other docstrings from HPAT do we (let sphinx) use?  I see hpat being imported, what do we need it for? I understand you  generate a small fake directory/module on the side to extract the docstrings.

rdesai16:
> What other docstrings from HPAT do we (let sphinx) use? I see hpat being imported, what do we need it for? I understand you generate a small fake directory/module on the side to extract the docstrings.

We still need to import HPAT because the files in the fake directory have some modules being imported inside. So if we don't import HPAT then the autodoc throws import errors.

docs/rename_function.py
rdesai16: @fschlimb Yes that's another way. Apart from the codestyle I don't see any other difference in both approach. So @shssf if you feel more comfortable with this second approach I can change this code. What do you think?

docs/CleanRSTfiles.py
```diff
@@ -0,0 +1,15 @@
...
+
+cur_dir = os.getcwd()
+api_dir = os.path.join(cur_dir, "usersource", "api")
+os.chdir(api_dir)
+for file in os.listdir(os.getcwd()):
+    with open(file, "r+") as rstfile:
+        l = rstfile.readlines()
+        rstfile.seek(0,0)
+        rstfile.truncate(0)
+        rstfile.writelines(l[3:])
```
fschlimb: Here as well, I'd rather use 2 `with`s.

docs/rename_function.py
```diff
@@ -0,0 +1,81 @@
...
+            content = content.replace(func, display_names[i])
+            i += 1
+        fn.seek(0, 0)
+        fn.write(content)
+
+
+            
+
+        
+
```
fschlimb: could also remove empty lines?

rdesai16: Done!

rdesai16:
@fschlimb @shssf I changed the files to have two "with" statements.

PR #174, opened 2019-09-27
Method Series.quantile in new style

akharche:
Implementation: parameter q is a float value


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -514,3 +514,43 @@ def hpat_pandas_series_floordiv_number_impl(lhs, rhs):
         return hpat_pandas_series_floordiv_number_impl
 
     raise TypingError('{} The object must be a pandas.series or scalar. Given rhs: {}'.format(_func_name, rhs))
+
+
+@overload_method(SeriesType, 'quantile')
+def hpat_pandas_series_quantile(self, q=.5):
```
shssf: it might be better to mention all parameters like in Pandas `(self, q=0.5, interpolation='linear')` and just don't use (mark as unsupported) parameter `interpolation`

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1128,47 @@ def hpat_pandas_series_le_impl(self, other):
...
+    if not isinstance(interpolation, types.Omitted) and interpolation is not 'linear':
+        raise TypingError(
+            '{} Unsupported parameters. Given interpolation: {}'.format(_func_name, interpolation))
+
+    if type(q) not in supported_q_types:
+        raise TypingError(
+            '{} The parameter must be float. Given type q: {}'.format(_func_name, type(q)))
+
+    def hpat_pandas_series_quantile_impl(self, q=0.5, interpolation='linear'):
+        return np.quantile(self._data, q)
```
shssf: Do we need to use Numpy here explicitly? Could we use `self._data.quantile()` instead?

akharche: I'm not sure, because Numba just overloads np.quantile. There is no such method for array type in Numba

shssf: https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html
This is new method presented since v1.15

akharche: I mean we can only call as np.quantile(self._data, q), like in examples by provided link

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -33,8 +33,8 @@
 
 import operator
 import pandas
+import numpy as np
```
shssf: please do not rename numpy

fschlimb: This is the the standard way of importing numpy...

PR #176, opened 2019-09-29
SeriesGroupBy implementation. ::count() method as an example

shssf:


shssf:
To reproduce the issue run `HPAT_CONFIG_PIPELINE_HPAT=0 python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_groupby_count`

hpat/datatypes/hpat_pandas_seriesgroupby_types.py
```diff
@@ -0,0 +1,123 @@
...
+        models.StructModel.__init__(self, dmm, fe_type, members)
+
+
+make_attribute_wrapper(SeriesGroupByType, 'data', '_data')
+
+
+@intrinsic
+def init_seriesgroupby(typingctx, data):
+
+    def codegen(context, builder, signature, data_val):
```
sklam: All the arguments are passed in a tuple in the 4th argument. That's why you are seeing the `'tuple' object has no attribute 'type'` error.

Instead, this should be:

```python

    def codegen(context, builder, signature, args):
        [data_val] = args    # pack the arguments
        series = cgutils.create_struct_proxy(signature.return_type)(context, builder)
        series.data = data_val
        ...
```

shssf: @sklam Thank you! But after this change got following:
```
    LoweringError: Failed in nopython mode pipeline (step: nopython mode backend)
Buffer dtype cannot be buffer, have dtype: array(series(int64, array(int64, 1d, C), none, False), 1d, A)
```

hpat/datatypes/hpat_pandas_seriesgroupby_types.py
```diff
@@ -0,0 +1,123 @@
...
+        series = cgutils.create_struct_proxy(signature.return_type)(context, builder)
+        series.data = data_val
+        
+        if context.enable_nrt:
+            context.nrt.incref(builder, types.Tuple(data), series.data)
```
shssf: @sklam My previous comment could be addressed by changing `types.Tuple(data)` to `data` here

PR #178, opened 2019-09-30
Change readme for tests perf

Rubtsowa:


densmirn:
Please create PR with only changes in README.md. You need to exclude changes in performance tests from the PR.

PR #179, opened 2019-09-30
Change runner.py

Rubtsowa:
Сompilation time was not ruled out. Fixed line 147.


densmirn:
Please create PR with only changes in performance tests runner. You need to exclude changes in performance tests and README.md from the PR.

PR #184, opened 2019-10-01
Fix for CSV read failed due to \r\n on Windows

kozlov-alexey:


hpat/tests/test_io.py
```diff
@@ -363,49 +364,67 @@ def test_impl():
...
 
-    @unittest.skipIf(platform.system() == 'Windows', "error on windows")
     def test_csv_const_dtype1(self):
-        def test_impl():
-            dtype = {'A': 'int', 'B': 'float64', 'C': 'float', 'D': 'int64'}
-            return pd.read_csv("csv_data1.csv",
-                               names=dtype.keys(),
-                               dtype=dtype,
-                               )
+        if platform.system() == 'Windows' and not IS_32BITS:
```
shssf: On Linux 64 bit, what type will be used for `A`?

kozlov-alexey: The expectation is to have default np.int for A which is np.int64 on both Linux and MacOS (both 64-bit OS). 
For Windows 32-bit we can still use old test_impl, as it's not affected by the issue, and we'll have both A's dtype and Numba's types.intp int32. And only for Windows 64-bit we will have these 'tweaked' tests, which just fail otherwise.

shssf: I have an idea.
Copy this original test (and others rlevent to this issue) to https://github.com/IntelPython/hpat/blob/master/hpat/tests/test_hpat_jit.py module to collect such issues.
Modify this test as you proposed before (with explicit typing).
Name of copied test must be the same as this one with `_issue` suffix to find original test later. Or add comment in the test body where it was copied from.

shssf:
Please squash commits in this PR

PR #185, opened 2019-10-01
Moved tests on unique method to test_series

akharche:
Changed incorrect assertion
Created test on unique that shows problems in algorithm implementation


shssf:
depends on #186

PR #186, opened 2019-10-01
Method Series.unique() in new style

akharche:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,44 @@ def hpat_pandas_series_le_impl(self, other):
...
+        return np.unique(self._data)
+
+    def hpat_pandas_series_unique_impl(self):
+        unique_values = []
+        for value in self._data:
+            if value not in unique_values:
+                unique_values.append(value)
+        return unique_values
+
+    return hpat_pandas_series_unique_impl_sorted if sorted else hpat_pandas_series_unique_impl
```
shssf: I think you could control parameter `soret` inside `_impl` function by `return_index` parameter to `np`
```
>>> import numpy as np
>>> a = [4,2,1,3,1,2,3,4]
>>> np.unique(a)
array([1, 2, 3, 4])
>>> indexes = np.unique(a, return_index=True)[1]
>>> [a[index] for index in sorted(indexes)]
[4, 2, 1, 3]
```
https://stackoverflow.com/questions/12926898/numpy-unique-without-sort

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,44 @@ def hpat_pandas_series_le_impl(self, other):
...
+         returns :obj:`numpy.array` ndarray
+    """
+
+    _func_name = 'Method unique().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    def hpat_pandas_series_unique_impl_sorted(self):
```
shssf: Please, do not forget to use correct parameters for `_impl` function. Example https://github.com/numba/numba/blob/master/numba/unicode.py#L741
Also, check for parameter is boolean

hpat/datatypes/hpat_pandas_series_functions.py
akharche: I'm afraid Numba doesn't support return_index parameter inside unique() method:
`TypeError: np_unique() got an unexpected keyword argument 'return_index'`

hpat/tests/test_series.py
```diff
@@ -1693,6 +1693,14 @@ def test_impl():
         hpat_func = hpat.jit(test_impl)
         np.testing.assert_array_equal(hpat_func(), test_impl())
 
+    def test_series_unique(self):
```
densmirn: It makes sense to add addition tests, e.g. to check raising exceptions when passed arguments of incorrect type.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,44 @@ def hpat_pandas_series_le_impl(self, other):
         return hpat_pandas_series_le_impl
 
     raise TypingError('{} The object must be a pandas.series and argument must be a number. Given: {} and other: {}'.format(_func_name, self, other))
+
+
+@overload_method(SeriesType, 'unique')
+def hpat_pandas_series_unique(self, sorted=False):
```
shssf: @akharche We had many internal discussions about `sorted` parameter. Finally, my opinion was rejected. Please ignore this parameter, return values as it was before, not in the way described in Pandas documentation.
**Important**: Please add `Note` to main documentation (not developer only. before line 1139) about this changed behavior. Some thin like `Note: Return values order sorted by values`

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1307,3 +1309,49 @@ def hpat_pandas_series_abs_impl(self):
...
+
+    _func_name = 'Method unique().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if self.dtype is types.unicode_type:
+        def hpat_pandas_series_unique_str_impl(self):
+            unique_strings = init_set_string()
```
shssf: What type returned by `init_set_string()`?
Is it ok to call `.add` on this type?

PR #188, opened 2019-10-02
Series.abs() in new style

akharche:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,35 @@ def hpat_pandas_series_le_impl(self, other):
...
+
+    Returns
+    -------
+    :obj:`pandas.Series`
+         returns :obj:`pandas.Series` containing the absolute value of elements
+    """
+
+    _func_name = 'Method abs().'
+
+    if not isinstance(self, SeriesType):
```
shssf: I think we need additional check like this. As I see this method can work with numeric data only but Series can contain, for example, strings (not at this time but in general).
In pandas.abs() documentation we have `This function only applies to elements that are all numeric.`. I think we need to raise exception if underling data is not numeric.

hpat/hiframes/pd_series_ext.py
```diff
@@ -598,14 +598,14 @@ def _resolve_combine_func(self, ary, args, kws):
...
 
-    @bound_function("series.abs")
-    def resolve_abs(self, ary, args, kws):
-        # call np.abs(A) to get return type
-        arr_typ = series_to_array_type(ary)
-        all_args = tuple([arr_typ] + list(args))
-        ret_typ = self.context.resolve_function_type(np.abs, all_args, kws).return_type
-        ret_typ = if_arr_to_series_type(ret_typ)
-        return signature(ret_typ, *args)
+    # @bound_function("series.abs")
```
shssf: This is good. This is typing stage. What happen if comment out definition of the function here https://github.com/IntelPython/hpat/blob/ea6b1a22ba3a04d4495b6a69bb762c09dc41be67/hpat/hiframes/series_kernels.py#L516

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -33,6 +33,7 @@
 
 import operator
 import pandas
+import numpy as np
```
shssf: Could you please do not rename numpy?

hpat/datatypes/hpat_pandas_series_functions.py
Rubtsowa: @shssf In 1135 line "def hpat_pandas_series_append(self):", but it must be   "def hpat_pandas_series_abs(self):".

PR #189, opened 2019-10-02
Series.max() in new style

akharche:
Due to overloaded method 'max' in numba it calls array.max for SeriesType instead of hpat_pandas_series_max. Firstly, need fix in importing.


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,55 @@ def hpat_pandas_series_le_impl(self, other):
...
+    if not (isinstance(axis, types.Omitted) or axis == 0) \
+            or not (isinstance(level, types.Omitted) or level is None) \
+            or not (isinstance(numeric_only, types.Omitted) or numeric_only is None):
+        raise TypingError(
+            '{} Unsupported parameters. Given axis: {}, level: {}, numeric_only: {}'.format(_func_name, axis, level,
+                                                                                            numeric_only))
+
+    def hpat_pandas_series_max_impl(self, axis=0, skipna=True, level=None, numeric_only=None):
+        if skipna:
+            return np.nanmax(self._data)
```
shssf: Do we expect Numpy data type returned? I suspect we need Series data type constructed here

hpat/hiframes/pd_series_ext.py
```diff
@@ -700,13 +700,13 @@ def resolve_idxmax(self, ary, args, kws):
...
         return signature(types.intp, *args)
 
-    @bound_function("series.max")
-    def resolve_max(self, ary, args, kws):
-        assert not kws
-        dtype = ary.dtype
-        dtype = (pandas_timestamp_type
-                 if isinstance(dtype, types.NPDatetime) else dtype)
-        return signature(dtype, *args)
+    # @bound_function("series.max")
```
shssf: This is good. This is typing stage. What happen if comment out definition of the function here https://github.com/IntelPython/hpat/blob/ea6b1a22ba3a04d4495b6a69bb762c09dc41be67/hpat/hiframes/series_kernels.py#L499

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1128,3 +1129,55 @@ def hpat_pandas_series_le_impl(self, other):
         return hpat_pandas_series_le_impl
 
     raise TypingError('{} The object must be a pandas.series and argument must be a number. Given: {} and other: {}'.format(_func_name, self, other))
+
+
+@overload_method(SeriesType, 'max')
+def hpat_pandas_series_max(self, axis=0, skipna=True, level=None, numeric_only=None):
```
shssf: Parameters are diffrent https://pandas.pydata.org/pandas-docs/version/0.25/reference/api/pandas.Series.max.html?highlight=max#pandas.Series.max
Please fix the parameters

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -33,6 +33,7 @@
 
 import operator
 import pandas
+import numpy as np
```
shssf: Could you please do not rename numpy?

hpat/datatypes/hpat_pandas_series_functions.py
akharche: We return scalars because Series can be returned if level specified. But we don't have multi indexing now and parameter 'level' is not supported.

hpat/datatypes/hpat_pandas_series_functions.py
shssf: parameter `skipna=True` is still diffrent

hpat/datatypes/hpat_pandas_series_functions.py
fschlimb: why not? This is how numpy is use is everywhere else.

hpat/datatypes/hpat_pandas_series_functions.py
shssf: Sorry, looked into pandas sources and found
```
def max(self, axis=None, skipna=True, *args, **kwargs):
```
it looks like they have documentations issue

hpat/datatypes/hpat_pandas_series_functions.py
shssf: @fschlimb to keep clear in the code what exactly used.

fschlimb: I understand and I also generally prefer fully qualified names. In this particular case we should stay with common practice (e.g. import numpy as np and import pandas ad pd)

PR #200, opened 2019-10-07
Implement isna

1e-to:


hpat/tests/test_series.py
```diff
@@ -1265,6 +1265,14 @@ def test_impl(S, values):
         values = {'b', 'c', 'e'}
         pd.testing.assert_series_equal(hpat_func(S, values), test_impl(S, values))
 
+    def test_series_isna(self):
```
akharche: We already have such test in test_series: test_series_isna1. Please, use it or specify name of these tests like: test_series_isnan_int and test_series_isnan_float

hpat/tests/test_series.py
```diff
@@ -1265,6 +1265,14 @@ def test_impl(S, values):
...
+        def test_impl(S):
+            return S.isna()
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series([0, np.NaN, -3])
```
akharche: Let's follow common definition: np.nan

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -356,6 +357,44 @@ def hpat_pandas_series_append_impl(self, to_append):
...
+         returns :obj:`pandas.Series` object
+    """
+
+    _func_name = 'Method isna().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, (types.Integer, types.Float)):
```
akharche: I think, we can add implementation for Strings too

akharche:
Please, define commit message more precisely. Now it is not obvious, which type this method is implemented for 
Ex: Implement series.isna

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -356,6 +357,65 @@ def hpat_pandas_series_append_impl(self, to_append):
     return hpat_pandas_series_append_impl
 
 
+@overload_method(SeriesType, 'isna')
+def hpat_pandas_series_isna(self):
+    """
+    Pandas Series method :meth:`pandas.Series.append` implementation.
```
kozlov-alexey: Wrong method name

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -356,6 +357,65 @@ def hpat_pandas_series_append_impl(self, to_append):
...
+
+    Parameters
+    -----------
+    self : :obj:`pandas.Series` object
+               input argument
+
+    Returns
+    -------
+    :obj:`pandas.Series`
+         returns :obj:`pandas.Series` object
```
kozlov-alexey: Can you please describe what this output series contains, e.g. 'that contains bool values indicating if corresponding element of self is NaN"

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -356,6 +357,65 @@ def hpat_pandas_series_append_impl(self, to_append):
...
+    """
+
+    _func_name = 'Method isna().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if isinstance(self.dtype, (types.Integer, types.Float)):
```
kozlov-alexey: types.Integer - do we really handle integer series? I thought there's no such thing as NaN in integer world.
Can you please check that numpy array does have int dtype?

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: I close this PR, look new one

hpat/tests/test_series.py
kozlov-alexey: Let's add np.inf as well and see if it's handled correctly.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -356,6 +357,65 @@ def hpat_pandas_series_append_impl(self, to_append):
...
 
 
+@overload_method(SeriesType, 'isna')
+def hpat_pandas_series_isna(self):
+    """
+    Pandas Series method :meth:`pandas.Series.append` implementation.
+
+    .. only:: developer
+
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_isna
```
kozlov-alexey: Test for string type -  test_series_str_isna1- should be mentioned here too

PR #202, opened 2019-10-08
Add support for DataFrame.astype method

kozlov-alexey:


hpat/hiframes/dataframe_pass.py
```diff
@@ -1218,11 +1233,11 @@ def _run_call_col_reduce(self, assign, lhs, rhs, func_name):
         for d in data_args:
             func_text += "  {} = hpat.hiframes.api.init_series({})\n".format(d + '_S', d)
             func_text += "  {} = {}.{}()\n".format(d + '_O', d + '_S', func_name)
-        func_text += "  data = np.array(({},))\n".format(
+            func_text += "  data = np.array(({},))\n".format(
             ", ".join(d + '_O' for d in data_args))
-        func_text += "  index = hpat.str_arr_ext.StringArray(({},))\n".format(
+            func_text += "  index = hpat.str_arr_ext.StringArray(({},))\n".format(
```
shssf: Why you moved these lines under `for`? Are you sure that previous algorithm was wrong?

shssf:
I see implementation of the method is relevant to HPAT compiler pipeline availability only.
Is it an initial idea to implement this in this way?

hpat/tests/test_series.py
```diff
@@ -256,22 +257,123 @@ def test_impl(A):
...
+        pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
+
+    def test_series_astype_str_to_str1(self):
+        def test_impl(S):
+            return S.astype(str)
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series(['aa', 'bb', 'cc'])
+        pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
+
```
fschlimb: Pls add comments/docstrings indicating differences between *_to_str? variants.

fschlimb:
I didn't look into the details but I am missing comments. Except TODOs there is only one comment describing the code. Every feature should have at least a high-level description of how it works.

hpat/hiframes/dataframe_pass.py
kozlov-alexey: No, I overlooked it in the difference before push. Reverted this change.

hpat/tests/test_series.py
kozlov-alexey: Added.

kozlov-alexey:
@shssf, @fschlimb 
> I see implementation of the method is relevant to HPAT compiler pipeline availability only.
> Is it an initial idea to implement this in this way?
Yes, we need support for DF.astype method and implementing it via Series.astype (currently heavily relying on HPAT legacy types and passes) seemed as the easiest way. Later when Series.astype is refactored we could re-design this too. But now it will probably be blocked by inability to iterate over DF columns in the overload impl anyway. 

> I didn't look into the details but I am missing comments. Except TODOs there is only one comment describing the code. Every feature should have at least a high-level description of how it works.
I added comments in the function generating impl and tests. Hope that's sufficient.

shssf:
please resolve conflicts

PR #205, opened 2019-10-09
Fix typo in abs method name

akharche:


akharche:
@Rubtsowa fix concerning your comment in closed PR

PR #206, opened 2019-10-09
[WIP]adding few examples to getting started section

rdesai16:
Added two basic examples with series and dataframe operations


shssf:
Is it still "work in progress" stage?

PR #207, opened 2019-10-10
Impl isna

1e-to:
In the implementation for strings, the creation of an array occurs through a workaround, how can I create an array of bools in a jit region?


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+
+        def hpat_pandas_series_isna_impl(self):
+
+            return pandas.Series(numpy.isnan(self._data))
+
+        return hpat_pandas_series_isna_impl
+
+    if isinstance(self.dtype, types.UnicodeType):
+
+        def hpat_pandas_series_isna_impl(self):
```
shssf: What is the algo here? Is it copied from somewhere? Please put a comment about it

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
     return hpat_pandas_series_groupby_impl
 
 
+@overload_method(SeriesType, 'isna')
+def hpat_pandas_series_isna(self):
+    """
+    Pandas Series method :meth:`pandas.Series.append` implementation.
+
```
kozlov-alexey: :meth:`pandas.Series.append`

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+    Parameters
+    -----------
+    self : :obj:`pandas.Series` object
+               input argument
+
+    Returns
+    -------
+    :obj:`pandas.Series`
+         returns :obj:`pandas.Series` object
```
kozlov-alexey: Can you please describe what this output series contains, e.g. 'that contains bool values indicating if corresponding element of self is NaN"

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+            byte_size = 8
+            for i in range(len(self._data)):
+                bmap_idx = i // byte_size
+                bit_idx = i % byte_size
+                bmap = self._data.null_bitmap[bmap_idx]
+                bit_value = (bmap >> bit_idx) & 1
+                result[i] = bit_value == 0
+            return pandas.Series(result)
+
+        return hpat_pandas_series_isna_impl
```
kozlov-alexey: We have hpat.str_arr_ext.str_arr_is_na implementation we can simply call here on self._data.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+            #Need to find way to create arr[bool] without workaround
+            result = arr == 4
+            byte_size = 8
+            for i in range(len(self._data)):
+                bmap_idx = i // byte_size
+                bit_idx = i % byte_size
+                bmap = self._data.null_bitmap[bmap_idx]
+                bit_value = (bmap >> bit_idx) & 1
+                result[i] = bit_value == 0
+            return pandas.Series(result)
```
kozlov-alexey: 'return pandas.Series(result)'
This will produce series with different index than self.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+
+        return hpat_pandas_series_isna_impl
+
+    if isinstance(self.dtype, types.UnicodeType):
+
+        def hpat_pandas_series_isna_impl(self):
+            arr = numpy.arange(len(self._data))
+            #Need to find way to create arr[bool] without workaround
+            result = arr == 4
+            byte_size = 8
```
kozlov-alexey: The sizes of types in bits highly depend on platform and compiler and it's no good hard-coding them.
You need to change that to the size of null_bitmap element in bits.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+            return pandas.Series(numpy.isnan(self._data))
+
+        return hpat_pandas_series_isna_impl
+
+    if isinstance(self.dtype, types.UnicodeType):
+
+        def hpat_pandas_series_isna_impl(self):
+            arr = numpy.arange(len(self._data))
+            #Need to find way to create arr[bool] without workaround
+            result = arr == 4
```
kozlov-alexey: Why don't you use something like
np.empty(len(self._data), dtype=np.bool_)

In any case - if there's any issue with NumPy functions support in Numba, they need to be reported.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
+            byte_size = 8
+            for i in range(len(self._data)):
+                bmap_idx = i // byte_size
+                bit_idx = i % byte_size
```
kozlov-alexey: Please add a comment here, at the top of the loop describing what it does - at the moment it's not clear what's stored in  null_bitmap from this code and why we are iterating over it (i.e. you have to know the type of self._data to understand that).

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Do you think this is necessary? This does not occur with the uniform style of other functions

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -409,6 +409,60 @@ def hpat_pandas_series_groupby_impl(
...
 
+@overload_method(SeriesType, 'isna')
+def hpat_pandas_series_isna(self):
+    """
+    Pandas Series method :meth:`pandas.Series.append` implementation.
+
+    .. only:: developer
+
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_isna
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_isna1
```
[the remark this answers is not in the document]
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: We are returning a new series

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: This implementation is better because it is clear and you do not need to turn to intrinsic

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: it doesn't matter here

hpat/str_arr_ext.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, dmm, fe_type):
     ('meminfo', types.MemInfoPointer(str_arr_payload_type)),
 ]
 
-
+make_attribute_wrapper(StringArrayType, 'null_bitmap', 'null_bitmap')
```
shssf: is any way to avoid using this?

1e-to: No, we dont have another way.
Also there is an idea in StringArrayType to create an overload method that would implement isna and which we could pull from the overload for the series.

PR #213, opened 2019-10-11
Unboxing for pandas Int64Index and RangeIndex

akharche:
Simple solution for unboxing Int64Index and RangeIndex


hpat/hiframes/boxing.py
```diff
@@ -155,10 +155,18 @@ def _infer_series_list_dtype(S):
...
         first_val = index[0]
         if isinstance(first_val, str):
             return string_array_type
+
+    elif index.dtype == np.int64:
+        # RangeIndex or Int64Index
+        dtype = numba.numpy_support.from_dtype(index.dtype)
+        return types.Array(dtype, 1, 'C')
+
     return types.none
```
shssf:
```suggestion
    return index.dtype
```
and no need code (elif) above?

hpat/hiframes/boxing.py
```diff
@@ -154,12 +154,21 @@ def _infer_series_list_dtype(S):
 
 
 def _infer_index_type(index):
-    # TODO: support proper inference
+    '''
+    Convertion input index type into Numba known type
+    need to return instance of the type class
+    '''
+
+    if isinstance(index, (types.NoneType, pd.RangeIndex)) or index is None:
```
kozlov-alexey: But the following (existing) tests will still fail if we don't create index from pd.RangeIndex during unboxing:

```
    def test_series_dropna_float1(self):
        def test_impl(A):
            return A.dropna()

        S1 = pd.Series([1.0, 2.0, np.nan, 1.0, np.inf])
        S2 = S1.copy()
        pd.testing.assert_series_equal(hpat_func(S1), test_impl(S2))
```
The error will be:
Series.index values are different (50.0 %)
[left]:  RangeIndex(start=0, stop=4, step=1)
[right]: Int64Index([0, 1, 3, 4], dtype='int64')


```
    def test_series_default_index(self):
        def test_impl():
            A = pd.Series([3, 2, 1, 5, 4])
            return A.index

```
The error will be:
AssertionError:
Arrays are not equal

Mismatch: 500%
 x: array(None, dtype=object)
 y: array([0, 1, 2, 3, 4], dtype=int64)

hpat/tests/test_series.py
```diff
@@ -303,6 +303,32 @@ def test_impl(S):
         S = pd.Series(['aa', 'bb', 'cc'])
         pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
 
+    def test_series_astype_str_to_str_index_str(self):
```
kozlov-alexey: This is not boxing/unboxing issue, so better to revert it and do it in a separate PR.

hpat/hiframes/pd_series_ext.py
```diff
@@ -448,7 +448,7 @@ def resolve_astype(self, ary, args, kws):
         dtype, = args
         if ((isinstance(dtype, types.Function) and dtype.typing_key == str)
                 or (isinstance(dtype, types.StringLiteral) and dtype.literal_value == 'str')):
-            ret_type = SeriesType(string_type)
+            ret_type = SeriesType(string_type, index=ary.index)
```
kozlov-alexey: Same as below

hpat/hiframes/boxing.py
```diff
@@ -154,12 +154,21 @@ def _infer_series_list_dtype(S):
...
+        return types.none
+
     if index.dtype == np.dtype('O') and len(index) > 0:
         first_val = index[0]
         if isinstance(first_val, str):
             return string_array_type
-    return types.none
+
+    numba_index_type = numpy_support.from_dtype(index.dtype)
+    return types.Array(numba_index_type, 1, 'C')
```
kozlov-alexey: But what if we deduce Array type here, but provided index is not something we can unbox to Array later?
Are there tests for TimedeltaIndex or some other non Numeric index (see https://pandas.pydata.org/pandas-docs/stable/reference/indexing.html#numeric-index)?

shssf: ` TimedeltaIndex` is disabled. All tests are passed. Please provide a test case.

hpat/hiframes/boxing.py
kozlov-alexey: @shssf Yes, because existing version of test_series_dropna_float1 omits checking Series indexes at all.
Please see that the assert in the code I provided is different:
`pd.testing.assert_series_equal(hpat_func(S1), test_impl(S2))`

kozlov-alexey: Also one more test that don't seem to work with current design:

```
    def test_pass_series_int_index(self):
        '''Verifies boxing and unboxing of a Series with integer index'''
        def test_impl(A):
            return A

        S = pd.Series([3, 5, 6], [1, 2, 3], name='A')
```
Also please use the following git patch to add more tests (including the one above), remove duplicate test and add docstrings: [patch]( https://github.com/kozlov-alexey/hpat/blob/e05ed780cf32311f27e506004fac8762654eb80e/index_tests_patch.txt)

PR #215, opened 2019-10-11
Handling None Index in Series init

akharche:


hpat/hiframes/pd_series_ext.py
```diff
@@ -1276,9 +1276,23 @@ def hpat_pandas_series_index_ctor_impl(
...
-        return hpat.hiframes.api.init_series(hpat.hiframes.api.fix_df_array(data), index, name)
+    def hpat_pandas_series_index_ctor_impl(
+            data=None,
+            index=None,
+            dtype=None,
+            name=None,
+            copy=False,
+            fastpath=False):
+
+        # TODO: Do we need keep the whole Array?
```
shssf: No. I would say more - we do not need to keep index if it is not set explicitly.

hpat/hiframes/pd_series_ext.py
```diff
@@ -1276,9 +1276,23 @@ def hpat_pandas_series_index_ctor_impl(
...
+        index = np.arange(len(data))
+
+        return hpat.hiframes.api.init_series(
```
shssf: It looks like the code can be compacted by calling `hpat.hiframes.api.init_series` only once
also, is it really required code style change? I thought it has less than 120 symbols in a row.

akharche:
Not optimal solution

PR #216, opened 2019-10-11
Implement idxmin

1e-to:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,57 @@ def hpat_pandas_series_ge_impl(self, other):
     raise TypingError('{} The object must be a pandas.series and argument must be a number. Given: {} and other: {}'.format(_func_name, self, other))
 
 
+@overload_method(SeriesType, 'idxmin')
+def hpat_pandas_series_idxmin(self, axis=None, skipna=True):
```
shssf: @1e-to I see in documentation the interface is a bit diffrent. It looks like `self, axis=0, skipna=True, *args` The last argument (`*args`)is not used but it might be good to put it in the function parameters

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,57 @@ def hpat_pandas_series_ge_impl(self, other):
...
+    Parameters
+    -----------
+    axis :  :obj:`int`, default: None
+                *unsupported*
+    skipna:  :obj:`bool`, default: True
+                Exclude nan values when computing the result
+
+    Returns
+    -------
+    :obj:`pandas.Series` or :obj:`int` or :obj:`float`
```
shssf: It looks like description is wrong. The method returns label of selected index. I'm not sure but it looks like it returns a value (corresponded type) of the index.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,57 @@ def hpat_pandas_series_ge_impl(self, other):
...
+         returns :obj:`pandas.Series` object or :obj:`int` or :obj:`float`
+    """
+
+    _func_name = 'Method idxmin().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
```
shssf: `Given self` -> `Given`

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,57 @@ def hpat_pandas_series_ge_impl(self, other):
...
+
+    if not isinstance(self.dtype, types.Number):
```
shssf: Do we really focused on numeric data in series? No strings array supported?
Also, I would vote to avoid `self.dtype` using because I see it ambiguous. I think it is better to use `self._data.dtype` instead.

hpat/hiframes/pd_series_ext.py
```diff
@@ -241,6 +241,7 @@ def __init__(self, dmm, fe_type):
 make_attribute_wrapper(SeriesType, 'name', '_name')
 
 
+from hpat.datatypes.hpat_pandas_series_functions import *
```
shssf: this line should not be moved here

hpat/hiframes/hiframes_typed.py
```diff
@@ -866,7 +866,7 @@ def _run_call_series(self, assign, lhs, rhs, series_var, func_name):
             return self._replace_func(func, [data], pre_nodes=nodes)
 
         if func_name in ('std', 'nunique', 'describe', 'isna',
-                         'isnull', 'median', 'idxmin', 'idxmax', 'unique'):
+                         'isnull', 'median', 'idxmax', 'unique'):
```
shssf: I think it will not work (didn't check). if parallel tests are not working, please leave this line as in original, and add `resolve_idxmin` value to the array as here https://github.com/IntelPython/hpat/pull/191/files#diff-f13c21aac60fb8f7ae3c9527239d1e17R990

hpat/hiframes/series_kernels.py
```diff
@@ -530,6 +530,6 @@ def gt_f(a, b):
     'head_index': lambda A, I, k, name: hpat.hiframes.api.init_series(A[:k], I[:k], name),
     'median': lambda A: hpat.hiframes.api.median(A),
     # TODO: handle NAs in argmin/argmax
-    'idxmin': lambda A: A.argmin(),
+    # 'idxmin': lambda A: A.argmin(),
```
shssf: I think you need to leave it as in original

hpat/tests/test_series.py
```diff
@@ -1556,6 +1557,32 @@ def test_impl(A):
...
+        S = pd.Series([8, 6, 34, np.nan])
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+    @unittest.skip("Enable after fixing index")
+    def test_series_idxmin_int(self):
+        def test_impl(S):
+            return S.idxmin()
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series([1, 2, 3], [4, 45, 14])
```
shssf: please see this comment https://github.com/IntelPython/hpat/pull/217#discussion_r335113325

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: I would like to correct a little bit: `self._data.dtype` -> `self.data.dtype`. `self` has not attribute `_data` on typing level.

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

hpat/hiframes/pd_series_ext.py
1e-to: Fixed

hpat/hiframes/hiframes_typed.py
1e-to: Initially, there were no parallel tests for this function, should I write them?
resolve_idxmin need to dont overlap arrays methods, but idxmin for array does not exist

hpat/tests/test_series.py
1e-to: Add 1 test and fix another

hpat/hiframes/hiframes_typed.py
shssf: No, let's wait test system results

hpat/tests/test_series.py
```diff
@@ -1556,6 +1556,96 @@ def test_impl(A):
...
+        print(hpat_func(S))
```
kozlov-alexey: Do we need this in test?

1e-to: Fixed

hpat/tests/test_series.py
```diff
@@ -1556,6 +1556,96 @@ def test_impl(A):
...
+        print(test_impl(S))
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+    @unittest.skip("Need index fix")
+    def test_series_idxmin(self):
+        def test_series_idxmin_impl(S):
```
kozlov-alexey: I think it's better to stick to the common name for the jitted func - test_impl.

hpat/tests/test_series.py
```diff
@@ -1556,6 +1556,96 @@ def test_impl(A):
...
+                       [6, 6.1, 2.2, 1, 3, 3, 2.2, 1, 2],
+                       ]
+
+        data_extra = [[np.nan, np.nan, np.nan, np.nan],
+                      [np.nan, np.nan, np.inf, np.inf],
+                      ]
+
+        test_input_data = data_simple + data_extra
+
+        for input_data in data_simple:
```
kozlov-alexey: Are the first two loops for testing default index in Series? Then they should make one separate test. You will have another compilation of test_func on the last two loops anyway, so there'll be no benefit to put this all in one test.

hpat/tests/test_series.py
```diff
@@ -1556,6 +1556,96 @@ def test_impl(A):
...
+
+        hpat_func = hpat.jit(test_series_idxmin_impl)
+
+        test_input_data = []
+        data_simple = [[6, 6, 2, 1, 3, 3, 2, 1, 2],
+                       [1.1, 0.3, 2.1, 1, 3, 0.3, 2.1, 1.1, 2.2],
+                       [6, 6.1, 2.2, 1, 3, 3, 2.2, 1, 2],
+                       ]
+
+        data_extra = [[np.nan, np.nan, np.nan, np.nan],
```
kozlov-alexey: I think it's better to have one 'series_data' list (instead of data_simple + data_extra) with it's elements covering all possible situations (i.e. containing numbers, NaNs or both, in different combinations). There's no obvious benefit of using loops in unittests - better identify and cover different cases yourself.

hpat/tests/test_series.py
```diff
@@ -1556,6 +1556,96 @@ def test_impl(A):
...
+        for input_data in test_input_data:
+            S = pd.Series(input_data)
+
+            result_ref = test_series_idxmin_impl(S)
+            result = hpat_func(S)
+            self.assertEqual(result, result_ref)
+
+        for input_data in data_simple:
+            for index_data in data_simple:
+                S = pd.Series(input_data, index_data)
```
kozlov-alexey: Why do we need to test index_data at all? Does our implementation depend on the indexes?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,66 @@ def hpat_pandas_series_ge_impl(self, other):
...
+def hpat_pandas_series_idxmin(self, axis=None, skipna=True, *args):
+    """
+    Pandas Series method :meth:`pandas.Series.idxmin` implementation.
+
+    .. only:: developer
+
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_idxmin1
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_idxmin_str
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_idxmin_int
+        Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_idxmin_no
```
densmirn: Please update the list of tests in docstring according to the real list.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1096,6 +1096,66 @@ def hpat_pandas_series_ge_impl(self, other):
...
+    """
+
+    _func_name = 'Method idxmin().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(self.data.dtype, types.Number):
+        raise TypingError(
+            '{} Currently function supports only numeric values. Given data type: {}'.format(_func_name, self.dtype))
```
densmirn: `self.dtype` -> `self.data.dtype`

PR #217, opened 2019-10-14
Implement Series.size in new style

PokhodenkoSA:


hpat/tests/test_series.py
```diff
@@ -186,7 +186,7 @@ def test_impl(A):
         S = pd.Series([3, 5, 6], ['a', 'b', 'c'], name='A')
         pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
 
-    def test_series_attr1(self):
+    def test_series_size(self):
```
shssf: We don't need to create DataFrame for this.
Please test at least two dataset: numeric (inlucde NaN) and strings

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -230,6 +230,37 @@ def hpat_pandas_series_index_impl(self):
     return hpat_pandas_series_index_impl
 
 
+@overload_attribute(SeriesType, 'size')
+def hpat_pandas_series_size(self):
+    """
+    Pandas Series attribute :attr:`pandas.Series.size` implementation
+
+    **Algorithm**: result = series.size
```
shssf: No need this section. Please refer to other methods

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -230,6 +230,37 @@ def hpat_pandas_series_index_impl(self):
...
+        Return the number of elements in the underlying data.
+    """
+
+    _func_name = 'Attribute size.'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    def hpat_pandas_series_size_impl(self):
+        return self._data.size
```
shssf: I would propose to use `len(self._data)` instead.

hpat/tests/test_series.py
PokhodenkoSA: Done

hpat/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: As I can see playing with Pandas in console: pd.Series() does not store _data = None in internal representation. So _data is always something even if Series is empty.
Also this could be replaced with `len(self._data)`.

hpat/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: Sorry, I do not understand. I see lines like this in many other methods.
I will remove `Algorithm` and add `.. only:: developer`. Is it ok?

hpat/tests/test_series.py
```diff
@@ -186,14 +186,23 @@ def test_impl(A):
...
+        S = pd.Series(np.arange(n))
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+        S = pd.Series()
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+        S = pd.Series([np.nan, 1, 2])
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+        S = pd.Series(['1', '2', '3'])
```
shssf: @PokhodenkoSA It might be better to make these set of tests as a loop.
I mean:
```
for input_data in test_input_data:
    S = pd.Series(input_data)
```
where data defined as https://github.com/IntelPython/hpat/blob/e45cc922afb855f0e5d87e30acc01ef6ef2e4002/hpat/tests/test_series.py#L2057

also, it might be better to use `pd.testing.assert_series_equal` instead `self.assertEqual` because, as I understood, the first one compare full `Series` datatype but second compares part of it. Yes, we use diffrent types of assertion in tests but it will be a one of the task for redeveloping.

hpat/datatypes/hpat_pandas_series_functions.py
shssf: ok, never mind, leave it as is.

hpat/tests/test_series.py
PokhodenkoSA: I think `pd.testing.assert_series_equal` is not applicable here because it compares who Series but Series.size returns scalar.

PR #218, opened 2019-10-14
change nunique

Rubtsowa:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1517,18 +1517,18 @@ def hpat_pandas_series_nunique(self, dropna=True):
...
-                [3, np.nan, 'b', 'b', 5.1, np.nan, 'ddd', 'dd']
             """
 
             str_set = set(self._data)
-            return len(str_set)
- 
+            if dropna == False:
+                return len(str_set)
+            else:
+                return len(str_set) + 1
```
shssf: This should not work with Numba pipeline (in HPAT these tests are disabled).
Input data is String, non None exists and parameter dropna=True
To do such thing I did `nan_exists` variable below

Rubtsowa: @shssf Shouldn't it work at all or is it implemented incorrectly?

shssf: @Rubtsowa I would expect failed tests

Rubtsowa:
@shssf I made changes. Now the error, it seems, does not occur. Check, please.

PR #220, opened 2019-10-14
Implement series.var() in new style

densmirn:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -198,6 +198,79 @@ def hpat_pandas_series_values_impl(self):
...
+    numeric_only: :obj:`bool`
+        Include only float, int, boolean columns.
+        If None, will attempt to use everything, then use only numeric data.
+        Not implemented for Series.
+        *unsupported*
+
+    Returns
+    -------
+    :obj:`scalar`
+         returns :obj:`scalar`
```
shssf: It can return scalar or Series

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -198,6 +198,79 @@ def hpat_pandas_series_values_impl(self):
...
+    """
+
+    _func_name = 'Method var().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, types.Number):
```
shssf: https://github.com/IntelPython/hpat/pull/216#discussion_r335117490

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -198,6 +198,79 @@ def hpat_pandas_series_values_impl(self):
...
+
+    def hpat_pandas_series_var_impl(self, axis=None, skipna=None, level=None, ddof=1, numeric_only=None):
+        if skipna is None:
+            skipna = True
+
+        if skipna:
+            valuable_length = len(self._data) - numpy.sum(numpy.isnan(self._data))
+            return numpy.nanvar(self._data) * valuable_length / (valuable_length - ddof)
+
+        return self._data.var() * len(self._data) / (len(self._data) - ddof)
```
shssf: I don't believe numpy has no such functionality

hpat/hiframes/hiframes_typed.py
```diff
@@ -851,7 +851,7 @@ def parse_impl(data):
 
     def _run_call_series(self, assign, lhs, rhs, series_var, func_name):
         # single arg functions
-        if func_name in ('sum', 'count', 'mean', 'var', 'min', 'max', 'prod'):
+        if func_name in ('sum', 'count', 'mean', 'min', 'max', 'prod'):
```
shssf: I think this should brake parallelism. not sure.

hpat/tests/test_series.py
```diff
@@ -2099,6 +2101,58 @@ def test_series_nunique_param1_impl(S, dropna):
...
+        ref_result = pyfunc()
+        result = cfunc()
+        np.testing.assert_equal(ref_result, result)
+
+    def test_series_var_unboxing(self):
+        def pyfunc(series, skipna, ddof):
+            return series.var(skipna=skipna, ddof=ddof)
+
+        cfunc = hpat.jit(pyfunc)
+        series = pd.Series([1.3, -2.7, np.nan, 0.1, 10.9])
```
shssf: https://github.com/IntelPython/hpat/pull/217#discussion_r335113325

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: I'm not sure we have to add `series` here, because `series` is returned in case of `level` specified, but currently we don't support parameter `level`. So `scalar` is always returned in our case. Do you still think I need to add `series` here?

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Numpy implementation has `ddof=0` by default, but Pandas has `ddof=1` by default. So I has to make this extra change to align the parameter between these implementations.

hpat/hiframes/hiframes_typed.py
densmirn: You are right, It breaks multiprocessing parallelism, but I still don't know how to save both approaches ("old" and "new") simultaneously. Please let me know if you have thoughts about it.

shssf: where is no single and simple recipe

densmirn: As I can see there is no multiprocessing parallelism for the algorithm in HPAT, only multi-threading based on `numba.parfor`. Numba implemented the method for numpy also using `numba.parfor` and we invoke the method in our implementation. So we are able to drop "old style" in this case.

densmirn: May previous comment was wrong.

densmirn:
@shssf, @kozlov-alexey could you take the next round to review the PR?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -256,6 +256,86 @@ def hpat_pandas_series_values_impl(self):
...
+    Parameters
+    ----------
+    self: :obj:`pandas.Series`
+        input series
+    axis: :obj:`int`, :obj:`str`
+        Axis along which the operation acts
+        0/None/'index' - row-wise operation
+        1/'columns'    - column-wise operation
+        *unsupported*
+    skipna: :obj:`bool`
```
shssf: I suspect skipna=True

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -256,6 +256,86 @@ def hpat_pandas_series_values_impl(self):
     return hpat_pandas_series_values_impl
 
 
+@overload_method(SeriesType, 'var')
+def hpat_pandas_series_var(self, axis=None, skipna=None, level=None, ddof=1, numeric_only=None):
```
shssf: skipna=True?

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: By default `skipna=None` in Pandas interface. So we did the same.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: By default `skipna=None` in Pandas interface.

PR #224, opened 2019-10-14
Rebase on numba 0.46

AlexanderKalistratov:


hpat/decorators.py
```diff
@@ -47,6 +47,8 @@ def jit(signature_or_function=None, **options):
         return numba.jit(signature_or_function, **options)
 
     _locals = options.pop('locals', {})
+    # print("\nlocals: " + str(_locals))
```
Hardcode84: remove

hpat/decorators.py
```diff
@@ -93,4 +95,5 @@ def jit(signature_or_function=None, **options):
     # this is for previous version of pipeline manipulation (numba hpat_req <0.38)
     # from .compiler import add_hpat_stages
     # return numba.jit(signature_or_function, user_pipeline_funcs=[add_hpat_stages], **options)
+    # return numba.jit(signature_or_function, pipeline_class=hpat.compiler.HPATPipeline, **options)
```
Hardcode84: remove

hpat/distributed_analysis.py
```diff
@@ -68,6 +68,9 @@ def _init_run(self):
         self.second_pass = False
         self.in_parallel_parfor = -1
 
+    def run_pass(self, state):
```
Hardcode84: remove and call `run` directly

hpat/tests/test_series.py
```diff
@@ -67,9 +68,19 @@ class TestSeries(unittest.TestCase):
 
     def test_create1(self):
         def test_impl():
+            # a = 2
```
Hardcode84: remove

Hardcode84:
I assume we need to update numba version in our public ci as well

Hardcode84:
Also, disable tests broken by parfor issue as were disscussed

hpat/tests/test_series.py
```suggestion
```

PokhodenkoSA: @Hardcode84 
There is code suggestion tool in GitHub (ctrl+g)

hpat/tests/test_series.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 import pyarrow.parquet as pq
 import hpat
+import numba
```
```suggestion
import numba

```

hpat/tests/test_groupby.py
```diff
@@ -8,6 +8,8 @@
                                    count_parfor_OneDs, count_array_OneDs, dist_IR_contains,
                                    get_start_end)
 
+from numba.special import literally
```
PokhodenkoSA: Is it used somewhere?

hpat/distributed_analysis.py
PokhodenkoSA: I think run_pass(self, state) is interface function. It is better to inline run() here.

hpat/distributed.py
```diff
@@ -147,47 +151,51 @@ def __init__(self, func_ir, typingctx, targetctx, typemap, calltypes,
...
+        self.state.func_ir._definitions = build_definitions(self.state.func_ir.blocks)
         dist_analysis_pass = DistributedAnalysis(
-            self.func_ir, self.typemap, self.calltypes, self.typingctx,
-            self.metadata)
-        self._dist_analysis = dist_analysis_pass.run()
-        # dprint_func_ir(self.func_ir, "after analysis distributed")
+            self.state.func_ir, self.state.typemap, self.state.calltypes, self.state.typingctx,
+            self.state.metadata)
+        self._dist_analysis = dist_analysis_pass.run_pass(self.state)
+        # dprint_func_ir(self.state.func_ir, "after analysis distributed")
```
PokhodenkoSA:
```suggestion
```

hpat/compiler.py
```diff
@@ -150,145 +63,104 @@ def inline_calls(func_ir, _locals):
...
     func_ir.blocks = ir_utils.simplify_CFG(func_ir.blocks)
 
+#TODO: remove these helper functions when Numba provide appropriate way to manipulate passes
+def pass_position(pm, location):
+    assert pm.passes
+    pm._validate_pass(location)
+    for idx, (x, _) in enumerate(pm.passes):
+        if x == location:
+            return idx
+    else:
```
densmirn: `else:` is excess and could be removed because we don't have `break` in the loop. Let's leave only `raise`.

hpat/compiler.py
```diff
@@ -150,145 +63,104 @@ def inline_calls(func_ir, _locals):
...
+
+    # if a pass has been added, it's not finalized
+    pm._finalized = False
+
+@register_pass(mutates_CFG=True, analysis_only=False)
+class InlinePass(FunctionPass):
+    _name = "hpat_inline_pass"
+
+    def __init__(self):
+        pass
```
densmirn: Looks like it should work the same way without the stub.

hpat/compiler.py
```diff
@@ -150,145 +63,104 @@ def inline_calls(func_ir, _locals):
...
-        Convert HiFrames after typing
-        """
-        # Ensure we have an IR and type information.
-        assert self.func_ir
-        df_pass = HiFramesTyped(self.func_ir, self.typingctx,
-                                self.type_annotation.typemap,
-                                self.type_annotation.calltypes)
-        df_pass.run()
+    def __init__(self):
+        pass
```
densmirn: Looks like it should work the same way without the constructor, constructor of the parent class will be invoked anyway.

hpat/hiframes/aggregate.py
```diff
@@ -1047,51 +1047,54 @@ def gen_top_level_agg_func(key_names, return_key, red_var_typs, out_typs,
 
 
 def compile_to_optimized_ir(func, arg_typs, typingctx):
+    state = namedtuple('State',
+                            ['typingctx', 'targetctx', 'args', 'func_ir', 'typemap', 'return_type',
+                                'calltypes', 'metadata'])
```
densmirn: Just in case there is another way to define named tuple:
```
from typing import NamedTuple

class State(NamedTuple):
    typingctx: <type>
    targetctx: <type>
    args: <type>
    func_ir: <type>
    typemap: <type>
    return_type: <type>
    calltypes: <type>
    metadata: <type>
```

hpat/tests/test_series.py
AlexanderKalistratov: What is the difference?

hpat/compiler.py
AlexanderKalistratov: ```__init__``` is marked as an abstract method in the base class

hpat/compiler.py
AlexanderKalistratov: ```__init__``` is marked as an abstract method in the base class

hpat/hiframes/aggregate.py
AlexanderKalistratov: If you don't have strong opinion it should be modified, I'd prefer to left as is

hpat/tests/test_basic.py
```diff
@@ -297,6 +297,7 @@ def test_array_reduce(self):
             self.assertEqual(count_array_OneDs(), 0)
             self.assertEqual(count_parfor_OneDs(), 1)
 
+    @unittest.skipIf(numba.__version__.startswith('0.46.'), "Broken in numba 0.46.x")
```
Hardcode84: add url to issue https://github.com/numba/numba/issues/4690

And IMO better to check exact version, i.e. if it will be still broken on 0.46.1 we will see that

hpat/compiler.py
AlexanderKalistratov: fixed

hpat/tests/test_groupby.py
AlexanderKalistratov: no. Removed

hpat/tests/test_series.py
AlexanderKalistratov: Removed

hpat/tests/test_groupby.py
```diff
@@ -394,6 +396,25 @@ def test_impl(df):
         hpat_func = hpat.jit(test_impl)
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_df_nonliteral(self):
```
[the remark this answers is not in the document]
AlexanderKalistratov: Removed

hpat/distributed_analysis.py
AlexanderKalistratov: renamed

hpat/decorators.py
AlexanderKalistratov: Removed

hpat/decorators.py
AlexanderKalistratov: Removed

hpat/hiframes/dataframe_pass.py
```diff
@@ -35,24 +36,36 @@
...
+    def run_pass(self, state):
+        return DataFramePassImpl(state).run_pass()
+
+
+class DataFramePassImpl(object):
 
-    def run(self):
-        blocks = self.func_ir.blocks
+    def __init__(self, state):
+        self.state = state
```
shssf: `self.func_ir, self.typemap, self.typingctx = self`
?
no need to keep new variable in this case

hpat/distributed.py
```diff
@@ -70,63 +72,65 @@
...
-        """
+    _name = "distributed_pass"
+    """ The summary for the class or init method can be mentioned here
+
+    The details can be multiple lines. You can mention all the supported
+    and not supported args here or you can do it in ``Parameters`` section
+
+    .. only:: developer
+
+        The content for **Developer's Guide** should go here
```
shssf: It would be great to have any documentation

hpat/tests/test_basic.py
Hardcode84: Also, maybe create some helper function to check numba version to avoid copypaste

AlexanderKalistratov:
Split PR into 5 commits:
* Update globals for generated function
* Replace numba.compiler with numba.typed_pass
* Rename self.<something> to self.state.<something>
* Modify aggregate.py to support state
* All other changes

Most of commits would not work on numba 0.45 and doesn't work separately from each other, so all commits must be squashed before merge

hpat/utils.py
```diff
@@ -531,3 +532,7 @@ def dump_node_list(node_list):
 
 def debug_prints():
     return numba.config.DEBUG_ARRAY_OPT == 1
+
+def update_globals(func, glbls):
```
shssf: it looks like this function used nowhere. please delete

hpat/utils.py
```diff
@@ -18,6 +18,7 @@
 from hpat.str_ext import string_type, list_string_array_type
 from hpat.str_arr_ext import string_array_type, num_total_chars, pre_alloc_string_array
 from enum import Enum
+import types as pytypes
```
shssf: is it used in this module?

hpat/tests/test_utils.py
```diff
@@ -53,3 +53,6 @@ def get_start_end(n):
     start = hpat.distributed_api.get_start(n, n_pes, rank)
     end = hpat.distributed_api.get_end(n, n_pes, rank)
     return start, end
+
+def check_numba_version(version):
+    return numba.__version__ == version
```
shssf: `numba.__version__` is shorter than `check_numba_version`
may be it is better to use this variable from Numba directly inplace?

hpat/utils.py
AlexanderKalistratov: It is used in distributed, dataframe, high frame typad & untyped passes. It updates globals in generated functions. It was introduced in order to restore functionality of ```inline_closure_call``` which doesn't update globals anymore.

You can see details in related commit: https://github.com/IntelPython/hpat/pull/224/commits/4b828c921ee19b3b12667f573eaab4105a91e22a

hpat/tests/test_utils.py
AlexanderKalistratov: I don't know. For me both versions are fine, though version with function is a little bit better, since it doesn't looks so ugly and encapsulates details

AlexanderKalistratov:
Rebased & squashed

PR #227, opened 2019-10-15
Implement Series.prod() in new style

PokhodenkoSA:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1043,6 +1043,66 @@ def hpat_pandas_series_max_impl(self, axis=None, skipna=True, level=None, numeri
...
+    axis:
+        *unsupported*
+    skipna: :obj:`bool` object
+        Exclude nan values when computing the result
+    level:
+        *unsupported*
+    numeric_only:
+        *unsupported*
+    min_count:
+        *unsupported*
```
densmirn: Anyway I propose to describe the parameters and add type of these ones.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1043,6 +1043,66 @@ def hpat_pandas_series_max_impl(self, axis=None, skipna=True, level=None, numeri
...
+    :obj:
+         returns :obj: scalar
+    """
+
+    _func_name = 'Method prod().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(self.data.dtype, (types.Integer, types.Float)):
```
densmirn: I would use `self.dtype` instead of `self.data.dtype`, but perhaps it's also ok.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1043,6 +1043,66 @@ def hpat_pandas_series_max_impl(self, axis=None, skipna=True, level=None, numeri
...
+        raise TypingError('{} Currently function supports only numeric values. Given data type: {}'.format(_func_name, self.data.dtype))
+
+    if not (isinstance(skipna, (types.Omitted, types.Boolean)) or skipna is None):
```
densmirn: Please test the implementation by the following function:
```
def test_impl(S):
    return S.prod(skipna=None)
```

Does it work?

hpat/tests/test_series.py
```diff
@@ -1004,17 +1004,31 @@ def test_impl(S):
...
+            [1.1, 0.3, 2.1, 1, 3, 0.3, 2.1, 1.1, 2.2],
+            [6, 6.1, 2.2, 1, 3, 3, 2.2, 1, 2],
+            [6, 6, np.nan, 2, np.nan, 1, 3, 3, np.inf, 2, 1, 2, np.inf],
+            [1.1, 0.3, np.nan, 1.0, np.inf, 0.3, 2.1, np.nan, 2.2, np.inf],
+            [1.1, 0.3, np.nan, 1, np.inf, 0, 1.1, np.nan, 2.2, np.inf, 2, 2],
+            [np.nan, np.nan, np.nan],
+            [np.nan, np.nan, np.inf],
+        ]
+
+        for skipna in [True, False]:
```
densmirn: Will be the test passed if to use `None` as value of parameter `skipna`? Let's add `None` to the list of `skipna`'s`

hpat/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: I have updated docstring. I do not know if it is a good approach to copy documentation from Pandas because we will have to update it. Maybe information that we reimplement Pandas function will be enough. @densmirn what do you think?

densmirn: Phenix documentation should be generated according to docstrings. I don't know details, but think it would be better to have more complete generated documentation, perhaps it doesn't have to be copied from Pandas.

PokhodenkoSA: @densmirn could you, please, provide me a good example to orient on it?

densmirn: Your docsting seems pretty good with the following minor changes:
```
    self: :obj:`pandas.Series`
        input series
    axis: :obj:`int`
        Axis for the function to be applied on.
        *unsupported*
    skipna: :obj:`bool`, default :obj:`True`
        Exclude nan values when computing the result
    level: :obj:`int`, :obj:`str`, default :obj:`None`
        If the axis is a MultiIndex (hierarchical), count along a particular level, collapsing into a scalar.
        *unsupported*
    numeric_only: :obj:`bool`, default :obj:`None`
        Include only float, int, boolean columns.
        If None, will attempt to use everything, then use only numeric data.
        Not implemented for Series.
        *unsupported*
    min_count: :obj:`int`, default 0
        The required number of valid values to perform the operation.
        If fewer than min_count non-NA values are present the result will be NA.
        *unsupported*
```

hpat/hiframes/pd_dataframe_ext.py
```diff
@@ -1554,7 +1554,7 @@ def generic(self, args, kws):
         df = args[0]
         # TODO: ignore non-numerics
         # get series prod output types
-        dtypes = tuple(hpat.hiframes.pd_series_ext.SeriesAttribute.resolve_prod(
+        dtypes = tuple(numba.typing.arraydecl.ArrayAttribute.resolve_prod(
```
shssf: What issue you trying to solve by this change?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1074,6 +1074,75 @@ def hpat_pandas_series_max_impl(self, axis=None, skipna=True, level=None, numeri
     return hpat_pandas_series_max_impl
 
 
+@overload_method(SeriesType, 'prod')
+def hpat_pandas_series_prod(self, axis=None, skipna=None, level=None, numeric_only=None, min_count=0):
```
```suggestion
def hpat_pandas_series_prod(self, axis=None, skipna=True, level=None, numeric_only=None, min_count=0):
```
According to documentation:
skipna : bool, default True

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1085,21 +1085,30 @@ def hpat_pandas_series_prod(self, axis=None, skipna=None, level=None, numeric_on
 
     Parameters
     -----------
-    axis: {index (0)}
-        *unsupported* Axis for the function to be applied on.
-    skipna: :obj:`bool` object, default :obj:`True`
+    self: :obj:`pandas.Series`
+        input series
+    axis: :obj:`int`
```
densmirn: Actually axis could be a string (`'index'` or `'columns'`), not only an integer. You could add string type here. Moreover you could check on typing stage of the implementation that axis type is able to be a string.

hpat/hiframes/pd_dataframe_ext.py
PokhodenkoSA: `SeriesAttribute` have no `resolve_prod()` because it was not copied from `ArrayAttribute`.
[Here](https://github.com/IntelPython/hpat/blob/94cb552f3e2c11fe62523661fe6fa2f1af2d34da/hpat/hiframes/pd_series_ext.py#L992
) I added `'resolve_prod'` into `_not_series_array_attrs` and it prevents copying of corresponding method from `ArrayAttribute` to `SeriesAttribute` [here](https://github.com/IntelPython/hpat/blob/94cb552f3e2c11fe62523661fe6fa2f1af2d34da/hpat/hiframes/pd_series_ext.py#L998).

PokhodenkoSA: The issue is exception with 'attribute not found'

hpat/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: None in Python is usually used as default value because Python consider default values as global values ([see this](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)).
Example:
```
def f(a=[]):
  a.append(1)
  return a

>>> f()
[1]
>>> f()
[1,1]
```
So common approach is:
```
def f(a=None):
   """Docstring: a: list, default []"""
  if a is None:
    a = []
  ...
```

PokhodenkoSA: In this case with `bool` True is global and not mutable. But implementation of Pandas takes None as argument. Moreover Pandas is ready to get None so we also should be ready to get None and set it to True in the function.

So I think `skipna : bool, default True` is about logical value. It meas if you drop out `skipna` it will be considered as True.

hpat/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: I have created https://github.com/IntelPython/hpat/issues/235 for discussing documentation for parameters.

PR #228, opened 2019-10-15
Refactor Series.median() in a new style via np.median

kozlov-alexey:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,53 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_default1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_false1
+
+    Parameters
+    -----------
+    self: :obj:`pandas.Series`
+          input series
+    axis: axis for the function to be applied on, default None
```
densmirn: Please specify type of the parameter.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,53 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+
+    """
+
+    _func_name = 'Method median().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, (types.Integer, types.Float)):
```
densmirn: You can consider shortcut `types.Number` for `(types.Integer, types.Float)`.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,53 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_default1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_false1
+
+    Parameters
+    -----------
+    self: :obj:`pandas.Series`
+          input series
+    axis: axis for the function to be applied on, default None
+         *unsupported*
```
densmirn: Please add check of unsupported parameters and raise exception in case of parameter is really unsupported. Moreover you could cover such cases by unit tests.

hpat/tests/test_series.py
```diff
@@ -1711,6 +1712,35 @@ def test_impl(S):
         S = pd.Series(np.random.ranf(m))
         self.assertEqual(hpat_func(S), test_impl(S))
 
+    def test_series_median_skipna_default1(self):
+        '''Verifies median implementation with default skipna=True argument on a series with NA values'''
+        def test_impl(S):
+            res = S.median()
+            print(res)
```
densmirn: Please remove `print` if it was used for debugging. I don't think we need that here.

hpat/tests/test_series.py
```diff
@@ -1711,6 +1712,35 @@ def test_impl(S):
...
+            return res
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series([2., 3., 5., np.nan, 5., 6., 7.])
+        self.assertEqual(hpat_func(S, ), test_impl(S))
```
densmirn: typo: hpat_func(S, ) -> hpat_func(S)

hpat/tests/test_series.py
```diff
@@ -1711,6 +1712,35 @@ def test_impl(S):
...
+
+    def test_series_median_skipna_false1(self):
+        '''Verifies median implementation with skipna=False on a series with NA values'''
+        def test_impl(S):
+            res = S.median(skipna=False)
+            print(res)
```
densmirn: Is it debuggable `print`? If it's so please remove.

hpat/tests/test_series.py
```diff
@@ -1711,6 +1712,35 @@ def test_impl(S):
...
+        hpat_func = hpat.jit(test_impl)
+
+        # np.inf is not NaN, so verify that a correct number is returned
+        S1 = pd.Series([2., 3., 5., np.inf, 5., 6., 7.])
+        self.assertEqual(hpat_func(S1), test_impl(S1))
+
+        # TODO: both return values are 'nan', but HPAT's is not np.nan, hence checking with
+        # assertIs() doesn't work - check if it's Numba relatated
+        S2 = pd.Series([2., 3., 5., np.nan, 5., 6., 7.])
+        self.assertTrue(np.isnan(hpat_func(S2)) and np.isnan(test_impl(S2)))
```
```suggestion
        self.assertEqual(np.isnan(hpat_func(S2)), np.isnan(test_impl(S2)))
```
It will give more info in case of error.

hpat/tests/test_series.py
```diff
@@ -1711,6 +1712,35 @@ def test_impl(S):
...
+
+    @unittest.skip('HPAT distribution is not working (new-style impl issue)')
```
shssf: This is because you delete this ability (see above)

hpat/hiframes/series_kernels.py
```diff
@@ -528,7 +528,6 @@ def gt_f(a, b):
     'nsmallest_default': lambda A, name: hpat.hiframes.api.init_series(hpat.hiframes.api.nlargest(A, 5, False, lt_f), None, name),
     'head': lambda A, I, k, name: hpat.hiframes.api.init_series(A[:k], None, name),
     'head_index': lambda A, I, k, name: hpat.hiframes.api.init_series(A[:k], I[:k], name),
-    'median': lambda A: hpat.hiframes.api.median(A),
```
shssf: do not delete this. Comment it out if you need. In this PR this code is needed for parallelism.

hpat/hiframes/pd_series_ext.py
```diff
@@ -683,14 +683,6 @@ def resolve_head(self, ary, args, kws):
...
         return signature(ary, *args)
 
-    @bound_function("series.median")
-    def resolve_median(self, ary, args, kws):
-        assert not kws
-        dtype = ary.dtype
-        # median converts integer output to float
-        dtype = types.float64 if isinstance(dtype, types.Integer) else dtype
-        return signature(dtype, *args)
-
```
shssf: do not delete this. Comment it out if you need. In this PR this code is needed for parallelism.

hpat/hiframes/hiframes_typed.py
```diff
@@ -866,7 +866,7 @@ def _run_call_series(self, assign, lhs, rhs, series_var, func_name):
             return self._replace_func(func, [data], pre_nodes=nodes)
 
         if func_name in ('std', 'nunique', 'describe', 'isna',
-                         'isnull', 'median', 'idxmin', 'idxmax', 'unique'):
+                         'isnull', 'idxmin', 'idxmax', 'unique'):
```
shssf: usually I do following https://github.com/IntelPython/hpat/pull/189/files#diff-f13c21aac60fb8f7ae3c9527239d1e17R990
instead removing this name from this loop

hpat/datatypes/hpat_pandas_series_functions.py
```suggestion
    if not isinstance(self._data.dtype, (types.Integer, types.Float)):
```

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,53 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, (types.Integer, types.Float)):
+        raise TypingError(
+            '{} The function only applies to elements that are all numeric. Given data type: {}'.format(_func_name, self.dtype))
+
+    def hpat_pandas_series_median_impl(self, axis=None, skipna=True, level=None, numeric_only=None):
+        if skipna:
```
shssf: I see no checks for `skipna` type. What happened if it is `None`?

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Minor correctness: `self._data.dtype` -> `self.data.dtype`

hpat/hiframes/hiframes_typed.py
densmirn: `_not_series_array_attrs = ['resolve_median', ...]` just avoids replacement of `series.median` with `array.median`. As I see we are not able to keep both approaches just by adding `'resolve_median'` to the list. Please correct me if I'm wrong.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: @shssf But why is that better (that's less typing and they should always be the same)?

hpat/hiframes/pd_series_ext.py
kozlov-alexey: @shssf But that's just typing the same we've got from the overload. Keeping it has no sense - if we don't use old-style implementation from series_kernels.py, i.e. hpat.hiframes.api.median.

hpat/hiframes/hiframes_typed.py
kozlov-alexey: @shssf @densmirn No, if we keep 'median' here we'll use and invoke the old-style implementation (so commenting hpat.hiframes.api.median in series_kernels would cause lowering error). It's not possible to keep both implementations (unless one or the other is selected basing on configured pipeline). But anyway, in the new-style the parallel test is broken, even if I call the old-style hpat.hiframes.api.median function from the overload itself.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Done.

hpat/tests/test_series.py
kozlov-alexey: Corrected.

hpat/tests/test_series.py
kozlov-alexey: Thank you.

hpat/tests/test_series.py
kozlov-alexey: Removed.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Corrected by adding checks.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Thank you.

hpat/tests/test_series.py
kozlov-alexey: Done.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,67 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_default1
+       Test: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_median_skipna_false1
+
+    Parameters
+    -----------
+    self: :obj:`pandas.Series`
+          input series
+    axis: {0 or `index`, None}, default None
```
densmirn: Let's specify type of the parameter as `:obj:<type>`.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,67 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+
+    """
+
+    _func_name = 'Method median().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, types.Number):
```
densmirn: @shssf eventually what do you think about extracting `dtype` from `self`? `self.dtype` or `self.data.dtype`. BTW for me `dtype` means "data type", `data.dtype` means "data data dtype". I vote for shorter option.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,67 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+        raise TypingError(
+            '{} The function only applies to elements that are all numeric. Given data type: {}'.format(_func_name, self.dtype))
+
+    if not (isinstance(axis, (types.Integer, types.UnicodeType, types.Omitted)) or axis is None):
+        raise TypingError('{} The axis must be an Integer or a String. Currently unsupported. Given: {}'.format(_func_name, axis))
+
+    if not (isinstance(skipna, (types.Boolean, types.Omitted)) or skipna == True):
```
densmirn: Do you exactly need to check `skipna == True`?

hpat/hiframes/hiframes_typed.py
```diff
@@ -879,6 +879,9 @@ def _run_call_series(self, assign, lhs, rhs, series_var, func_name):
                 data = self._get_series_data(series_var, nodes)
             return self._replace_func(func, [data], pre_nodes=nodes)
 
+        if func_name == 'median':
+            return [assign]
```
densmirn: You will achieve the line never, because `'median'` is processed above. Moreover if `func_name` doesn't match any condition in this function `return [assign]` is by default.

kozlov-alexey: @densmirn Yes, it was needed only when 'median' was removed from the tuple above. 
It'll be removed in the next commit.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Yes, without it the check will fail, because the check is passed several times during typing. So if skipna argument is omitted during one pass skipna will have types.Omitted and during the other pass it will have type(skipna)=bool, hence the second part of the check is needed to work properly and not raise exception.

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Done .

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1547,3 +1547,67 @@ def hpat_pandas_series_nunique_impl(self, dropna=True):
...
+
+    """
+
+    _func_name = 'Method median().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError(
+            '{} The object must be a pandas.series. Given self: {}'.format(_func_name, self))
+
+    if not isinstance(self.dtype, types.Number):
```
```suggestion
    if not isinstance(self.data.dtype, types.Number):
```

hpat/datatypes/hpat_pandas_series_functions.py
shssf: @densmirn I think this is a bug in Numba. Without second check it will not work.

PR #230, opened 2019-10-15
Added tests on index handling for Series

akharche:


hpat/tests/test_series.py
```diff
@@ -411,6 +411,32 @@ def test_impl(A):
...
+        '''Verifies Series.astype implementation with function 'str' as argument
+           handles string series not changing it
+        '''
+
+        def test_impl(S):
+            return S.astype(str)
+
+        hpat_func = hpat.jit(test_impl)
+
+        S = pd.Series(['aa', 'bb', 'cc'], index=[2, 3, 5])
```
shssf:
```suggestion
        for input_data in (['aa', 'bb', 'cc'], [1,2,3]):
            for input_index in (['a', 'b', 'c'], [2,3,5]):
                S = pd.Series(input_data, index=input_index)
```

shssf: Also, it might be godd to play with None, NaN and Inf

PR #231, opened 2019-10-17
Adding support for getting Series.ndim and Series.T attributes

kozlov-alexey:


hpat/hiframes/hiframes_typed.py
```diff
@@ -356,23 +356,14 @@ def _run_getattr(self, assign, rhs):
             if rhs.attr in hpat.hiframes.pd_timestamp_ext.timedelta_fields:
                 return self._run_Timedelta_field(assign, assign.target, rhs)
 
-        if isinstance(rhs_type, SeriesType) and rhs.attr in ('size', 'shape'):
+        if isinstance(rhs_type, SeriesType) and rhs.attr in ('size'):
```
densmirn: `rhs.attr in ('size')` checks sub-string in string `'size'`, because `('size')` is a string, not a tuple, e.g. `('size',)`. It's not entirely correct. In case of `rhs.attr` is part of `'size'` we get strange behavior. I propose to use condition `rhs.attr == 'size'` here.

hpat/tests/test_series.py
```diff
@@ -256,6 +256,27 @@ def test_impl(A):
...
+
+        n = 11
+        S = pd.Series(np.arange(n))
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+    def test_series_getattr_T(self):
+        def test_impl(S):
+            res = S.T
+#             print(res)
+            return res
```
densmirn: Why not return `S.T`?

hpat/tests/test_series.py
```diff
@@ -256,6 +256,27 @@ def test_impl(A):
...
+        hpat_func = hpat.jit(test_impl)
+
+        n = 11
+        S = pd.Series(np.arange(n))
+        self.assertEqual(hpat_func(S), test_impl(S))
+
+    def test_series_getattr_T(self):
+        def test_impl(S):
+            res = S.T
+#             print(res)
```
densmirn: Please remove commented out `print`.

hpat/tests/test_series.py
```diff
@@ -256,6 +256,27 @@ def test_impl(A):
...
+            return res
+        hpat_func = hpat.jit(test_impl)
+
+        n = 11
+        S = pd.Series(np.arange(n))
+#         test_impl(S)
```
densmirn: I think the line could be removed.

hpat/hiframes/hiframes_typed.py
kozlov-alexey: Fixed, thank you.

hpat/tests/test_series.py
kozlov-alexey: Somehow overlooked

hpat/tests/test_series.py
kozlov-alexey: Yeap

hpat/tests/test_series.py
kozlov-alexey: Yeap

PR #241, opened 2019-10-18
Implement Series.nsmallest()/Series.nlargest() in new style

densmirn:


hpat/tests/test_series.py
```diff
@@ -1685,35 +1740,26 @@ def test_impl(S):
...
+        def test_impl():
+            series = pd.Series([1.0, np.nan, -1.0, 0.0, 5e-324])
+            return series.nsmallest(4)
         hpat_func = hpat.jit(test_impl)
 
-        m = 100
-        np.random.seed(0)
-        S = pd.Series(np.random.randint(-30, 30, m))
-        np.testing.assert_array_equal(hpat_func(S).values, test_impl(S).values)
+        np.testing.assert_array_equal(hpat_func(), test_impl())
```
PokhodenkoSA: [Pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.nsmallest.html): nsmallest returns Series.
Consider use pandas.testing.assert_series_equal(). It knows how to compare series better.
```suggestion
        pandas.testing.assert_series_equal(hpat_func(), test_impl())
```

hpat/tests/test_series.py
```diff
@@ -1685,35 +1740,26 @@ def test_impl(S):
...
+    def test_series_nsmallest_unboxing(self):
+        def test_impl(series):
+            return series.nsmallest()
         hpat_func = hpat.jit(test_impl)
 
-        S = pd.Series([1.0, np.nan, 3.0, 2.0, np.nan, 4.0])
-        np.testing.assert_array_equal(hpat_func(S).values, test_impl(S).values)
+        for data in test_global_input_data_numeric + [[]]:
+            series = pd.Series(data * 3)
+            np.testing.assert_array_equal(hpat_func(series), test_impl(series))
```
```suggestion
            pandas.testing.assert_series_equal(hpat_func(series), test_impl(series))
```

hpat/tests/test_series.py
```diff
@@ -34,6 +36,23 @@
     ),
 ]]
 
+test_global_input_data_float64 = [
+    [1.0, np.nan, -1.0, 0.0, 5e-324],
+    [np.nan, np.inf, np.NINF, np.NZERO]
+]
+
+min_int64 = -9223372036854775808
+max_int64 = 9223372036854775807
```
shssf: sys.maxint

hpat/tests/test_series.py
```diff
@@ -34,6 +36,23 @@
     ),
 ]]
 
+test_global_input_data_float64 = [
+    [1.0, np.nan, -1.0, 0.0, 5e-324],
```
shssf:
```suggestion
    [1.0, np.nan, -1.0, 0.0, sys.float_info.min],
```

```
sys.float_info
sys.floatinfo(max=1.7976931348623157e+308, max_exp=1024, max_10_exp=308, min=2.2
250738585072014e-308, min_exp=-1021, min_10_exp=-307, dig=15, mant_dig=53, epsil
on=2.2204460492503131e-16, radix=2, rounds=1)
```

hpat/tests/test_series.py
```diff
@@ -1723,7 +1769,52 @@ def test_impl():
...
+                     'Series.nsmallest() parameter keep unsupported')
+    def test_series_nsmallest_full(self):
+        def test_impl(series, n, keep):
+            return series.nsmallest(n=n, keep=keep)
+        hpat_func = hpat.jit(test_impl)
+
+        keep = 'first'
+        for data in test_global_input_data_numeric + [[]]:
+            series = pd.Series(data * 3)
+            for n in range(-1, 10):
```
shssf: We need this iteration over `n` unskipped somewhere.

hpat/hiframes/api.py
```diff
@@ -559,13 +559,13 @@ def select_k_nonan_overload(A, m, k):
     dtype = A.dtype
     if isinstance(dtype, types.Integer):
         # ints don't have nans
-        return lambda A, m, k: (A[:k].copy(), k)
+        return lambda A, m, k: (A[:max(k, 0)].copy(), k)
```
shssf: not sure we have to cut off negatives here

densmirn: `nsmallest(0) == nsmallest(-1)` returns empty series. So we need to set up `0` instead of negative `k` to return empty array. E.g. `k = -2`, `A = [1, 2, 3, 4, 5]`:
`A[:k] # [1, 2, 3]`
`A[:0] # [] - what we want to get when we have negative k`

densmirn: BTW old functionality didn't work with negative `k`.

hpat/tests/test_series.py
densmirn: @PokhodenkoSA You are right, but functionality "in old style" doesn't work with indexes correctly. So until we completely moved to "new style" the check is valid. Now we don't drop code "in old style" to keep MPI parallelism.

hpat/tests/test_series.py
densmirn: @PokhodenkoSA You are right, but functionality "in old style" doesn't work with indexes correctly. As I see `np.testing.assert_array_equal` checks only data inside of series.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -173,6 +173,114 @@ def hpat_pandas_series_shape_impl(self):
...
+
+    _func_name = 'Method nsmallest().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(n, (types.Omitted, int, types.Integer)):
+        raise TypingError('{} The object must be an integer. Given n: {}'.format(_func_name, n))
+
+    if not isinstance(keep, (types.Omitted, str, types.UnicodeType)):
```
kozlov-alexey: I think we don't need this, as it checks correct type of supported argument and keep seems to be unsupported.
We only need to check that it was omitted argument (hence it should have either types.Omitted type or str type and 'first' value), i.e. I suggest using following check at typing step:
```
    if not (keep == 'first' or isinstance(keep, types.Omitted)):
        raise TypingError('{} Unsupported parameters. Given keep: {}'.format(_func_name, keep))
```

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -173,6 +173,114 @@ def hpat_pandas_series_shape_impl(self):
...
+    keep: :obj:`str`, default 'first'
+        When there are duplicate values that cannot all fit in a Series of n elements:
+        first : return the first n occurrences in order of appearance.
+        last : return the last n occurrences in reverse order of appearance.
+        all : keep all occurrences. This can result in a Series of size larger than n.
+        *unsupported*
+
+    Returns
+    -------
+    :obj:`scalar`
```
kozlov-alexey: Why scalar?

hpat/tests/test_series.py
kozlov-alexey: @densmirn A better solution would be to change the test as @PokhodenkoSA suggested and skip it if executed with old-style (as there's a bug which has to be fixed anyway), while new-style impl can be written so that it handles index correctly from the start.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -173,6 +173,114 @@ def hpat_pandas_series_shape_impl(self):
...
+            raise TypingError('Method nsmallest(). Unsupported parameters. Given keep != \'first\'')
+
+        local_data = hpat.hiframes.api.nlargest_parallel(self._data, n, False, hpat.hiframes.series_kernels.lt_f)
+        local_index = []
+        for a in numpy.unique(local_data):
+            for indices in numpy.where(self._data == a):
+                local_index.extend(indices)
+        local_index = local_index[:n]
+
+        return pandas.Series(local_data, local_index)
```
kozlov-alexey: Don't we have to copy series name as well?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -173,6 +173,114 @@ def hpat_pandas_series_shape_impl(self):
...
+    def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
+        if keep != 'first':
+            raise TypingError('Method nsmallest(). Unsupported parameters. Given keep != \'first\'')
+
+        local_data = hpat.hiframes.api.nlargest_parallel(self._data, n, False, hpat.hiframes.series_kernels.lt_f)
+        local_index = []
+        for a in numpy.unique(local_data):
+            for indices in numpy.where(self._data == a):
+                local_index.extend(indices)
+        local_index = local_index[:n]
```
kozlov-alexey: @densmirn This looks inefficient as we call numpy.where on the self._data up to n times. 
I think better would be to use dict/set of unique values to look for and iterate over self._data only once.
If the self._data[i] is in that dict/set we can append i to a list of indexes where it was found. Then we will need to merge all such lists into one local_index and cut off first n elements as before. What do you think?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -173,6 +173,114 @@ def hpat_pandas_series_shape_impl(self):
...
+    _func_name = 'Method nsmallest().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(n, (types.Omitted, int, types.Integer)):
+        raise TypingError('{} The object must be an integer. Given n: {}'.format(_func_name, n))
+
+    if not isinstance(keep, (types.Omitted, str, types.UnicodeType)):
+        raise TypingError('{} The object must be an unicode. Given keep: {}'.format(_func_name, keep))
```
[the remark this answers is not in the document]
densmirn: In this case if to pass explicitly parameter `keep`=`'first'` as argument the exception will be raised, because `keep` will be equal to `unicode_type` and the condition will be `False`.

hpat/tests/test_series.py
densmirn: If I skipped the tests it would be no tests "in old style" at all. Or do you propose to duplicate the tests for "old style" and "new style" code with different asserts?

hpat/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
kozlov-alexey: @densmirn True, but why do this if this parameter is unsupported? If it's unsupported then the user should not provide any value for it (and hence only default value can be checked).

densmirn:
@shssf, @kozlov-alexey could you re-review the PR?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -165,6 +200,174 @@ def hpat_pandas_series_shape_impl(self):
...
+        inner_list_type = types.ListType(types.intp)
+        def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
+            if keep != 'first':
+                raise ValueError("Method nsmallest(). Unsupported parameter. "
+                                 "Given 'keep' != 'first'")
+
+            nlargest = hpat.hiframes.api.nlargest(self._data, n, False,
+                                                  hpat.hiframes.series_kernels.lt_f)
+            indices = Dict.empty(dtype, inner_list_type)
+            for item in set(nlargest):
```
kozlov-alexey:
```suggestion
            for item in nlargest:
```

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -165,6 +200,174 @@ def hpat_pandas_series_shape_impl(self):
...
+                                              hpat.hiframes.series_kernels.lt_f)
+        indices = Dict.empty(dtype, inner_list_type)
+        for item in set(nlargest):
+            indices[item] = List.empty_list(index_type)
+
+        for idx, item in enumerate(self._data):
+            if item in indices:
+                indices[item].append(self._index[idx])
+
+        local_index = [idx for item in sorted(indices) for idx in indices[item]]
```
kozlov-alexey: In theory we can have situation when all our Big Array is filled with 5 numbers and with this loop comprehension we will have to iterate over it once again, i.e. what we lack is the ability to leave early after we found exactly n first indices. What do you think if we use explicit loop here?
```suggestion
            local_index = []
            for item in nsmallest:
                for index in indices[item]:
                    local_index.append(index)
                    if len(local_index) == n:
                        break
                else:
                    continue
                break
```

densmirn: Done.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Done.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -188,33 +223,63 @@ def hpat_pandas_series_nsmallest(self, n=5, keep='first'):
...
 
-    if not isinstance(n, (types.Omitted, int, types.Integer)):
-        raise TypingError('{} The object must be an integer. Given n: {}'.format(_func_name, n))
+            for idx, item in enumerate(self._data):
+                if item in indices:
+                    indices[item].append(idx)
 
-    if not isinstance(keep, (types.Omitted, str, types.UnicodeType)):
-        raise TypingError('{} The object must be an unicode. Given keep: {}'.format(_func_name, keep))
+            local_index = [idx for item in sorted(indices) for idx in indices[item]]
```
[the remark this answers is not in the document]
densmirn: Done.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -188,33 +223,63 @@ def hpat_pandas_series_nsmallest(self, n=5, keep='first'):
...
+
+    dtype = self.data.dtype
+    if isinstance(self.index, types.NoneType):
+        inner_list_type = types.ListType(types.intp)
+        def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
+            if keep != 'first':
+                raise ValueError("Method nsmallest(). Unsupported parameter. "
+                                 "Given 'keep' != 'first'")
+
+            nlargest = hpat.hiframes.api.nlargest(self._data, n, False,
```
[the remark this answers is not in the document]
densmirn: Done.

densmirn:
@kozlov-alexey could you please take the next round of review?

hpat/tests/test_series.py
shssf: @densmirn please use
```
if hpat.config.config_pipeline_hpat_default:
else
  pd.testing.assert_series_equal(hpat_func(), test_impl())

```

hpat/tests/test_series.py
```diff
@@ -2104,7 +2179,84 @@ def test_impl():
...
+
+        # TODO: check data == [] when index fixed
+        for data in test_global_input_data_numeric:
+            data *= 3
+            for index in [gen_srand_array(len(data)), range(len(data))]:
+                series = pd.Series(data, index)
+                for n in range(-1, 10):
+                    ref_result = test_impl(series, n)
+                    jit_result = hpat_func(series, n)
+                    pd.testing.assert_series_equal(ref_result, jit_result)
```
shssf: you can use `np.testing.assert_array_equal(ref_result.value, jit_result.value)` to check for "old style"

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -43,6 +44,40 @@
 from hpat.utils import to_array
 
 
+class TypingChecker:
+    msg_template = '{} The object must be {}. Given{}: {}'
+    integer_types = (types.Omitted, int, types.Integer)
```
shssf: types.Omitted is not Integer

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -43,6 +44,40 @@
 from hpat.utils import to_array
 
 
+class TypingChecker:
```
shssf: please add doc string with example of output

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -43,6 +44,40 @@
...
+    msg_template = '{} The object must be {}. Given{}: {}'
```
```suggestion
    msg_template = '{__func_name__} The object {__param__}\n  given: {__given_type__}\n  expected: {__expected_type__}'
```
or something like this to make it more clear.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -43,6 +44,40 @@
...
+
+    def check_series(self, data):
+        if not isinstance(data, SeriesType):
+            raise TypingError(self.msg('a pandas.series', data))
+
+    def check_integer(self, data, name):
+        if not isinstance(data, (types.Omitted, int, types.Integer)):
+            raise TypingError(self.msg('an integer', data, name=name))
+
+    def check_float(self, data, name):
```
shssf: all such checks uses predefined list of checked types. This is ambiguous.  Also, types check might be complex (type1 && type2 || type3).
Please do not implement a function for particular check like `check_float`. Please use a list of types as a parameter.
Types or equation will be provided as a parameter.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+        all : keep all occurrences. This can result in a Series of size larger than n.
+        *unsupported*
+
+    Returns
+    -------
+    :obj:`series`
+         returns :obj:`series`
+    """
+
+    _func_name = 'Method nsmallest().'
```
shssf: move to 201

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+    ty_checker.check_series(self)
+    ty_checker.check_integer(n, 'n')
+    ty_checker.check_string(keep, 'keep')
+
+    dtype = self.data.dtype
+    if isinstance(self.index, types.NoneType):
+        inner_list_type = types.ListType(types.intp)
+        def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
+            if keep != 'first':
+                raise ValueError("Method nsmallest(). Unsupported parameter. "
```
shssf: put it in one line. Also, I think (not sure) this check is better to do in type checking stage (out of this nested function)

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+
+    ty_checker = TypingChecker(_func_name)
+    ty_checker.check_series(self)
+    ty_checker.check_integer(n, 'n')
+    ty_checker.check_string(keep, 'keep')
+
+    dtype = self.data.dtype
+    if isinstance(self.index, types.NoneType):
+        inner_list_type = types.ListType(types.intp)
+        def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
```
shssf: put blank line before

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+                        break
+                    local_index.append(index)
+
+            return pandas.Series(nsmallest, local_index, name=self._name)
+
+        return hpat_pandas_series_nsmallest_impl
+
+    index_type = self.index.dtype
+    inner_list_type = types.ListType(index_type)
+    def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
```
shssf: make function names diffrent for diffrent cases

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+                    indices[item] = List.empty_list(types.intp)
+                indices[item].append(idx)
+
+            max_length = len(nsmallest)
+            local_index = []
+            for item in numpy.unique(nsmallest):
+                for index in indices[item]:
+                    if len(local_index) == max_length:
+                        break
+                    local_index.append(index)
```
shssf: I suspect this algorithm might be replaced by `numpy.argsort`

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+
+    dtype = self.data.dtype
+    if isinstance(self.index, types.NoneType):
+        inner_list_type = types.ListType(types.intp)
+        def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
+            if keep != 'first':
+                raise ValueError("Method nsmallest(). Unsupported parameter. "
+                                 "Given 'keep' != 'first'")
+
+            nsmallest = hpat.hiframes.api.nlargest(self._data, n, False,
```
shssf: Why do we need to call it? Put doc string here.

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +173,182 @@ def hpat_pandas_series_iloc_impl(self):
...
+
+            max_length = len(nsmallest)
+            local_index = []
+            for item in numpy.unique(nsmallest):
+                for index in indices[item]:
+                    if len(local_index) == max_length:
+                        break
+                    local_index.append(index)
+
+            return pandas.Series(nsmallest, local_index, name=self._name)
```
shssf: I'm afraid you changing name of the type, not label the Series here.
To set a name to the Series object you have to implement parameter `name` in Series::ctor(). Not sure it will work or not because base class `types.IterableType` has the same name of a parameter. In this case it might be required to change field name to `label`.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: I replaced the algorithm with `numpy.argsort` for `nsmallest`. I'm not able to apply that for `nlargest` due to descending ordering of the result (`[::-1]` works incorrectly).

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Added comment.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Done.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Done.

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: I think it's ok to validate values inside of jit region. At the least we can check if passed incorrect value which couldn't be checked on typing stage, e.g. `keep == 'smth_invalid'`.

hpat/tests/test_series.py
```diff
@@ -2007,35 +2017,29 @@ def test_impl(S):
...
-        np.random.seed(0)
-        S = pd.Series(np.random.randint(-30, 30, m))
-        np.testing.assert_array_equal(hpat_func(S).values, test_impl(S).values)
-
-    def test_series_nlargest_default1(self):
-        def test_impl(S):
-            return S.nlargest()
+    def test_series_nlargest(self):
+        def test_impl():
+            series = pd.Series([1.0, np.nan, -1.0, 0.0, 5e-324])
```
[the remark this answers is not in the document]
densmirn: Done.

hpat/tests/test_series.py
```diff
@@ -2066,35 +2147,29 @@ def test_impl(S):
...
-        np.random.seed(0)
-        S = pd.Series(np.random.randint(-30, 30, m))
-        np.testing.assert_array_equal(hpat_func(S).values, test_impl(S).values)
-
-    def test_series_nsmallest_default1(self):
-        def test_impl(S):
-            return S.nsmallest()
+    def test_series_nsmallest(self):
+        def test_impl():
+            series = pd.Series([1.0, np.nan, -1.0, 0.0, 5e-324])
```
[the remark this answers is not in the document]
densmirn: Done.

hpat/tests/test_series.py
```diff
@@ -67,6 +67,16 @@
 ]
 
 
+def gen_srand_array(size, nchars=8):
+    """Generate array of strings of specified size based on [a-zA-Z] + [0-9]"""
+    accepted_chars = list(string.ascii_letters + string.digits)
+    rands_chars = np.array(accepted_chars, dtype=(np.str_, 1))
+
+    np.random.seed(100)
+    return np.random.choice(rands_chars,
```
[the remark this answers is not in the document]
densmirn: Done.

hpat/hiframes/pd_series_ext.py
```diff
@@ -1006,6 +1006,12 @@ def generic_expand_cumulative_series(self, args, kws):
             and attr not in _not_series_array_attrs):
         setattr(SeriesAttribute, attr, func)
 
+# remove some attributes from SeriesAttribute for non-hpat pipeline
+if not hpat.config.config_pipeline_hpat_default:
+    _non_hpat_pipeline_attrs = ['resolve_nsmallest', 'resolve_nlargest']
```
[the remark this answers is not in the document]
densmirn: Done.

hpat/tests/test_series.py
```diff
@@ -2045,7 +2049,84 @@ def test_impl():
...
+            data *= 3
+            for index in [gen_srand_array(len(data)), range(len(data))]:
+                series = pd.Series(data, index)
+                for n in range(-1, 10):
+                    ref_result = test_impl(series, n)
+                    jit_result = hpat_func(series, n)
+                    pd.testing.assert_series_equal(ref_result, jit_result)
+
+    @unittest.skipIf(hpat.config.config_pipeline_hpat_default,
+                     'Series.nlargest() types validation unsupported')
```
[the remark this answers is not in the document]
densmirn: This functionality is unsupported in "old style" (in HPAT pipeline).

hpat/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: @shssf Why to use argsort here? It will sort the whole array whereas we need just first n elements?

densmirn: Now I'm not able to explain it, but looks like implementation with using of `argsort` is up to 3 times faster on laptop on 4 threads.

name | type | size | n | median | min | max
-- | -- | -- | -- | -- | -- | --
series_nsmallest | JIT | 10002024 | 16 | 1,928377 | 1,914718 | 1,956772
  |   |   | 64 | 1,878858 | 1,866676 | 1,910763
  |   |   | 512 | 1,869665 | 1,853625 | 2,024022
  |   |   | 1024 | 1,877665 | 1,844636 | 1,878451
series_nsmallest   (with arsort) | JIT | 10002024 | 16 | 0,722629 | 0,717613 | 0,76169
  |   |   | 64 | 0,725377 | 0,71963 | 0,765624
  |   |   | 512 | 0,738325 | 0,725678 | 0,743378
  |   |   | 1024 | 0,745736 | 0,730581 | 0,759536

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Fixed.

densmirn:
@shssf could you take the next round of review?

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Done.

hpat/tests/test_series.py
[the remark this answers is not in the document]
densmirn: We are not able to catch specific `TypingError` in "old style".

hpat/tests/test_series.py
```diff
@@ -2045,7 +2049,84 @@ def test_impl():
...
+        for keep in ['last', 'all', '']:
+            with self.assertRaises(ValueError) as raises:
+                hpat_func(series, n=5, keep=keep)
+            self.assertIn(msg, str(raises.exception))
+
+        with self.assertRaises(ValueError) as raises:
+            hpat_func(series, n=5, keep='last')
+        self.assertIn(msg, str(raises.exception))
 
     @unittest.skip('Unsupported functionality: failed to handle index')
```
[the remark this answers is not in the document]
densmirn: The test still doesn't work in "old style".

densmirn:
@shssf I moved TypeChecker from this PR to another one. So could you please re-review this PR?

hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -138,6 +138,103 @@ def hpat_pandas_series_iloc_impl(self):
...
+    if not isinstance(keep, (types.Omitted, str, types.UnicodeType, types.StringLiteral)):
+        raise TypingError('{} The object keep\n given: {}\n expected: {}'.format(_func_name, keep, 'str'))
+
+    def hpat_pandas_series_nlargest_impl(self, n=5, keep='first'):
+        if keep != 'first':
+            raise ValueError("Method nlargest(). Unsupported parameter. Given 'keep' != 'first'")
+
+        # data: [0, 1, -1, 1, 0] -> [1, 1, 0, 0, -1]
+        # index: [0, 1,  2, 3, 4] -> [1, 3, 0, 4,  2] (not [3, 1, 4, 0, 2])
+        indices = (-self._data - 1).argsort(kind='mergesort')[:max(n, 0)]
```
shssf: Quite strange algorithm.

hpat/hiframes/pd_series_ext.py
```diff
@@ -999,13 +999,20 @@ def generic_expand_cumulative_series(self, args, kws):
...
     if (attr.startswith('resolve_')
             and attr not in SeriesAttribute.__dict__
             and attr not in _not_series_array_attrs):
         setattr(SeriesAttribute, attr, func)
 
+# remove some attributes from SeriesAttribute for non-hpat pipeline
+if not hpat.config.config_pipeline_hpat_default:
+    for attr in _non_hpat_pipeline_attrs:
+        if attr in SeriesAttribute.__dict__:
+            delattr(SeriesAttribute, attr)
```
shssf: I still don't understand why we need to remove attributes after adding them. I still think it would be better to merge `_non_hpat_pipeline_attrs` with `_not_series_array_attrs` and remove this piece of code.

hpat/tests/test_series.py
```diff
@@ -73,6 +73,15 @@
 ]
 
 
+def gen_srand_array(size, nchars=8):
+    """Generate array of strings of specified size based on [a-zA-Z] + [0-9]"""
+    accepted_chars = list(string.ascii_letters + string.digits)
+    rands_chars = np.array(accepted_chars, dtype=(np.str_, 1))
+
+    np.random.seed(100)
+    return np.random.choice(rands_chars, size=nchars * size).view((np.str_, nchars))
```
shssf: I think this is a bad idea to use random input data. If some test failed by some input data - it would be more difficult to reproduce the issue.

hpat/tests/test_series.py
```diff
@@ -2063,56 +2072,117 @@ def test_impl():
...
+            data *= 3
+            for index in [gen_srand_array(len(data)), range(len(data))]:
+                series = pd.Series(data, index)
+                for n in range(-1, 10):
+                    ref_result = test_impl(series, n)
+                    jit_result = hpat_func(series, n)
+                    pd.testing.assert_series_equal(ref_result, jit_result)
 
+    @unittest.skipIf(hpat.config.config_pipeline_hpat_default,
+                     'Series.nlargest() types validation unsupported')
```
shssf:
```suggestion
                     'Series.nlargest() do not throw an exception')
```

hpat/tests/test_series.py
```diff
@@ -2122,26 +2192,87 @@ def test_impl():
...
+    @unittest.skipIf(hpat.config.config_pipeline_hpat_default,
+                     'Series.nsmallest() index unsupported')
+    def test_series_nsmallest_index(self):
+        def test_impl(series, n):
+            return series.nsmallest(n)
+        hpat_func = hpat.jit(test_impl)
+
+        # TODO: check data == [] when index fixed
+        for data in test_global_input_data_numeric:
+            data *= 3
```
shssf: Why this is?

hpat/tests/test_series.py
shssf: I see this test is also dosn't work with index of strings. Also, as far s i remember, old-style is working with index of strings.
Anyway, you could use if/else as you did in other tests to make diffrent checks for diffrent styles. Please note, I proposed `ref_result.value` which returns array (not series)

hpat/tests/test_series.py
```diff
@@ -2122,26 +2192,87 @@ def test_impl():
...
+                    ref_result = test_impl(series, n)
+                    jit_result = hpat_func(series, n)
+                    pd.testing.assert_series_equal(ref_result, jit_result)
 
-    @unittest.skip('Unsupported functionality: failed to handle index')
-    def test_series_nsmallest_index_int(self):
-        def test_impl(S):
-            return S.nsmallest(3)
+    @unittest.skipIf(hpat.config.config_pipeline_hpat_default,
+                     'Series.nsmallest() types validation unsupported')
```
shssf:
```suggestion
                     'Series.nsmallest() do not throw an exception')
```

hpat/tests/test_series.py
```diff
@@ -2122,26 +2192,87 @@ def test_impl():
...
+
+        for keep, dtype in [(True, types.boolean), (None, types.none),
+                            (0.1, 'float64'), (1, 'int64')]:
+            with self.assertRaises(TypingError) as raises:
+                hpat_func(series, n=5, keep=keep)
+            msg = '{} The object keep\n given: {}\n expected: str'
+            self.assertIn(msg.format(_func_name, dtype), str(raises.exception))
 
+    @unittest.skipIf(hpat.config.config_pipeline_hpat_default,
+                     'Series.nsmallest() values validation unsupported')
```
shssf:
```suggestion
                     'Series.nsmallest() do not throw an exception')
```

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: `(-self._data - 1)`: subtracted 1 to ensure reverse ordering at boundaries, e.g. to turn min into max integer. `self._data.argsort(kind='mergesort')[::-1]` is invalid in case of duplicates in data.

hpat/hiframes/pd_series_ext.py
densmirn: We cannot merge them. The key idea of the code is to remove predefined series attributes from `SeriesAttribute`, not added to `SeriesAttribute` via previous loop. E.g. if to add `'resolve_nsmallest'` to `_not_series_array_attrs` and remove the "deleter" then `SeriesAttribute` will still contain attribute `'resolve_nsmallest'` and "new style" won't be used.

hpat/tests/test_series.py
densmirn: BTW I set up seed here `np.random.seed(100)`. So the result of the function should be always identical. Anyway I will think how to avoid random generation.

hpat/tests/test_series.py
densmirn: Let me keep Pythonic style and replace `throw` with `raise`:)

hpat/tests/test_series.py
densmirn: Multiplied number of data to test with duplicates.

hpat/datatypes/hpat_pandas_series_functions.py
shssf: do you think it will work if `self._data[i]==type_max`?

hpat/tests/test_series.py
shssf: In this case it might be more clear to use
```
data_duplicated = data * 3
```

hpat/datatypes/hpat_pandas_series_functions.py
densmirn: Yes, I checked it. Moreover the similar approach is used in Pandas.

PR #249, opened 2019-10-23
PyArrow porting 0.14.1 -> 0.15.0

PokhodenkoSA:
More info: https://gist.github.com/PokhodenkoSA/becf462bfc68791dd35b7e0a2691905b

### DIFF
--- a/buildscripts/parquet-reader-conda-recipe/meta.yaml
+++ b/buildscripts/parquet-reader-conda-recipe/meta.yaml
@@ -14,11 +14,11 @@ requirements:
     - {{ compiler('cxx') }}
     - cmake >=3.2
     - python 3.6.*
-    - pyarrow ==0.14.1
-    - arrow-cpp ==0.14.1
+    - pyarrow ==0.15.0
+    - arrow-cpp ==0.15.0
   run:
-    - pyarrow ==0.14.1
-    - arrow-cpp ==0.14.1
+    - pyarrow ==0.15.0
+    - arrow-cpp ==0.15.0
 
 
 about:
--- a/parquet_reader/hpat_parquet_reader.cpp
+++ b/parquet_reader/hpat_parquet_reader.cpp
@@ -86,10 +86,12 @@ int64_t pq_get_size_single_file(std::shared_ptr<FileReader> arrow_reader, int64_
 int64_t
     pq_read_single_file(std::shared_ptr<FileReader> arrow_reader, int64_t column_idx, uint8_t* out_data, int out_dtype)
 {
-    std::shared_ptr<::arrow::ChunkedArray> arr;
-    arrow_reader->ReadColumn(column_idx, &arr);
-    if (arr == NULL)
+    std::shared_ptr<::arrow::ChunkedArray> chunked_array;
+    arrow_reader->ReadColumn(column_idx, &chunked_array);
+    if (chunked_array == NULL)
         return 0;
+
+    auto arr = chunked_array->chunk(0);
 
     int64_t num_values = arr->length();
     // std::cout << "arr: " << arr->ToString() << std::endl;
--- a/parquet_reader/hpat_parquet_reader.cpp
+++ b/parquet_reader/hpat_parquet_reader.cpp
@@ -391,10 +391,11 @@ int64_t pq_read_string_single_file(std::shared_ptr<FileReader> arrow_reader,
 {
     // std::cout << "string read file" << '\n';
     //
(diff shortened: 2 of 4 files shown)


buildscripts/parquet-reader-conda-recipe/meta.yaml
```diff
@@ -14,11 +14,11 @@ requirements:
     - {{ compiler('cxx') }}
     - cmake >=3.2
     - python 3.6.*
-    - pyarrow ==0.14.1
-    - arrow-cpp ==0.14.1
+    - pyarrow ==0.15.0
```
shssf: Is it possible to have this version of the required package in one place for whole project? It might be some macro variable or something like this.

parquet_reader/hpat_parquet_reader.cpp
```diff
@@ -86,11 +86,13 @@ int64_t pq_get_size_single_file(std::shared_ptr<FileReader> arrow_reader, int64_
...
 {
-    std::shared_ptr<::arrow::Array> arr;
-    arrow_reader->ReadColumn(column_idx, &arr);
-    if (arr == NULL)
+    std::shared_ptr<::arrow::ChunkedArray> chunked_array;
+    arrow_reader->ReadColumn(column_idx, &chunked_array);
+    if (chunked_array == NULL)
         return 0;
 
+    auto arr = chunked_array->chunk(0);
```
shssf: Does it meant that we always have one chunk after reading?

buildscripts/parquet-reader-conda-recipe/meta.yaml
PokhodenkoSA: It is possible to use [Jinja2 template](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#templating-with-jinja).
Define in begin of the file:
```
{% set PYARROW_VERSION = ">=0.14.0,<0.15.0a0" %}
```
```suggestion
    - pyarrow {{ PYARROW_VERSION }}
```
Is it ok?

parquet_reader/hpat_parquet_reader.cpp
```diff
@@ -650,15 +655,18 @@ void pq_init_reader(const char* file_name, std::shared_ptr<FileReader>* a_reader
...
+    // std::vector<int> column_indices;
+    // column_indices.push_back(column_idx);
 
     std::shared_ptr<::arrow::Schema> col_schema;
-    auto descr = arrow_reader->parquet_reader()->metadata()->schema();
-    auto parquet_key_value_metadata = arrow_reader->parquet_reader()->metadata()->key_value_metadata();
-    parquet::arrow::FromParquetSchema(descr, column_indices, parquet_key_value_metadata, &col_schema);
+    // auto descr = arrow_reader->parquet_reader()->metadata()->schema();
+    // auto parquet_key_value_metadata = arrow_reader->parquet_reader()->metadata()->key_value_metadata();
+    // parquet::arrow::FromParquetSchema(descr, column_indices, parquet_key_value_metadata, &col_schema);
```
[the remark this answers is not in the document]
PokhodenkoSA: No. Will remove it.

parquet_reader/hpat_parquet_reader.cpp
PokhodenkoSA: It corresponds to [original implementation](https://github.com/apache/arrow/blob/9fb5c3d8f249482a5e3c0d33d7626c438e28c774/cpp/src/parquet/arrow/reader.cc#L775) of `ReadColumn(... <Array> ...). It gets `ChunkedArray` end returns only the first `Array` in it.
I think this code should be improved in the future, e.g. for using all arrays in ChunkedArray. But for porting reasons it is ok because it works as previously.

buildscripts/parquet-reader-conda-recipe/meta.yaml
PokhodenkoSA: Example with eliminating duplication inside one file https://github.com/IntelPython/hpat/pull/252.
I think use one version for whole project is not possible because all places plays different roles. Some places points range of available versions for use in customer environment, some points to concrete version we use for building and development.

@Vyacheslav-Smirnov have you some ideas about it?

PokhodenkoSA: This [article](https://caremad.io/posts/2013/07/setup-vs-requirement/) explains differences between versions in Python infrastructure files.

shssf: @PokhodenkoSA yes

PR #253, opened 2019-10-24
Add numba_pandas extension

PokhodenkoSA:
This PR shows how to create Numba extension for Pandas according the [documentation](http://numba.pydata.org/numba-doc/latest/extending/entrypoints.html).

See test `numba_pandas/tests/test_pandas.py`.

_Implicit_ loading of extension is supported in Numba >= 0.46.
Test file also shows how to load extension _explicitly_ for Numba < 0.46.

For HPAT version not ported to Numba 0.46 it is possible to run the test with Numba 0.46:
comment the line `import hpat.compiler` in `hpat/__init__.py`.

This test is not in CI yet.

`_init_extension()` imports only overloads used in the test. It should be extended for all overloads for Pandas.

I think we should consider moving Pandas overloads to `numba_pandas` package.

### DIFF
--- a/hpat/__init__.py
+++ b/hpat/__init__.py
@@ -35,3 +35,12 @@ multithread_mode = False
 
 __version__ = get_versions()['version']
 del get_versions
+
+
+def _init_extension():
+    '''Register Pandas classes and functions with Numba.
+
+    This exntry_point is called by Numba when it initializes.
+    '''
+    # Importing SDC is already happened
+    pass
--- a/setup.py
+++ b/setup.py
@@ -528,4 +528,10 @@ setup(name='hpat',
       install_requires=['numba'],
       extras_require={'HDF5': ["h5py"], 'Parquet': ["pyarrow"]},
       cmdclass=hpat_build_commands,
-      ext_modules=_ext_mods)
+      ext_modules=_ext_mods,
+      entry_points={
+          "numba_extensions": [
+              "init = hpat:_init_extension",
+          ],
+      },
+)


shssf:
I do not understand why you've done it.
Numba extension for Pandas called HPAT.
Try to create pandas.DataFrame instead pandas.Seires in the PR test
We need to implement HPAT usage without any HPAT decorators (Numba has to detect need of HPAT)

shssf:
Please discard my previous changes. I missed that there is no `hpat.jit` decorator.

numba_pandas/__init__.py
```diff
@@ -0,0 +1,11 @@
+
+def _init_extension():
```
shssf: As far as I understand (may be I'm wrong) this code should be called from Numba or should be existed in Numba.
I don't think we need additional `__init__` for HPAT. I think we already have initialization procedure.

numba_pandas/tests/test_pandas.py
```diff
@@ -0,0 +1,24 @@
+import numba
+import pandas
+
+from packaging import version
+
+# numba <0.46 requires explicit import extension
```
shssf: No need to check this. We will work with latest version. No legacy support is required.

numba_pandas/tests/test_pandas.py
```diff
@@ -0,0 +1,24 @@
...
+if version.parse(numba.__version__) < version.parse('0.46'):
+    from numba_pandas import _init_extension
+    _init_extension()
+
+
+def test_function():
+    pandas_func = pandas.Series
+
+    @numba.njit
```
shssf: I don't think we need additional test suite for this functionality. It will be the only functionality we have.
Could you please chenge several `hpat.jit` dicorators in our regular test system? For example, in `test_series.py` with several tests which are already support `new style` functionality?

numba_pandas/__init__.py
PokhodenkoSA: 1. I will explain how Numba calls this code. No magic.
In `setup.py` I have added an extension point:
```
      entry_points={
          "numba_extensions": [
              "init = numba_pandas:_init_extension",
          ],
      },
```
This extension point sets `init` to exactly this function.
Numba >=0.46 can find this extension point and call the function.
This is a standard Python packaging technique when one package access to plugins provided by other packages. Each package can publish extension points which are visible to other packages.
Numba states that this technique is supported in 0.46 - it is described in Numba documentation - http://numba.pydata.org/numba-doc/latest/extending/entrypoints.html.
2. I agree, initialization part should be revised. Will make issue and PR for it.
Quick thoughts:
Dedicated function `_init_extension()` makes it possible to control which parts of HPAT will be loaded for extending Numba. With `import hpat` the whole HPAT will be loaded.

numba_pandas/tests/test_pandas.py
PokhodenkoSA: It is just for demonstrating how to test with previous versions of Numba <0.46.
I will remove because now porting to Numba 0.46 is merged to master.

numba_pandas/__init__.py
shssf: @PokhodenkoSA 
2. Thank you
1. Does it mean, the SDC (old HPAT) will initialized every time when Numba works? I thought it happen if Pandas package required only.

shssf:
@PokhodenkoSA Do we really need `_init_extension` in some new file? I think we can implement it in any existed `init` file

numba_pandas/__init__.py
PokhodenkoSA: 1. SDC will be initializes every time when Numba is imported. We can control the part of SDC which will be loaded.
> I thought it happen if Pandas package required only.
Could you explain "required"? Does it mean "used in code - import pandas" or "Pandas is installed"?
We could check if Pandas is installed but Pandas is in requirements of SDC. Actually we could remove Pandas from requirements of SDC and load SDC parts depending on whether Pandas is installed.

numba_pandas/tests/test_pandas.py
PokhodenkoSA: I will create separate PR.

numba_pandas/__init__.py
PokhodenkoSA: Can I keep this function `_init_extension()` in `numba_pandas/__init__.py`? Or should I move it to `hpat` (`sdc`) folder?

shssf: I not sure but I would propose to move it into `hpat/__init__.py`

shssf:
@PokhodenkoSA I would like to merge this PR but we need to move initializer into `hpat/__init__` and remove other changes to related to `setup.py`

numba_pandas/tests/test_pandas.py
PokhodenkoSA: #256

PR #259, opened 2019-10-28
Skiping hanging tests

AlexanderKalistratov:
Skip tests which starts to hang on Windows after rebase on Numba 0.46


shssf:
Do you think it is better to implement some common procedure to enable/disable them?
It would be helpful to have a simple switch to control these tests.

AlexanderKalistratov:
@shssf I hope it is a temporally solution and I'll turn these tests back this week. In this case creating a switch is not that useful

AlexanderKalistratov:
@densmirn and @Hardcode84  could you please review and if you are fine - merge this MR since it blocking others?

shssf:
I don't see a reason to merge this since it is temporal solution for this week. If you need this for your development - use commit cherry picking

PR #261, opened 2019-10-29
Fix pd.Series overload check for None index 

kozlov-alexey:


hpat/hiframes/pd_series_ext.py
```diff
@@ -1270,23 +1270,18 @@ def typer(val1, val2):
...
-                hpat.hiframes.api.fix_df_array(index),
-                name)
-
-        return hpat_pandas_series_index_ctor_impl
+    is_index_none = True if (isinstance(index, types.NoneType) or index is None) else False
 
     def hpat_pandas_series_ctor_impl(data=None, index=None, dtype=None, name=None, copy=False, fastpath=False):
-        return hpat.hiframes.api.init_series(hpat.hiframes.api.fix_df_array(data), index, name)
+
+        # use binop here as otherwise Numba's dead branch pruning doesn't work
```
shssf: Please form it as regular doc string

AlexanderKalistratov: Docstrings are for auto generation of documentation. To provide description of classes/functions. Not for code explaining. But comments for code explaining

kozlov-alexey: @shssf Sorry, I don't fully understand what change you are requesting. Should I put a comment with TODO about one specific line of code in the function doc string? Or do you want to add general doc string for this function, not related to this comment?

PokhodenkoSA:
@kozlov-alexey could, please, you describe the problem in description of PR

hpat/hiframes/pd_series_ext.py
```diff
@@ -1270,23 +1270,18 @@ def typer(val1, val2):
...
-                name=None,
-                copy=False,
-                fastpath=False):
-            return hpat.hiframes.api.init_series(
-                hpat.hiframes.api.fix_df_array(data),
-                hpat.hiframes.api.fix_df_array(index),
-                name)
-
-        return hpat_pandas_series_index_ctor_impl
+    is_index_none = True if (isinstance(index, types.NoneType) or index is None) else False
```
densmirn: Will the following line work the same?
`is_index_none = isinstance(index, types.NoneType) or index is None`

hpat/hiframes/pd_series_ext.py
shssf: @kozlov-alexey I just thought that it is better to have
```
'''
comment
'''
```
instead
`# comment`

hpat/hiframes/pd_series_ext.py
kozlov-alexey: @densmirn Well, yes, I tend to avoid using python or this way though. But since we have Booleans exprs here anyway, I think I'll use shorter version. Thanks!

PR #265, opened 2019-10-31
Fix getitem and loc/iloc/at/iat (make a new style work)

1e-to:

### DIFF
--- a/hpat/datatypes/hpat_pandas_series_functions.py
+++ b/hpat/datatypes/hpat_pandas_series_functions.py
@@ -63,7 +63,6 @@ def hpat_pandas_series_getitem(self, idx):
     :class:`pandas.Series` or an element of the underneath type
             object of :class:`pandas.Series`
     """
-
     _func_name = 'Operator getitem().'
 
     if not isinstance(self, SeriesType):
@@ -74,7 +73,6 @@ def hpat_pandas_series_getitem(self, idx):
             """
             **Test**: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_iloc1
             """
-
             result = self._data[idx]
             return result
 
@@ -85,7 +83,6 @@ def hpat_pandas_series_getitem(self, idx):
             """
             **Test**: python -m hpat.runtests hpat.tests.test_series.TestSeries.test_series_iloc2
             """
-
             result = pandas.Series(self._data[idx])
             return result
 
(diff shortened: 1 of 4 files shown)


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -107,27 +100,61 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+    _func_name = 'Operator at/iat().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    if not isinstance(idx, types.Integer):
+        raise TypingError('{} The index must be an Integer. Given: {}'.format(_func_name, idx))
+
+    def hpat_pandas_series_iat_impl(self):
+        return self
```
shssf: Why you implement separate method if the realization (`return self`) is the same?

hpat/hiframes/hiframes_typed.py
```diff
@@ -177,8 +177,8 @@ def _run_assign(self, assign):
             self._type_changed_vars.append(lhs)
 
         if isinstance(rhs, ir.Expr):
-            if rhs.op == 'getattr':
-                return self._run_getattr(assign, rhs)
+            # if rhs.op == 'getattr':
```
shssf: As far as I understand this handles diffrent input types. Not only Series. If you like to remove it for Series you need to implement type check here (`not isinstance(rhs.something, SeriesType`)

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: I want to rewrite getitem() so that I can get the item by index, not just by id, to implement methods at/iat/loc/iloc like in pandas

shssf:
@1e-to Could you please resolve code conflicts?

1e-to:
> @1e-to Could you please resolve code conflicts?

Work is still in progress, no solution has been found, I can resolve conflicts but it’s too early to merge into the master

shssf:
@1e-to I think good solution will be in `SeriesType::series_access_type` variable implementation (like `SeriesType::is_named`). This variable gives a necessary information for `getiterm()`. Also, we do not have to implement these complicated datatypes like `SeiesTypeLoc`, `SeiesTypeAt` and etc

PR #269, opened 2019-11-02
Add config for PEP8Speak

Vyacheslav-Smirnov:


.pep8speaks.yml
```diff
@@ -0,0 +1,14 @@
...
+scanner:
+    diff_only: True  # If False, the entire file touched by the Pull Request is scanned for errors. If True, only the diff is scanned.
+    linter: pycodestyle  # Other option is flake8
+ 
+pycodestyle:  # Same as scanner.linter value. Other option is flake8
+    max-line-length: 120  # Default is 79 in PEP 8
+    ignore:  # Errors and warnings to ignore
+        - W504  # line break after binary operator
+        - E402  # module level import not at top of file
+        - E731  # do not assign a lambda expression, use a def
```
shssf: Why do we need these exceptions?
I would vote to leave check procedure with default errors and warnings

.pep8speaks.yml
```diff
@@ -0,0 +1,14 @@
...
+ 
+no_blank_comment: True  # If True, no comment is made on PR without any errors.
```
shssf: insert end line

.pep8speaks.yml
Vyacheslav-Smirnov: Ok, will remove them

PR #279, opened 2019-11-07
fix for argsort/sort_values indices check

1e-to:


hpat/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2730,7 +2730,7 @@ def hpat_pandas_series_argsort_noidx_impl(self, axis=0, kind='quicksort', order=
         na = self.isna().sum()
         result = numpy.empty(len(self._data), dtype=numpy.int64)
         na_data_arr = hpat.hiframes.api.get_nan_mask(self._data)
-        sort_nona = numpy.argsort(self._data[~na_data_arr])
+        sort_nona = numpy.argsort(self._data[~na_data_arr], kind='mergesort')
```
densmirn: I think it's not correctly to use `'mergesort'` when input `kind='quicksort'`.

hpat/tests/test_series.py
```diff
@@ -3010,6 +3018,9 @@ def test_series_sort_values_full_unicode4(self):
         def test_impl(series, ascending):
             return series.sort_values(axis=0, ascending=ascending, inplace=False, kind='quicksort', na_position='last')
 
+        def pandas_impl(series, ascending):
+            return series.sort_values(axis=0, ascending=ascending, inplace=False, kind='mergesort', na_position='last')
```
densmirn: Why not pass kind of sorting to the implementation as argument to have a single implementation?

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: We cannot use 'quicksort' sorting because then we will not compare indices

densmirn: Let's imagine we already support parameter `kind`. So are you still going to sort data via `'mergesort'`?

hpat/tests/test_series.py
1e-to: Fixed

hpat/datatypes/hpat_pandas_series_functions.py
1e-to: When other sortings are implemented, it will be possible to change this parameter, but now it’s more efficient to leave it and not spend much time on it

1e-to: Maybe I should write that now only mergesort is supported

densmirn: Anyway I vote for unification of parameter `kind` for `pd.argsort` and `np.argsort` to be able to expand the functionality simpler in further, in particular to add support of parameter `kind`.

hpat/tests/test_series.py
```diff
@@ -3035,8 +3034,8 @@ def test_impl(series, ascending):
             for index in [gen_srand_array(len(data)), gen_frand_array(len(data)), range(len(data))]:
                 for ascending in [True, False]:
                     series = pd.Series(data, index)
-                    ref_result = test_impl(series, ascending)
-                    jit_result = hpat_func(series, ascending)
+                    ref_result = test_impl(series, ascending, kind='mergesort')
+                    jit_result = hpat_func(series, ascending, kind='quicksort')
```
densmirn: That looks fundamentally wrong because potentially we are able to get different results and then compare these ones. Now it works fine for current input data, but in theory some input can break it.

PR #280, opened 2019-11-07
Tests for getitem + loc/iloc/at/iat

1e-to:


hpat/tests/test_series.py
```diff
@@ -3627,6 +3627,132 @@ def test_impl(s, axis):
...
+            return S.loc[2:5]
+
+        jit_impl = hpat.jit(test_impl)
+
+        S = pd.Series([11, 22, 33], [2, 3, 5])
+        self.assertEqual(jit_impl(S), test_impl(S))
+
+    def test_series_loc_slice_str(self):
+        def test_impl(S):
+            return S.loc['2':'5']
```
shssf: @densmirn Could you please help to combine similar tests into one with loop over diffrent input data?

hpat/tests/test_series.py
```diff
@@ -3627,6 +3627,132 @@ def test_impl(s, axis):
...
+
+        jit_impl = hpat.jit(test_impl)
+
+        S = pd.Series([11, 22, 33], ['2', '3', '5'])
+        self.assertEqual(jit_impl(S), test_impl(S))
```
shssf: Please note, it is better to use `series_equal` or `vector_equal` instead `assetEqual` if more than one output value is expected

hpat/tests/test_series.py
densmirn: @shssf the similar tests were combined, but the tests are failing as before. So @1e-to could you fix all issues to get all tests passed?

hpat/tests/test_series.py
```diff
@@ -3627,6 +3627,75 @@ def test_impl(s, axis):
...
+
+        jit_impl = hpat.jit(test_impl)
+
+        starts = [0, 0, '2']
+        ends = [2, 2, '5']
+        indices = [[2, 3, 5], ['2', '3', '5'], ['2', '3', '5']]
+        for start, end, index in zip(starts, ends, indices):
+            S = pd.Series([11, 22, 33], index)
+            ref_result = test_impl(S, start, end)
+            print(type(ref_result))
```
densmirn: I forgot to remove the debug print. @1e-to could you please remove it with the next patch?

PR #282, opened 2019-11-11
Add tests with numba.jit for Series

### LINKED ISSUE #256: Change several `hpat.jit` to `numba.njit` decorators in regular test system
Could you please chenge several `hpat.jit` to `numba.njit` (please use njit with default paralelization)  dicorators in our regular test system? For example, in `test_series.py` with several tests which are already support `new style` functionality?

_Originally posted by @shssf in https://github.com/IntelPython/hpat/pull/253_

PokhodenkoSA:
Fixes #256

For new tests for Series you should:
1. use `self.jit(...)` instead of `hpat.jit(...)`
2. Add test name to `hpat.tests.test_series._numba_not_supported_tests` only if it does not support numba.jit

Also, old Series tests should be revised for reasons why they do not support numba.jit. See `_numba_not_supported_tests` list.

### DIFF
--- a/hpat/tests/test_series.py
+++ b/hpat/tests/test_series.py
@@ -4107,5 +4107,140 @@ class TestSeries(unittest.TestCase):
         self.assertIn(msg, str(raises.exception))
 
 
+# Sort Lines Ascending
+_numba_not_supported_tests = [
+    "test_create_str",
+    "test_create1",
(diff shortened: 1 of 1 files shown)


PokhodenkoSA:
I will add `numba.jit(nopython=True, parallel=True)`.

PR #305, opened 2019-11-19
Add support for Series.operator.add in a new-style

kozlov-alexey:


sdc/datatypes/common_functions.py
```diff
@@ -131,3 +132,19 @@ def _append_list_string_array_impl(A, B):
                 return new_data
 
             return _append_list_string_array_impl
+
+
+@numba.njit
+def _hpat_ensure_array_capacity(new_size, arr):
+    '''Function creating a copy of numpy array with a size more than specified'''
+    # TODO: replace this function with np.resize when supported by Numba
```
densmirn: `np.resize` resizes array in other way than this algorithm. The algorithm can generates array of size more than `new_size` and doesn't repeat values of `arr` in the resulting array. Looks like there is mismatch somewhere :)

sdc/datatypes/common_functions.py
```diff
@@ -131,3 +132,19 @@ def _append_list_string_array_impl(A, B):
                 return new_data
 
             return _append_list_string_array_impl
+
+
+@numba.njit
+def _hpat_ensure_array_capacity(new_size, arr):
+    '''Function creating a copy of numpy array with a size more than specified'''
```
densmirn: Could you apply below recommendation everywhere in this PR?
[pep8 recommendation](https://www.python.org/dev/peps/pep-0008/#string-quotes): For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in PEP 257.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,343 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+    Returns
+    -------
+    :obj:`pandas.Series`
+        The result of the operation
+    """
+
+    _func_name = 'Operator add().'
+
+    if not isinstance(self, SeriesType):
+        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
```
densmirn: Could you use `TypeChecker` where it's possible?

sdc/datatypes/common_functions.py
kozlov-alexey: Removed the comment at all)

sdc/datatypes/common_functions.py
kozlov-alexey: Applied

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Done

kozlov-alexey:
@AlexanderKalistratov @Hardcode84 Can you please review too?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+
+    # specializations for numeric series - TODO: support arithmetic operation on StringArrays
+    if (isinstance(other, types.Number)):
+        def hpat_pandas_series_add_scalar_impl(self, other):
+            return pandas.Series(self._data + other, self._index)
+
+        return hpat_pandas_series_add_scalar_impl
+
+    elif (isinstance(other, SeriesType)):
+        is_numeric_index = isinstance(self.index, (types.Array, types.NoneType))
```
AlexanderKalistratov: Shouldn't it be like: `(isinstance(self.index, types.NoneType) or isinstance(self.index, types.Array) and isinstance(self.index.dtype, types.Number)`. The same as check above? Or Check above should be the same as this one?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+
+        if is_numeric_index:
+            ty_left_index_dtype = types.int64 if isinstance(self.index, types.NoneType) else self.index.dtype
```
AlexanderKalistratov: Looks like you can move this block to line 3958 (after line `if is_numeric_index == True:  # noqa`

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+            def hpat_pandas_series_add_impl(self, other):
+
+                if (len(self._data) == len(other._data)):
+                    return pandas.Series(numpy.asarray(self._data + other._data, numpy.float64))
+                else:
+                    min_data_size = min(len(self._data), len(other._data))
+                    max_data_size = max(len(self._data), len(other._data))
+                    new_data = numpy.empty(max_data_size, dtype=numpy.float64)
+                    new_data[:min_data_size] = self._data[:min_data_size] + other._data[:min_data_size]
+                    new_data[min_data_size:] = numpy.repeat(numpy.nan, max_data_size - min_data_size)
```
AlexanderKalistratov: This behavior could be changed by parameters passed to `add` method of series.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.add.html#pandas.Series.add

This PR is about '+' operator, but what about 'add' method?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+
+                    return pandas.Series(new_data, self._index)
+        else:
+            def hpat_pandas_series_add_impl(self, other):
+                left_index, right_index = self.index, other.index
+
+                # check if indexes are equal and series don't have to be aligned
+                if is_numeric_index == True:  # noqa
+                    if (numpy.array_equal(left_index, right_index)):
+                        return pandas.Series(numpy.asarray(self._data + other._data, numpy.float64),
```
AlexanderKalistratov: Are you sure about `numpy.float64`? shouldn't is be something like `common_dtype`?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+                    # TODO: replace with StringArrays comparison
+                    is_index_equal = (len(self._index) == len(other._index)
+                                      and num_total_chars(self._index) == num_total_chars(other._index))
+                    for i in numpy.arange(len(self._index)):
+                        if (self._index[i] != other._index[i]
+                                or str_arr_is_na(self._index, i) is not str_arr_is_na(other._index, i)):
+                            is_index_equal = False
+
+                    if is_index_equal:
+                        return pandas.Series(numpy.asarray(self._data + other._data, numpy.float64),
```
AlexanderKalistratov: Are you sure about `numpy.float64`? shouldn't is be something like `common_dtype`?

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: Alright, most likely you can't. Since line 3958 inside jitted region

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+    ty_checker = TypeChecker('Operator add().')
+    ty_checker.check(self, SeriesType)
+
+    if not isinstance(other, (SeriesType, types.Number)):
+        ty_checker.raise_exc(other, 'pandas.series or scalar', 'other')
+
+    series_indexes_alignable = False
+    if isinstance(other, SeriesType):
+        if (other.index == string_array_type and self.index == string_array_type):
+            series_indexes_alignable = True
```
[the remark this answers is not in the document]
kozlov-alexey: Well, pandas will use dtype that can hold objects of diff types, i.e. resulting series will have index with dtype=object and all elements of resulting series will be NaN, since indexes won't align.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+
+        if ((isinstance(self.index, types.NoneType) or
+             isinstance(self.index, types.Array) and isinstance(self.index.dtype, types.Number))
```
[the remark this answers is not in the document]
kozlov-alexey: Done

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+            and (isinstance(other.index, types.NoneType) or
+                 isinstance(other.index, types.Array) and isinstance(other.index.dtype, types.Number))):
```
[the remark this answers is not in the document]
kozlov-alexey: Added

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Given that the above check passed, we can only check first series here, that's why it worked. But I agree that it's better to reuse the check from above, so I'll change to using 'none_or_numeric_indexes' variable in the next commit.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Well, yes, but this block can be moved to the else branch below, where it's probably a better place for it. Will do in the next commit.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: @AlexanderKalistratov Yes, I understand, I actually planned to have two different templates: one for binary operators and the other for add/sub/div/... methods, which can then be used with auto-generation script, because latter need special handling for method args and so on. But yes, I will address that in subsequent PRs.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+
+@overload(hpat_join_series_indexes)
+def hpat_join_series_indexes_overload(left, right):
+    '''Function for joining arrays left and right in a way similar to pandas.join 'outer' algorithm'''
+
+    if (isinstance(left, types.Array) and isinstance(right, types.Array)):
+        np_dtypes = [numpy_support.as_dtype(left.dtype), numpy_support.as_dtype(right.dtype)]
+        np_common_dtype = numpy.find_common_type([], np_dtypes)
+        numba_common_dtype = numpy_support.from_dtype(np_common_dtype)
+        if (isinstance(left.dtype, types.Number) and isinstance(right.dtype, types.Number)):
```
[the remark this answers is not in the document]
kozlov-alexey: Fixed

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3646,3 +3649,344 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+                elif ridx[i] != -1:
+                    joined[i] = right[ridx[i]]
+                    if (str_arr_is_na(right, ridx[i])):
+                        str_arr_set_na(joined, i)
+                else:
+                    str_arr_set_na(joined, i)
+
+            return joined, lidx, ridx
+
+        return hpat_join_series_indexes_impl
```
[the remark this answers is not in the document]
kozlov-alexey: Yeap, done.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Yes, thank you, that's incorrect of course.

AlexanderKalistratov: @kozlov-alexey I'm already not that sure it is incorrect. In case we are adding `nan` dtype should become numpy.float64 anyway

sdc/tests/test_series.py
```diff
@@ -4392,6 +4615,7 @@ def test_series_pct_change_impl(S, periods=1, fill_method='pad', limit=None, fre
     "test_series_op4",
     "test_series_op6",
     "test_series_op7",
+    "test_series_operator_add_numeric_same_index_default",
```
PokhodenkoSA: This code is not in master after #340.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Right, just verified that the result of an operation in Pandas has unstable dtype: if indexes match it will be common numpy dtype of two series, but if one series has indexes absent in other, then np.nan will appear in data and resulting dtype will be float. So we have to use types.float64 due to type stability requirements.

sdc/tests/test_series.py
kozlov-alexey: Yes, thank you, I'll use new skip decorators.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3795,3 +3798,132 @@ def hpat_pandas_series_pct_change_impl(self, periods=1, fill_method='pad', limit
...
+                if (len(self._data) == len(other._data)):
+                    return pandas.Series(numpy.asarray(self._data + other._data, numpy.float64))
+                else:
+                    min_data_size = min(len(self._data), len(other._data))
+                    max_data_size = max(len(self._data), len(other._data))
+                    new_data = numpy.empty(max_data_size, dtype=numpy.float64)
+                    new_data[:min_data_size] = self._data[:min_data_size] + other._data[:min_data_size]
+                    new_data[min_data_size:] = numpy.repeat(numpy.nan, max_data_size - min_data_size)
+
+                    return pandas.Series(new_data, self._index)
```
AlexanderKalistratov: @kozlov-alexey In some cases resulting series should have the same name as original one. This cases, seems to be are:
* series + scalar (i.e. A + 1)
* series + series, but both series has the same name

This is very important in order to implement DataFrame filters. I'm ok to do it in separate PR, but this must be done. Also tests for named series should be added

AlexanderKalistratov: Well, after some more investigations, it's not important for filters, but still we should be consistent with pandas

AlexanderKalistratov: It looks like, at the moment we can't support second case (when two series have the same name) since it is runtime condition, but whether series has name or not we should know at compile time.

kozlov-alexey: @AlexanderKalistratov Thank you! I think we should address this in a separate PR. BTW, I just checked if there's any difference between '1 + A' and 'A + 1' and there's no, in both cases resulting series takes name from A. I'm going to dig in pandas sources to clarify this behavior.

PR #314, opened 2019-11-21
performance tests for series.min/max

Rubtsowa:


sdc/tests/tests_perf/__init__.py
```diff
@@ -1,2 +1,3 @@
 from sdc.tests.tests_perf.test_perf_unicode import *
 from sdc.tests.tests_perf.test_perf_series_str import *
+from sdc.tests.tests_perf.test_perf_series import *
```
PokhodenkoSA: I am not happy with importing via *. It always make me confusing when I try to decide where a thing was imported from.

sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -0,0 +1,79 @@
...
+# OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+# OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+import pandas as pd
+
+from sdc.tests.test_utils import *
+from sdc.tests.tests_perf.test_perf_base import TestBase
+from sdc.tests.tests_perf.test_perf_utils import *
```
PokhodenkoSA: Why not import particular things or use it like this:
```python
from sdc.tests import test_utils
...
... test_utils.some_function(...)
```

densmirn: @PokhodenkoSA I agree it's more clear way to see where a function called from. It could be done in separate PR.

sdc/tests/tests_perf/__init__.py
densmirn: @PokhodenkoSA do you propose to list all tests instead of asterisk?

PR #336, opened 2019-11-26
fix Series.count

Rubtsowa:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2954,7 +2954,8 @@ def hpat_pandas_series_count(self, level=None):
     if isinstance(self.data, StringArrayType):
         def hpat_pandas_series_count_str_impl(self, level=None):
 
-            return len(self._data)
+            isna = self.isna()
```
densmirn: Maybe name it something like `nan_mask`?

PR #337, opened 2019-11-26
Adding performance test for Series (A+B).sum()

kozlov-alexey:
With the following test two observations were made (both require further investigation):
1. Changing NUMBA_NUM_THREADS env variable (TBB and default threading layer) has no impact on the execution times both in new-style and old-style implementations,
2. Old-style implementation is 1.5-2 times faster than the new-style impl.

### DIFF
--- a/sdc/tests/tests_perf/test_perf_series.py
+++ b/sdc/tests/tests_perf/test_perf_series.py
@@ -25,6 +25,7 @@
 # EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *****************************************************************************
 import pandas as pd
+import numpy as np
 
 from sdc.tests.test_utils import *
 from sdc.tests.tests_perf.test_perf_base import *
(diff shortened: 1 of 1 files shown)


sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -265,6 +276,28 @@ def _test_series(self, pyfunc, name, input_data=None):
...
+            # TODO: replace with generic function to generate random sequence of floats
+            data1 = np.random.ranf(data_length)
+            data2 = np.random.ranf(data_length)
+            A = pd.Series(data1)
+            B = pd.Series(data2)
+
+            compile_results = calc_compilation(pyfunc, A, B, iter_number=self.iter_number)
+
+            # Warming up
+            hpat_func(A, B)
```
kozlov-alexey: This uses sdc.jit, but it'd be better to use do_jit function instead, as warming should use the same decorator as subsequent get_times.

PR #339, opened 2019-11-26
Fix series perf tests in new style

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3104,10 +3104,10 @@ def hpat_pandas_series_argsort_idx_impl(self, axis=0, kind='quicksort', order=No
                 sort_nona = numpy.argsort(self._data[~na_data_arr])
             q = 0
             for id, i in enumerate(sort):
-                if id not in list(sort[len(self._data) - na:]):
-                    result[id] = sort_nona[id-q]
-                else:
+                if id in list(sort[len(self._data) - na:]):
```
densmirn: Not the best name `sort` for the Python variable. Why do you use constructor `list()` here? Is it exactly required?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3104,10 +3104,10 @@ def hpat_pandas_series_argsort_idx_impl(self, axis=0, kind='quicksort', order=No
...
                     q += 1
+                else:
+                    result[id] = sort_nona[id-q]
```
densmirn: Please follow [PEP8 recommendation](https://www.python.org/dev/peps/pep-0008/#other-recommendations): "Always surround these binary operators with a single space on either side".

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3131,10 +3131,10 @@ def hpat_pandas_series_argsort_noidx_impl(self, axis=0, kind='quicksort', order=
             sort_nona = numpy.argsort(self._data[~na_data_arr])
         q = 0
         for id, i in enumerate(sort):
-            if id not in list(sort[len(self._data) - na:]):
-                result[id] = sort_nona[id - q]
-            else:
+            if id in list(sort[len(self._data) - na:]):
                 q += 1
+            else:
+                result[id] = sort_nona[id - q]
```
densmirn: The same comments as above.

PR #344, opened 2019-11-27
Merge numba master

Hardcode84:
based on numba `ceab10620764d1ff29e0c0e550d1d5975674819b`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,21 @@
 
 dist: trusty
 
+branches:
+  only:
+  - master
+
 env:
     global:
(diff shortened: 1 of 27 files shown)


Vyacheslav-Smirnov:
Hi @Hardcode84 @AlexanderKalistratov  ,
Unfortunately, currently CI cannot test changes for Numba master. As we decided earlier, I will create internal testing with result publishing in PR for numba-master branch.
That is why if locally your changes works fine, I suggest to merge this PR.
Later we can fix it if something is wrong in next PR.

PR #345, opened 2019-11-28
Add functions for dataframe: median, mean, min, max, sum

Rubtsowa:


sdc/config.py
```diff
@@ -91,3 +91,8 @@
 '''
 Default value for a pointer intended to use as Numba.DefaultPassBuilder.define_nopython_pipeline() in overloaded function
 '''
+
+use_default_dataframe = distutils_util.strtobool(os.getenv('SDC_CONFIG_USE_DEFAULT_DATAFRAME', 'True'))
+'''
+Default value used to select compiler pipeline in a function decorator
```
kozlov-alexey: Description is not right - should be something like "Config variable used to select DataFrameType model (default is legacy model)"

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,340 @@
...
+                item_name = dataframe_item._name
+                result_data.append(item_count)
+                result_index.append(item_name)
+
+            return pandas.Series(data=result_data, index=result_index)
+
+        return sdc_pandas_dataframe_count_impl
+
+else:
+    def reduce(df, name):
```
kozlov-alexey: I think function name should be more specific, e.g. "sdc_pandas_dataframe_reduce_columns"

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,340 @@
...
+        saved_columns = df.columns
+        n_cols = len(saved_columns)
+        data_args = tuple('data{}'.format(i) for i in range(n_cols))
+        func_text = "def _reduce_impl(df, axis=None, skipna=None, level=None, numeric_only=None):\n"
```
kozlov-alexey: This line defines function signature, which is now actually hardcoded to signature that 'median' has (the other function like max, min, mean also have the same signature, but not all). That won't work for other functions, e.g. Series.sum, which has additional arguments:
   Series.sum(self, axis=None, skipna=None, level=None, numeric_only=None, **min_count=0,** **kwargs)
You have to pass additional arguments to this function to be able to generate correct function text that matches the signature used in overload.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,340 @@
...
+           :obj:`pandas.Series` or `pandas.DataFrame`
+                   return the median of the values for the requested axis.
+           """
+
+        name = 'median'
+
+        if not isinstance(df, DataFrameType):
+            raise TypingError('{} The object must be a pandas.dataframe. Given: {}'.format(name, df))
+
+        if not (isinstance(axis, types.Omitted) or axis is None):
```
kozlov-alexey: Probably we should use ty_checker here and everywhere else, e.g.
```suggestion
    if (not isinstance(axis, (int, types.Integer, str, types.UnicodeType, types.StringLiteral, types.Omitted))
        and axis not in (0, 'index')):
        ty_checker.raise_exc(axis, 'integer or string', 'axis')
```

sdc/tests/test_dataframe.py
```diff
@@ -1121,6 +1121,29 @@ def test_impl(n):
         hpat_func = sdc.jit(test_impl)
         pd.testing.assert_series_equal(hpat_func(n), test_impl(n))
 
+    def test_dataframe_min(self):
+        def test_impl(df):
+            return df.min()
+        sdc_func = sdc.jit(test_impl)
+        df = pd.DataFrame({"A": [12, 4, 5, 44, 1],
```
kozlov-alexey: Since we just use Series method for each column, I prefer to use an aggregated DF with many columns that actually differ between each other, that is, each testing some specific case - ints, floats, datatimes, etc.

sdc/datatypes/hpat_pandas_dataframe_functions.py
AlexanderKalistratov: All functions seems to have more or less the same checks. Shouldn't we move generic checks into generic function in order to avoid copy and paste?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,340 @@
...
+            func_text += "  {} = {}.{}()\n".format(d + '_O', d + '_S', name)
+        func_text += "  data = np.array(({},))\n".format(
+            ", ".join(d + '_O' for d in data_args))
+        func_text += "  index = hpat.str_arr_ext.StringArray(({},))\n".format(
+            ", ".join("'{}'".format(c) for c in saved_columns))
+        func_text += "  return hpat.hiframes.api.init_series(data, index)\n"
+        loc_vars = {}
+
+        print()
+        print(func_text)
```
AlexanderKalistratov: I believe it should be removed

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,340 @@
...
+        func_text = "def _reduce_impl(df, axis=None, skipna=None, level=None, numeric_only=None):\n"
+        for i, d in enumerate(data_args):
+            func_text += "  {} = hpat.hiframes.api.init_series(hpat.hiframes.pd_dataframe_ext.get_dataframe_data(df, {}))\n".format(
+                d + '_S', i)
+            func_text += "  {} = {}.{}()\n".format(d + '_O', d + '_S', name)
+        func_text += "  data = np.array(({},))\n".format(
+            ", ".join(d + '_O' for d in data_args))
+        func_text += "  index = hpat.str_arr_ext.StringArray(({},))\n".format(
+            ", ".join("'{}'".format(c) for c in saved_columns))
+        func_text += "  return hpat.hiframes.api.init_series(data, index)\n"
```
densmirn: You concatenate strings via operator `+`, but more effective way is to create `list`, fill the list via `.append()` (do NOT forget to cut off `\n`), then join the list to the string via `'\n'.join()`, e.g.:
```
func_lines = []
func_lines.append('the first line')
func_lines.append('the second line')
func_lines.append('the last line')
```

densmirn:
> Hello @Rubtsowa! Thanks for opening this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`sdc/datatypes/hpat_pandas_dataframe_functions.py`](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py):
> 
> > [Line 47:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L47): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 70:121](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L70): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (122 > 120 characters)
> > [Line 108:121](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L108): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (132 > 120 characters)
> > [Line 128:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L128): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 176:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L176): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 224:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L224): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 272:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L272): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 320:5](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/datatypes/hpat_pandas_dataframe_functions.py#L320): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> 
> * In the file [`sdc/hiframes/pd_dataframe_ext.py`](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/hiframes/pd_dataframe_ext.py):
> 
> > [Line 1641:1](https://github.com/IntelPython/sdc/blob/063e5ea323f1883534ab8ee7f0ad3f895f95618b/sdc/hiframes/pd_dataframe_ext.py#L1641): [E402](https://duckduckgo.com/?q=pep8%20E402) module level import not at top of file

Please fix the code style issues.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,473 @@
...
+        saved_columns = df.columns
+        n_cols = len(saved_columns)
+        data_args = tuple('data{}'.format(i) for i in range(n_cols))
+        help_param = ', {}={}):'
+        func_text = 'def _reduce_impl(df):'
+        for key, value in param:
+            func_text = func_text[:-2]
+            func_text = func_text + help_param
+            func_text = func_text.format(key, value)
+        func_lines = [func_text]
```
densmirn: Let's refactor this part:
```
all_params = ['df']
for name, value in param:
    all_params.append('{}={}'.format(name, value))
func_definition =  'def _reduce_impl({}):'.format(', '.join(all_params))
func_lines = [func_definition]
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,473 @@
...
+                item_name = dataframe_item._name
+                result_data.append(item_count)
+                result_index.append(item_name)
+
+            return pandas.Series(data=result_data, index=result_index)
+
+        return sdc_pandas_dataframe_count_impl
+
+else:
+    def sdc_pandas_dataframe_reduce_columns(df, name, param):
```
densmirn: Maybe replace `param` with `params`.

sdc/datatypes/hpat_pandas_dataframe_functions.py
Rubtsowa: I don't know why, but this variant not work. An error occurs. Parametrs become Series.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1994,8 +1994,7 @@ def hpat_pandas_series_sum(
     .. only:: developer
 
         Tests:
-            python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_sum1
-            # python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_sum2
+            python -m sdc.runtests -k sdc.tests.test_series.TestSeries.test_series_sum
```
kozlov-alexey: Need to refer to all tests for sum method:
python -m sdc.runtests -k sdc.tests.test_series.TestSeries.test_series_sum*

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2651,18 +2650,21 @@ def hpat_pandas_series_prod(self, axis=None, skipna=True, level=None, numeric_on
...
+            or not (isinstance(numeric_only, (types.Omitted, types.NoneType)) or numeric_only is None) \
+            or not (isinstance(min_count, (types.Omitted, types.Integer)) or min_count == 0):
         raise TypingError(
             '{} Unsupported parameters. Given axis: {}, level: {}, numeric_only: {}, min_count: {}'.format(
                 _func_name, axis, level, numeric_only, min_count))
 
-    def hpat_pandas_series_prod_impl(self, axis=None, skipna=True, level=None, numeric_only=None, min_count=0):
+    def hpat_pandas_series_prod_impl(self, axis=None, skipna=None, level=None, numeric_only=None, min_count=0):
+        if skipna is None:
+            skipna = True
```
kozlov-alexey: This should not compile, since type won't be unified for skipna variable which is None and bool at the same time. You can add a branch based on compile time value, i.e. define skipna_is_none variable at typing and refer to it with:
if skipna_is_none == True:  #noqa
    _skipna = True
else:
    _skipna = skipna

sdc/tests/test_dataframe.py
kozlov-alexey: @Rubtsowa Why haven't you applied the above comment? You don't need two different tests for min, you can just use one with:
```
        df = pd.DataFrame({
                           "A": [12, 4, 5, 44, 1],
                           "B": [5.0, np.nan, 9, 2, -1],
                           "C": ['a', 'aa', 'd', 'cc', None],
                           "D": [True, True, False, True, True]
         })
```

sdc/tests/test_dataframe.py
```diff
@@ -1221,28 +1213,144 @@ def test_impl(n):
...
+        hpat_func = sdc.jit(test_impl)
+        df = pd.DataFrame({"A": [.2, .0, .6, .2],
+                           "B": [.5, .6, .7, .8],
+                           "C": [2, 0, 6, 2],
+                           "D": [.2, .1, np.nan, .5],
+                           "E": [-1, np.nan, 1, np.inf],
+                           "F": [np.nan, np.nan, np.inf, np.nan]})
+        pd.testing.assert_series_equal(hpat_func(df), test_impl(df))
+
+    def test_prod(self):
```
kozlov-alexey: All these tests test default argument values, so I suggest adding _default suffix for all of them.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -36,8 +36,9 @@
 
 from numba import types
 from numba.extending import (overload, overload_method, overload_attribute)
-from sdc.hiframes.pd_dataframe_type import DataFrameType
 from numba.errors import TypingError
+from sdc.hiframes.pd_dataframe_ext import DataFrameType
+import sdc.datatypes.hpat_pandas_dataframe_types
```
AlexanderKalistratov:
```suggestion
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -36,8 +36,9 @@
 
 from numba import types
 from numba.extending import (overload, overload_method, overload_attribute)
-from sdc.hiframes.pd_dataframe_type import DataFrameType
 from numba.errors import TypingError
+from sdc.hiframes.pd_dataframe_ext import DataFrameType
```
AlexanderKalistratov:
```suggestion
from sdc.hiframes.pd_dataframe_type import DataFrameType
```

PR #347, opened 2019-11-28
Add CSV support for TestResults

PokhodenkoSA:
This PR makes it possible to run benchmarking with results aggregation (by LOAD_PREV_RESULTS=1) using CSV instead of Excel.

### DIFF
--- a/sdc/tests/tests_perf/test_perf_utils.py
+++ b/sdc/tests/tests_perf/test_perf_utils.py
@@ -195,20 +195,77 @@ def get_times(f, *args, iter_number=5):
     return exec_times, boxing_times
 
 
+class ResultsDriver:
+    """Base class. Load and dump results."""
+
+    def __init__(self, file_name, raw_file_name=None, logger=None):
+        self.file_name = file_name
+        self.raw_file_name = raw_file_name if raw_file_name else f'raw_{file_name}'
+        self.logger = logger
+
+
+class ExcelResultsDriver(ResultsDriver):
+    # openpyxl need to be installed
+
+    def dump_grouped_data(self, grouped_data):
+        try:
+            with pandas.ExcelWriter(self.file_name) as writer:
+                grouped_data.to_excel(writer)
+        except ModuleNotFoundError as e:
+            if self.logger:
+                msg = 'Could not dump the results to "%s": %s'
+                self.logger.warning(msg, self.file_name, e)
+
+    def dump_test_results_data(self, test_results_data):
+        try:
+            with pandas.ExcelWriter(self.raw_file_name) as writer:
+                test_results_data.to_excel(writer, index=False)
+        except ModuleNotFoundError as e:
+            if self.logger:
+                msg = 'Could not dump raw results to "%s": %s'
+                self.logger.warning(msg, self.raw_file_name, e)
+
+    def load(self):
+        raw_perf_results_xlsx = Path(self.raw_file_name)
+        if raw_perf_results_xlsx.exists():
+            with raw_perf_results_xlsx.open('rb') as fd:
+                # xlrd need to be installed
+                try:
+                    return pandas.read_excel(fd)
+                except ModuleNotFoundError as e:
+                    msg = 'Could not load previous results from %s: %s'
+                    self.logger.warning(msg, self.raw_file_name, e)
+
+class CSVResultsDriver(ResultsDriver):
+
+    def dump_grouped_data(self, grouped_data):
+        grouped_data.to_csv(self.file_name)
+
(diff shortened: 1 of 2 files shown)


sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -195,13 +195,78 @@ def get_times(f, *args, iter_number=5):
     return exec_times, boxing_times
 
 
+class ResultsDriver:
+    """Base class. Load and dump results."""
+
+    def __init__(self, file_name, raw_file_name=None):
+        self.file_name = file_name
+        self.raw_file_name = raw_file_name if raw_file_name else f'raw_{file_name}'
```
densmirn: It could be replaced with `self.raw_file_name = raw_file_name or f'raw_{file_name}'`

sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -195,13 +195,78 @@ def get_times(f, *args, iter_number=5):
...
+
+
+class ExcelResultsDriver(ResultsDriver):
+    # openpyxl need to be installed
+
+    def dump_grouped_data(self, grouped_data, logger=None):
```
densmirn: Would we like to make `logger` as object attribute, maybe of base class (`self.logger`)?

sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -195,13 +195,78 @@ def get_times(f, *args, iter_number=5):
...
+            with pandas.ExcelWriter(self.raw_file_name) as writer:
+                test_results_data.to_excel(writer, index=False)
+        except ModuleNotFoundError as e:
+            if logger:
+                msg = 'Could not dump raw results to "%s": %s'
+                logger.warning(msg, self.raw_file_name, e)
+
+    def load(self, logger=None):
+        raw_perf_results_xlsx = Path(self.raw_file_name)
+        if raw_perf_results_xlsx.exists():
```
densmirn: Let's decrease indentations via the following construction:
```
if not raw_perf_results_xlsx.exists():
    # maybe log here the fact that file doesn't exist
    return None
# remaining code
```

sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -195,13 +195,78 @@ def get_times(f, *args, iter_number=5):
...
+    def dump_test_results_data(self, test_results_data, logger=None):
+        test_results_data.to_csv(self.raw_file_name)
+
+    def load(self, logger=None):
+        raw_perf_results_csv = Path(self.raw_file_name)
+        if raw_perf_results_csv.exists():
+            with raw_perf_results_csv.open('rb') as fd:
+                # xlrd need to be installed
+                try:
+                    return pandas.read_csv(fd)
```
densmirn: Do we need extra modules for CSV loading?

sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -272,35 +337,18 @@ def dump(self):
...
-                # xlrd need to be installed
-                try:
-                    self.test_results_data = pandas.read_excel(fd)
-                except ModuleNotFoundError as e:
-                    msg = 'Could not load previous results from %s: %s'
-                    self.logger.warning(msg, raw_perf_results_xlsx, e)
+        if self.default_driver:
+            test_results_data = self.default_driver.load(self.logger)
+            if test_results_data is not None:
+                self.test_results_data = test_results_data
```
densmirn: I think we can try to load data via each driver from the all specified drivers until we get the data something like:
```
for d in self.drivers:
    test_results_data = d.load()
    if test_results_data:
        self.test_results_data = test_results_data
        break
```

sdc/tests/tests_perf/test_perf_utils.py
PokhodenkoSA: Driver is created before creating TestResults. Logger is created inside TestResults. I think working with logger should be redesigned completely. Not in this PR.

sdc/tests/tests_perf/test_perf_utils.py
PokhodenkoSA: I know this trick. This code is copied from the original.

sdc/tests/tests_perf/test_perf_utils.py
PokhodenkoSA: ok

Vyacheslav-Smirnov:
@PokhodenkoSA ,
Could you please add benchmark result's dump to html in current PR?

PR #349, opened 2019-11-28
Make PyArrow cpu_count to correspond NUMBA_NUM_THREADS

PokhodenkoSA:
This PR makes PyArrow treads count equal to Numba threads count.
I do not know which threading layer is used by PyArrow. it is possible that there is no intersection with Numba layers.
By default PyArrow uses all available cpus for reading.

### DIFF
--- a/sdc/io/csv_ext.py
+++ b/sdc/io/csv_ext.py
@@ -25,6 +25,8 @@
 # *****************************************************************************
 
 
+import functools
+
 import llvmlite.binding as ll
 from llvmlite import ir as lir
 from .. import hio
@@ -413,6 +415,7 @@ class pyarrow_cpu_count(object):
 
 def pyarrow_cpu_count_equal_numba_num_treads(func):
     """Decorator. Set pyarrow cpu_count the same as NUMBA_NUM_THREADS."""
+    @functools.wraps(func)
     def wrapper(*args, **kwargs):
         with pyarrow_cpu_count(numba.config.NUMBA_NUM_THREADS):
             return func(*args, **kwargs)
--- a/sdc/io/csv_ext.py
+++ b/sdc/io/csv_ext.py
@@ -25,6 +25,7 @@
 # *****************************************************************************
 
 
+import contextlib
 import functools
 
 import llvmlite.binding as ll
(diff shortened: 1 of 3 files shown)


sdc/io/csv_ext.py
```diff
@@ -398,6 +399,18 @@ def to_varname(string):
     return re.sub(r'\W|^(?=\d)','_', string)
 
 
+def pyarrow_cpu_count_equal_numba_num_treads(func):
+    """Decorator. Set pyarrow cpu_count the same as NUMBA_NUM_THREADS."""
+    def wrapper(*args, **kwargs):
+        cpu_count = pyarrow.cpu_count()
+        pyarrow.set_cpu_count(numba.config.NUMBA_NUM_THREADS)
+        result = func(*args, **kwargs)
```
Hardcode84: nitpick: if wrapped func raise exception you will miss second call to `set_cpu_count` (but I am not sure numba supports finally)

AlexanderKalistratov: @Hardcode84 as far as I get it this code is not jitted anyway, so `finally` should work

sdc/io/csv_ext.py
```diff
@@ -398,6 +399,28 @@ def to_varname(string):
...
+        self.old_cpu_count = pyarrow.cpu_count()
+        pyarrow.set_cpu_count(self.cpu_count)
+
+    def __exit__(self, type, value, traceback):
+        pyarrow.set_cpu_count(self.old_cpu_count)
+
+
+def pyarrow_cpu_count_equal_numba_num_treads(func):
+    """Decorator. Set pyarrow cpu_count the same as NUMBA_NUM_THREADS."""
+    def wrapper(*args, **kwargs):
```
densmirn: Consider using of `functools.wraps` here: https://docs.python.org/3/library/functools.html#functools.wraps.

sdc/io/csv_ext.py
```diff
@@ -24,6 +24,7 @@
 # EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *****************************************************************************
 
+import os
```
densmirn: Do you exactly need this import?

sdc/tests/tests_perf/test_perf_read_csv.py
```diff
@@ -53,14 +54,21 @@ def generate_csv(rows=10**5, columns=10):
...
     def _function():
         start = time.time()
         df = pandas.read_csv(file_name)
         return time.time() - start, df
+
+    if pyarrow:
+        def _function():
+            start = time.time()
+            df = sdc.io.csv_ext.pandas_read_csv(file_name)
+            return time.time() - start, df
```
densmirn: In case if `pyarrow is True` `_function()` is defined twice.

densmirn: Maybe do something like that:
```
if pyarrow:
    def _func():
        # ...
    return _func

def _func():
    # ...
return _func
```

sdc/tests/tests_perf/test_perf_read_csv.py
```diff
@@ -43,12 +43,8 @@ def generate_csv(rows=10**5, columns=10):
...
     r = random.Random(0)  # seed=0
     generate(rows,
-        ['A', 'B', 'C'],
-        [
-            lambda: int(r.random() * 10000),
-            lambda: r.uniform(-1.0, 1.0),
-            lambda: r.uniform(-1.0, 1.0)
-        ],
+        [f"f{c}" for c in range(columns)],
+        [lambda: r.uniform(-1.0, 1.0) for _ in range(columns)],
```
densmirn: Why not just call `r.uniform(-1.0, 1.0)` here without lambda?
`[r.uniform(-1.0, 1.0) for _ in columns]`

PokhodenkoSA: Because it is generators, not values.

PokhodenkoSA:
This PR looks very big. I will split it to many smaller ones with dedicated improvements. Now I use it for prepare benchmarking report only.

PR #351, opened 2019-11-28
Replace sdc.jit with self.jit for using numba.jit for new pipeline

PokhodenkoSA:


PokhodenkoSA:
I think this PR makes problems in CI.

PR #353, opened 2019-11-29
Fix corr1

Rubtsowa:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1200,7 +1200,20 @@ def hpat_pandas_series_corr_impl(self, other, method='pearson', min_periods=None
         if len(self_arr) < min_periods:
             return numpy.nan
 
-        return numpy.corrcoef(self_arr, other_arr)[0, 1]
+        new_self = pandas.Series(self_arr)
+        new_other = pandas.Series(other_arr)
```
densmirn: Can we avoid constructing of Series and use arrays directly?

Rubtsowa: used methods working only with Series.

PR #354, opened 2019-11-29
Add check_jitted() function for TestIO

PokhodenkoSA:
This PR closes #411.
It introduces method check_jitted() for comparing py and jit functions results. This method uses pretty universal method assertDataEqual() for comparing data.

### DIFF
--- a/sdc/tests/test_io.py
+++ b/sdc/tests/test_io.py
@@ -76,8 +76,6 @@ class TestIO(TestCase):
 
 class TestParquet(TestIO):
 
-    @unittest.skip('Error - fix needed\n'
-                   'NUMA_PES=3 build')
     def test_pq_read(self):
         def test_impl():
             t = pq.read_table('kde.parquet')
@@ -102,8 +100,6 @@ class TestParquet(TestIO):
         self.assertEqual(count_array_REPs(), 0)
         self.assertEqual(count_parfor_REPs(), 0)
 
-    @unittest.skip('Error - fix needed\n'
-                   'NUMA_PES=3 build')
     def test_pq_read_freevar_str1(self):
         kde_file2 = 'kde.parquet'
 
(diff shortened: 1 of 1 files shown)


sdc/tests/test_io.py
```diff
@@ -72,8 +72,30 @@ def setUp(self):
             A = np.random.ranf(n)
             A.tofile("np_file1.dat")
 
-    @unittest.skip('Error - fix needed\n'
-                   'NUMA_PES=3 build')
+    def assertDataEqual(self, left, right):
```
densmirn: Let's support passing of extra arguments through `*args` and `**kwargs`. Maybe it will be required.

sdc/tests/test_io.py
```diff
@@ -72,8 +72,30 @@ def setUp(self):
...
+        if isinstance(right, pd.Series):
```
densmirn: Why not to check also type of `left` value?

sdc/tests/test_io.py
PokhodenkoSA: @densmirn could you propose the modification?

sdc/tests/test_io.py
PokhodenkoSA: `right` is expected value. It is used only for chosing comparison algorithm.

AlexanderKalistratov:
@densmirn if you are fine with this PR could you please approve it?

sdc/tests/test_io.py
densmirn: E.g. let's consider parameter `check_names=False`. You could pass the parameter to assert via `kwargs`. You could make as below:
```
def assertDataEqual(self, left, right, *args, **kwargs):
    # ...
    pd.testing.assert_series_equal(left, right, *args, check_categorical=False, **kwargs)
    # ...
```

densmirn:
> @densmirn if you are fine with this PR could you please approve it?

I left the latest comment from my side. After applying the comment I will approve the PR.

Hardcode84:
outdated

PokhodenkoSA:
> outdated

@Hardcode84 I suggest to merge it after release. It is not necessary now.

PR #356, opened 2019-11-29
Introducing Rewrite which implements DataFrame constructor

AlexanderKalistratov:
Introducing **rewrite** which replaces `DataFrame({'A': a, 'B': b})` with `init_dataframe(a, b, index, 'A', 'B')`
This is a temporally solution and the only reason it is existed is lack of support in Numba of literal lists and dicts (and dicts of lists?). And should be removed as soon as such support appears.

All current tests should works in the same way as before (except few tests which unexpectedly get broken due to unused columns elimination stopped to work)

Introduced new test: **test_create_string_index**. This test wouldn't work without **rewrite**


sdc/rewrites/dataframe_constructor.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+    def apply(self):
+        init_df_stmt = import_function(init_dataframe, self._block, self._func_ir)
+
+        for stmt, extract, cleanup in self._calls_to_rewrite:
+            args = get_call_parameters(call=stmt.value, arg_names=self._df_arg_list)
+
+            old_data = args['data']
+            old_columns = args['columns'] if 'columns' in args else None
```
densmirn: Looks like this line is equal to `old_columns = args.get('columns')`.

sdc/rewrites/dataframe_constructor.py
```diff
@@ -0,0 +1,200 @@
...
+
+        return data_args, columns_args
+
+
+    @staticmethod
+    def _extract_dict_args(args, block, func_ir):
+        dict_items = get_dict_items(args['data'], block, func_ir)
+
+        data_args = [item[1] for item in dict_items]
+        columns_args = [item[0] for item in dict_items]
```
densmirn: Probably you can use `dict_items.keys()` and `dict_items.values()` here.

sdc/rewrites/dataframe_constructor.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+    @staticmethod
+    def _replace_call(stmt, new_call, args, block, func_ir):
+        func = stmt.value
+
+        data_args = args['data']
+        columns_args = args['columns']
+
+        index_args = args['index'] if 'index' in args else None
```
densmirn: `index_args = args.get('index')`

sdc/rewrites/dataframe_constructor.py
```diff
@@ -0,0 +1,200 @@
...
+                if is_list(var, block, func_ir):
+                    call_stmt = make_call(fix_df_array, [var], {}, block, func_ir, var.loc)
+                    insert_before(block, call_stmt, stmt)
+                    new_args.append(call_stmt.target)
+                else:
+                    new_args.append(var)
+
+            return new_args
+
+        raise ValueError()
```
densmirn: Do not we want to add message for the exception?

sdc/rewrites/dataframe_constructor.py
AlexanderKalistratov: Unfortunately no. dict_items is list of tuples, not dict

sdc/rewrites/dataframe_constructor.py
AlexanderKalistratov: Thanks, fixed

sdc/rewrites/dataframe_constructor.py
AlexanderKalistratov: Completely removed

sdc/rewrites/dataframe_constructor.py
AlexanderKalistratov: Thanks, fixed

AlexanderKalistratov:
@shssf @kozlov-alexey please pay attention. This PR is disabling workaround with lightweight passes and skips a bunch of tests

AlexanderKalistratov:
@shssf are you ok how workaround is disabled? It is just commented. Are you fine with this? Or do you want to have special config parameter?

PR #358, opened 2019-12-02
DataFrame Sort in new style. part 1

shssf:


sdc/hiframes/dataframe_pass.py
```diff
@@ -76,6 +76,18 @@ def __init__(self, state):
         self.state = state
 
     def run_pass(self):
+        """
+        The function could return exxeption. It means that the IR transformation can not be completed.
```
densmirn: typo: `exxeption` -> `exception`

PR #359, opened 2019-12-02
Skip hanging tests in test_basic

PokhodenkoSA:
This PR fixes #351. There are some difference between sdc.jit and numba.jit:
1. sdc.jit do not set parallel=True.

### DIFF
--- a/sdc/tests/test_basic.py
+++ b/sdc/tests/test_basic.py
@@ -86,6 +86,7 @@ class BaseTest(TestCase):
 
 class TestBasic(BaseTest):
 
+    @skip_numba_jit("hang with numba.jit. ok with sdc.jit")
     def test_getitem(self):
         def test_impl(N):
             A = np.ones(N)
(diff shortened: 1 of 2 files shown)


PokhodenkoSA:
One test in old pipeline - test_str_split_filter - fails because #321.

densmirn:
> Hello @PokhodenkoSA! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`sdc/tests/test_hiframes.py`](https://github.com/IntelPython/sdc/blob/e3b2817d141cda0bd76eaeae80d7aaf70815e558/sdc/tests/test_hiframes.py):
> 
> > [Line 47:36](https://github.com/IntelPython/sdc/blob/e3b2817d141cda0bd76eaeae80d7aaf70815e558/sdc/tests/test_hiframes.py#L47): [E127](https://duckduckgo.com/?q=pep8%20E127) continuation line over-indented for visual indent

Could you fix the indentations issue?

PokhodenkoSA:
> Could you fix the indentations issue?

Not in this PR. This code is inline with other lines before.
I think introducing new style is better with one big PR with autostyling tool like **black**. Otherwise it will confuse people.

PR #360, opened 2019-12-02
Overload Series(StringArraySplitView).str.len()

densmirn:


sdc/hiframes/split_impl.py
```diff
@@ -131,6 +135,42 @@ def __init__(self, dmm, fe_type):
 make_attribute_wrapper(StringArraySplitViewType, 'data', '_data')
 
 
+class SplitViewStringMethodsType(types.IterableType):
```
shssf: How this class related to `pandas.core.strings.StringMethods`? Why we need it?

densmirn: This class related to the pandas string methods through existing unit test:
```
    def test_str_split_filter(self):
        def test_impl(df):
            B = df.A.str.split(',')
            df2 = pd.DataFrame({'B': B})
            return df2[df2.B.str.len() > 1]
        # ...
```
where `split()` returns `Series with StringArraySplitView`. So I made `str.len()` to work with such data this way. Otherwise we get something like that:
`Invalid store of {{i64, i32*, i32*, {i8*, i8*}, i8*}, i8*, {i8*, i64, i32, i32, i64, i8*, i8*}} to {{i64, i64, i32*, i8*, i8*, i8*}, i8*, {i8*, i64, i32, i32, i64, i8*, i8*}}`

PR #361, opened 2019-12-02
Implement getitem

1e-to:
Branch units implementation of getitem/loc/iloc/at/iat and tests for them


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -117,78 +123,217 @@ def hpat_pandas_series_getitem(self, idx):
...
-            """
-            **Test**: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_iloc1
-            """
-
-            result = self._data[idx]
-            return result
-
-        return hpat_pandas_series_getitem_idx_integer_impl
-
+    if isinstance(self, SeriesOperatorTypeIloc):
```
shssf: Could you please use `self._getitem_selector` variable instead classes like `SeriesOperatorTypeIloc`?

1e-to: Yes, I'll see how to do it.

1e-to: @shssf I made a selector. What do you think about this code?

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,80 @@
...
+from numba.extending import (models, overload, register_model, make_attribute_wrapper, intrinsic)
+from numba.datamodel import (register_default, StructModel)
+from numba.typing.templates import signature
+
+
+class SeriesGetitemSelectorType(types.IterableType):
+    def __init__(self, data, selector):
+        self.data = data
+        self.selector = selector
+        super(SeriesGetitemSelectorType, self).__init__('SeriesGetitemSelectorType')
```
AlexanderKalistratov: please try the following code:
`super(SeriesGetitemSelectorType, self).__init__('SeriesGetitemSelectorType({}, {})'.format(data, selector))`

Your issue is the same as @densmirn is fighting with. So, you could ask him about details.

Also, even if it would helps it would berequired to perform some additional tests

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@intrinsic
+def series_getitem_selector_init(typingctx, data, selector=0):
+    def series_getitem_selector_init_codegen(context, builder, signature, args):
+        data_val, selector_val = args
+        getitem_selector = cgutils.create_struct_proxy(
+            signature.return_type)(context, builder)
+        getitem_selector.data = data_val
+        getitem_selector.selector = selector_val
```
AlexanderKalistratov: shouldn't it be `getitem_selector.selector = selector_var.literal_value`?

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if context.enable_nrt:
+            context.nrt.incref(builder, signature.args[0], data_val)
+            context.nrt.incref(builder, signature.args[1], selector_val)
```
AlexanderKalistratov: I don't think you should do `incref` for `int` value.

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pandas
+
+from numba import types, cgutils
+from numba.extending import (models, overload, register_model, make_attribute_wrapper, intrinsic)
+from numba.datamodel import (register_default, StructModel)
+from numba.typing.templates import signature
+
+
+class SeriesGetitemSelectorType(types.IterableType):
```
shssf: @1e-to Why we need standalone, special class for `getitem()`? Why we can't just use some variable inside `SeriesType` class to provide required information to `[]` operator?

sdc/datatypes/hpat_pandas_series_functions.py
shssf: @1e-to I think it is more complicated than I thought. I didn't mean new class implementation. I thought about some small variable only which set by, for example, `iloc` method to the Series constructor and read by `getitem` implementation to distinguish these types.

sdc/datatypes/hpat_pandas_getitem_types.py
1e-to: Variable inside class makes the series model heavier, plus using this option we will not be able to support the functionality:
a = S.loc
a[1]

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -103,77 +186,172 @@ def hpat_pandas_series_getitem(self, idx):
...
         def hpat_pandas_series_getitem_idx_series_impl(self, idx):
-            """
-            **Test**: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_setitem_series_bool2
-            """
-            super_index = idx._data
-            result = self._data[super_index]
-            return result
+            res = numpy.copy(self._data[:len(idx._data)])
+            for i in range(len(res)):
+                for j in range(len(self.index)):
```
densmirn: I would say we should use `numba.prange` instead of python `range` to achieve better performance.

sdc/tests/test_series.py
```diff
@@ -4548,7 +4673,7 @@ def test_impl(s, axis):
             msg = 'Method cumsum(). Unsupported parameters. Given axis: int'
             self.assertIn(msg, str(raises.exception))
 
-    @skip_sdc_jit('Series.cov() parameter "min_periods" unsupported')
+    @unittest.skip("not work")
```
densmirn: Could you provide the reason here "why does it not work"?

1e-to: Fixed

sdc/datatypes/hpat_pandas_series_functions.py
1e-to: Fixed

AlexanderKalistratov:
@1e-to getitem example fails

sdc/tests/test_series.py
```diff
@@ -3498,8 +3641,8 @@ def test_impl(S):
         for n in range(1, 5):
             S = pd.Series(['a', 'ab', 'abc', 'c', 'f', 'hh', ''] * n)
             start, end = get_start_end(len(S))
-            pd.testing.assert_series_equal(hpat_func(S[start:end]), test_impl(S))
-            self.assertTrue(count_array_OneDs() > 0)
+            pd.testing.assert_series_equal(hpat_func(S[start:end]), test_impl(S[start:end]))
+            # self.assertTrue(count_array_OneDs() > 0)
```
kozlov-alexey: I don't think it's a good idea to remove the check that array was distributed. This test when run in the old pipeline should not be affected by adding impl via new one.

sdc/hiframes/pd_series_ext.py
```diff
@@ -1037,72 +1037,72 @@ def generic_expand_cumulative_series(self, args, kws):
...
-                    sig.return_type = if_arr_to_series_type(sig.return_type)
-                    arg1 = if_arr_to_series_type(arg1)
-                if is_idx_series:
-                    arg2 = if_arr_to_series_type(arg2)
-                sig.args = (arg1, arg2)
-                # dt_index and Series(dt64) should return Timestamp
-                if is_arr_dt_index and sig.return_type == types.NPDatetime('ns'):
-                    sig.return_type = pandas_timestamp_type
-            return sig
+# PR135. This needs to be commented out
```
kozlov-alexey: By doing this you will affect the old-style pipeline, i.e. impl via overload will always be called.
This (and in other places) needs to be executed, but only when:
if sdc.config.config_pipeline_hpat_default:

sdc/hiframes/pd_series_ext.py
```diff
@@ -434,18 +434,18 @@ def resolve_dt(self, ary):
...
-        return SeriesIatType(ary)
+    # def resolve_iloc(self, ary):
+    #     # TODO: support iat/iloc differences
+    #     return SeriesIatType(ary)
 
 # PR135. This needs to be commented out
-    def resolve_loc(self, ary):
-        # TODO: support iat/iloc differences
-        return SeriesIatType(ary)
+    # def resolve_loc(self, ary):
```
kozlov-alexey: All these resolve_ functions should stay uncommented, as they are used in the old pipeline.
Instead of commenting them add them to the following list: _non_hpat_pipeline_attrs

sdc/hiframes/hiframes_typed.py
```diff
@@ -943,30 +943,30 @@ def run_call_series_quantile_default(A):
...
-            nodes = []
-            n_arg = self._get_arg('Series.head', rhs.args, dict(rhs.kws), 0,
-                                  'n', default=False)  # TODO: proper default handling
-            if n_arg is False:
-                n_arg = ir.Var(lhs.scope, mk_unique_var('head_n'), lhs.loc)
-                # default is 5
-                self.state.typemap[n_arg.name] = types.IntegerLiteral(5)
-                nodes.append(ir.Assign(
-                    ir.Const(5, lhs.loc), n_arg, lhs.loc))
+        # if func_name == 'head':
```
kozlov-alexey: This should not be commented, as it provides impl for head for old pipeline.

sdc/hiframes/hiframes_typed.py
```diff
@@ -222,8 +222,8 @@ def _run_assign(self, assign):
                 return self._run_unary(assign, rhs)
 
             # replace getitems on Series.iat
-            if rhs.op in ('getitem', 'static_getitem'):
-                return self._run_getitem(assign, rhs)
+            # if rhs.op in ('getitem', 'static_getitem'):
```
kozlov-alexey: Same comment as below

kozlov-alexey:
> Hello @1e-to! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`sdc/datatypes/hpat_pandas_series_functions.py`](https://github.com/IntelPython/sdc/blob/d31b0935d7947e2fd97d11a844f6d305ab3d44d7/sdc/datatypes/hpat_pandas_series_functions.py):
> 
> > [Line 108:5]
> .........
> ##### Comment last updated at 2019-12-16 15:39:17 UTC

Please also fix the PEP style issues.

sdc/hiframes/pd_series_ext.py
AlexanderKalistratov: @kozlov-alexey  I thought we are keeping old functionality only in case overload implementation is incompatible with the old pipeline. In case it is we can simply remove old implementation. Was I wrong?

kozlov-alexey: @AlexanderKalistratov It seems there will always be some impact, like here (if I understood correctly) combining new-style getitem with old-style distributed impl for head causes parallel test to fail (distribution report line was commented). I'm not sure, but it might actually be better no to change anything in old-style pipeline at all, just to ensure we keep it working (or more precisely, the same state as it was).

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -104,77 +210,194 @@ def hpat_pandas_series_getitem(self, idx):
...
         def hpat_pandas_series_getitem_idx_series_impl(self, idx):
-            """
-            **Test**: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_setitem_series_bool2
-            """
-            super_index = idx._data
-            result = self._data[super_index]
-            return result
+            if (idx._data[0] is True or idx._data[0] is False):
+                return pandas.Series(self._data[idx._data], self._index[idx._data])
+            res = numpy.copy(self._data[:len(idx._data)])
```
AlexanderKalistratov: That's not valid.
```
s = pd.Series([0, 1, 2], [0, 0, 0])
print(s[[0]])
0    0
0    1
0    2
dtype: int64
```

So, output size is larger than size of idx._data

Also, i think it would be better to use dict here

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,114 @@
...
 from sdc.hiframes.pd_series_ext import SeriesType
 from sdc.str_arr_ext import (StringArrayType, cp_str_list_to_array, num_total_chars, string_array_type,
                              str_arr_is_na, pre_alloc_string_array, str_arr_set_na)
 from sdc.utils import to_array, sdc_overload, sdc_overload_method, sdc_overload_attribute
 
 
+@sdc_overload(operator.getitem)
+def hpat_pandas_series_accessor_getitem(self, idx):
+    """
+    Pandas Series operator :attr:`pandas.Series.get` implementation
```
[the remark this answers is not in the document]
1e-to: What should I write inplace of the getitem_accessor dosc?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,114 @@
...
+
+            return hpat_pandas_series_iat_impl
+
+        raise TypingError('{} The index must be a Integer. Given: {}'.format(_func_name, idx))
+
+    if accessor == 'loc':
+    #note: Loc return Series
+        index_is_none = (self.series.index is None or
+                         isinstance(self.series.index, numba.types.misc.NoneType))
+        if isinstance(idx, types.SliceType) and not index_is_none:
```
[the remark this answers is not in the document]
1e-to: I do it in 112 line

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,114 @@
...
+            def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
+                start = -1
+                stop = -1
+                index = numpy.arange(len(self._series._data))
+                for i in numba.prange(len(index)):
+                    if index[i] == idx.start and start == -1:
+                        start = i
+                    if index[i] == idx.stop and start != -1 and stop == -1:
+                        stop = i 
+                return pandas.Series(self._series._data[start:stop+1], index[start:stop+1])
```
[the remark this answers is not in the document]
1e-to: Yes, fixed

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,114 @@
...
+        raise TypingError('{} The index must be a Integer. Given: {}'.format(_func_name, idx))
+
+    if accessor == 'loc':
+    #note: Loc return Series
+        index_is_none = (self.series.index is None or
+                         isinstance(self.series.index, numba.types.misc.NoneType))
+        if isinstance(idx, types.SliceType) and not index_is_none:
+            def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
+                start = -1
+                stop = -1
```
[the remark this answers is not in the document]
1e-to: Everything will be fine, because we remember the position of the argument, not the index itself. The position is always positive.

sdc/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
AlexanderKalistratov: Does slice have nested type? ['a':'z'] - slice of strings. [0:5] - slice of integers

sdc/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
AlexanderKalistratov: Yes, you are right

sdc/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
1e-to: We cant work with string slice, start stop for them not supported

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,114 @@
...
+                for i in numba.prange(len(self._series._index)):
+                    if self._series._index[i] == idx.start and start == -1:
+                        start = i
+                    if self._series._index[i] == idx.stop and start != -1 and stop == -1:
+                        stop = i 
+                return pandas.Series(self._series._data[start:stop+1], self._series._index[start:stop+1])
```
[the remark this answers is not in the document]
1e-to: It seems to me that we should write a restriction for users that the slice supports only the specified start and end. Because if they are not specified, start = 0 and end = 9223372036854775807 come to the registered function, in the case of non-standard indexes we cannot distinguish the data that the user set from the default

sdc/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
AlexanderKalistratov: I see

sdc/datatypes/hpat_pandas_series_functions.py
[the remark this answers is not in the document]
AlexanderKalistratov: How pandas behave in such cases?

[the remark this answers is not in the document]
AlexanderKalistratov: 9223372036854775807 + 1 is zero. So, in case of stop == 9223372036854775807 you souldn't do +1

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +50,168 @@
...
+                        if self._series._index[i-1] > self._series._index[i]:
+                            dec = 1
+                        else:
+                            inc = 1
+                    if self._series._index[i] == idx.start and start == -1:
+                        start = i
+                    if self._series._index[i] == idx.stop and start != -1 and stop == -1:
+                        stop = i
+                if inc + dec != 1:
+                    is_monotonic = False
```
densmirn: You could use `np.diff()` to check if an array is monotonic, e.g.:
```
def monotonic(x):
    dx = np.diff(x)
    return np.all(dx <= 0) or np.all(dx >= 0)
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +50,168 @@
...
+                                if t < start_value:
+                                    tmp = i
+                                    start_value = t
+                            else:
+                                t = (math.fabs(idx.start -
+                                    self._series._index[i]))
+                                if t < start_value:
+                                    tmp = i
+                                    start_value = t
+                    start = tmp
```
densmirn: Could we move it to a function and call the func twice (here and below)?

examples/series_at/series_at_many_idx.py
```diff
@@ -0,0 +1,40 @@
...
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_at_many_idx():
+    series = pd.Series([5, 4, 3, 2, 1], index=[0, 2, 0, 6, 0])
+    return series.at[0]
+    # Expected Array:
+    # [5 3 1]
```
kozlov-alexey: Not sure but it might be clearer to name these as "series_at_single_result" and "series_at_multiple_result".

examples/series_getitem/series_getitem_array.py
```diff
@@ -0,0 +1,49 @@
...
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+# OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+import numpy as np
+import pandas as pd
+from numba import njit
+
+from numba import njit
```
kozlov-alexey: The lines 29 and 31 are identical

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,76 @@
...
+class SeriesGetitemAccessorType(types.IterableType):
+    def __init__(self, series, accessor):
+        self.series = series
+        self.accessor = accessor
+        super(SeriesGetitemAccessorType, self).__init__('SeriesGetitemAccessorType({}, {})\
+            '.format(series, accessor))
+
+    @property
+    def iterator_type(self):
+        return None
```
kozlov-alexey: At least some comment (TODO?) is needed here: why we make this class to be derived from IterableType and then use a stub to register absence of iterator support in Numba? From the first glance accessors might not be iterable at all, am I wrong?

sdc/datatypes/hpat_pandas_getitem_types.py
```diff
@@ -0,0 +1,76 @@
...
+        models.StructModel.__init__(self, dmm, fe_type, members)
+
+
+make_attribute_wrapper(SeriesGetitemAccessorType, 'series', '_series')
+
+
+@intrinsic
+def series_getitem_accessor_init(typingctx, series, accessor):
+    def series_getitem_accessor_init_codegen(context, builder, signature, args):
+        series_val, accessor_val = args
```
kozlov-alexey: Where do we use accessor_val and why don't we store it to the struct?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+    """
+    Pandas Series operator :attr:`pandas.Series.get` implementation
+    **Algorithm**: result = series[idx]
+
+    **Test**: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_static_getitem_series1
+
+    Parameters
+    ----------
+    series: :obj:`pandas.Series`
+           input series
```
kozlov-alexey: Please use the same indentation for descriptions of parameters

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+    idx: :obj:`int`, :obj:`slice` or :obj:`pandas.Series`
+        input index
+
+    Returns
+    -------
+    :class:`pandas.Series` or an element of the underneath type
```
kozlov-alexey:
```suggestion
    :class:`pandas.Series` or a value of :obj:`pandas.Series.dtype`
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+            object of :class:`pandas.Series`
+    """
+
+    _func_name = 'Operator getitem().'
+
+    if not isinstance(self, SeriesGetitemAccessorType):
```
kozlov-alexey: I think a comment here (or maybe in the docstring) might be useful to explain why we are returning None here, i.e. "This operator.getitem overload is to handle this specific case of getitem on SeriesGetitemAccessorType object"

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+        return None
+
+    accessor = self.accessor.literal_value
+
+    if accessor == 'iloc':
+        if isinstance(idx, types.SliceType):
+            def hpat_pandas_series_iloc_slice_impl(self, idx):
+                return pandas.Series(self._series._data[idx])
```
kozlov-alexey: Is it correct to return series without name and index here? Shouldn't it be 
```suggestion
                result_data = self._series._data[idx]
                result_index = self._series.index[idx]
                return pandas.Series(data=result_data, index=result_index, name=self._series._name)
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+
+                def search(idx, choice):
+                    x_value = max_slice
+                    tmp = 0
+                    for i in numba.prange(len(self._series._index)):
+                        case_start = idx < self._series._index[i]
+                        case_stop = idx > self._series._index[i]
+                        if choice == 0 and case_start or choice == 1 and case_stop:
+                            if (idx < 0 and self._series._index[i] > 0 or idx > 0 and self._series._index[i] < 0):
+                                t = math.fabs(idx) + math.fabs(self._series._index[i])
```
kozlov-alexey: It's not clear what you're trying to achieve with this function. Maybe it's because of the lack of the comments, but it might also be that it can be rewritten in a simplier way / moved from here to a separate file as a common utility function for slices.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -46,12 +51,146 @@
...
+                if not is_monotonic and start == -1 and stop == -1:
+                    raise ValueError("Incorrect values entered")
+                return pandas.Series(self._series._data[start:stop+1], self._series._index[start:stop+1])
+
+            return hpat_pandas_series_getitem_idx_slice_impl
+
+        if isinstance(idx, types.SliceType) and index_is_none:
+            def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
+                index = numpy.arange(len(self._series._data))
+                return pandas.Series(self._series._data[idx.start:idx.stop+1], index[idx.start:idx.stop+1])
```
kozlov-alexey: Again, should we keep series name in result?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -105,77 +244,212 @@ def hpat_pandas_series_getitem(self, idx):
...
+        return None
 
-    if isinstance(idx, types.Integer):
-        def hpat_pandas_series_getitem_idx_integer_impl(self, idx):
-            """
-            **Test**: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_iloc1
-            """
+    # Note: Getitem return Series
+    index_is_none = self.index is None or isinstance(self.index, numba.types.misc.NoneType)
+    index_is_number = index_is_none or (self.index and isinstance(self.index.dtype, types.Number))
```
kozlov-alexey: Hmm, is this valid? There should be check that self.index is of types.Array type somewhere.
Also it's very confusing to call index_is_number when it can also be None. These are different cases, which may require different handling later. In any case if you want to check for "None or numeric index" just call it that way: index_is_none_or_numeric

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -105,77 +244,212 @@ def hpat_pandas_series_getitem(self, idx):
...
+    index_is_string = not index_is_none and isinstance(self.index.dtype, (types.UnicodeType, types.StringLiteral))
+
+    if (
+        isinstance(idx, types.Number) and index_is_number or
+        (isinstance(idx, (str, types.UnicodeType, types.StringLiteral)) and index_is_string)
```
kozlov-alexey: There's no need to check for str here, as idx has no default value, and hence it should never have pure Python type, i.e. (types.UnicodeType, types.StringLiteral)

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -104,77 +210,194 @@ def hpat_pandas_series_getitem(self, idx):
...
+        def hpat_pandas_series_getitem_index_impl(self, idx):
+            mask = numpy.empty(len(self._data), numpy.bool_)
+            for i in numba.prange(len(self.index)):
+                mask[i] = self.index[i] == idx
+            return pandas.Series(self._data[mask], self.index[mask])
 
-        return hpat_pandas_series_getitem_idx_integer_impl
+        return hpat_pandas_series_getitem_index_impl
+
+    if (isinstance(idx, types.Integer) and index_is_string):
```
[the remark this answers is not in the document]
kozlov-alexey: Yes, definitely a comment is needed here, to explain Pandas behavior.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -105,77 +244,212 @@ def hpat_pandas_series_getitem(self, idx):
...
 
-        return hpat_pandas_series_getitem_idx_integer_impl
+    if (isinstance(idx, types.Integer) and index_is_string):
+        def hpat_pandas_series_idx_impl(self, idx):
+            return self._data[idx]
+
+        return hpat_pandas_series_idx_impl
 
     if isinstance(idx, types.SliceType):
+        # Return slice for str values not implement
```
kozlov-alexey: Actually it's implemented (might not be working though), see (in str_arr_ext.py):
```
@lower_builtin(operator.getitem, StringArrayType, types.SliceType)
def lower_string_arr_getitem_slice(context, builder, sig, args):
```

sdc/datatypes/hpat_pandas_getitem_types.py
AlexanderKalistratov: We don't need `accessor_val`. Accessor type is saved in type information.

kozlov-alexey: Then I suggest writing it (otherwise it really looks like we missed something in code):
```suggestion
        series_val, _ = args
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,169 @@
...
+
+        raise TypingError('{} The index must be a Integer. Given: {}'.format(_func_name, idx))
+
+    if accessor == 'loc':
+        # Note: Loc return Series
+        # Note: Index 0 in slice not supported
+        index_is_none = (self.series.index is None or
+                         isinstance(self.series.index, numba.types.misc.NoneType))
+        if isinstance(idx, types.SliceType) and not index_is_none:
+            def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
```
AlexanderKalistratov: Here is two options how this function could be implemented:

Serial:
```python
def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
    start_position = 0
    stop_position = 0
    
    series = self._series
    index = series.index

    if idx.start <= idx.stop:
        start_cmp = lambda indx, strt: indx < strt
        stop_cmp = lambda indx, stp: indx >= stp
        mono_cmp = lambda diff: diff > 0
    else:
        start_cmp = lambda indx, strt: indx > strt
        stop_cmp = lambda indx, stp: indx <= stp
        mono_cmp = lambda diff: diff < 0

    while start_cmp(index[start_position], idx.start) and start_position < len(index):
        start_position += 1

    stop_position = start_position
    prev_value = index[stop_position]
    while stop_cmp(index[stop_position], idx.stop) and stop_position < len(index):
        if not mono_cmp(prev_value - index[stop_position]):
            raise ValueError("index must be monotonic increasing or decreasing")
        prev_value = index[stop_position]
        stop_position += 1

    if start_position >= len(index) or stop_position <= 0 or stop_position <= start_position:
        return pd.Series(data=[], index=[], name=series._name)
    
    return pandas.Series(data=series._data[start_position:stop_position], 
                         index=index[start_position:stop_position], 
                         name=series._name)
```

Potentially parallel:
```python
def hpat_pandas_series_getitem_idx_slice_impl(self, idx):
    series = self._series
    index = series.index

    start_position = len(index)
    stop_position = 0
    max_diff = 0
    min_diff = 0
    for i in prange(len(index)):
        if index[i] >= idx.start:
            start_position = min(start_position, i)
        if index[i] <= idx.stop:
            stop_position = max(stop_position, i)
        if i > 0:
            max_diff = max(max_diff, index[i] - index[i - 1])
            min_diff = min(min_diff, index[i] - index[i - 1])
    
    if max_diff*min_diff < 0:
        raise ValueError("index must be monotonic increasing or decreasing")

    if start_position >= len(index) or stop_position <= 0 or stop_position <= start_position:
        return pd.Series(data=[], index=[], name=series._name)
    
    return pandas.Series(data=series._data[start_position:stop_position], 
                         index=index[start_position:stop_position], 
                         name=series._name)
```

Please note, I haven't verified either of them, so it may contain bugs.
Also behavior of parallel version is slightly different from behavior of serial one

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,12 +49,169 @@
...
+                result_data = self._series._data[start:stop+1]
+                result_index = index[start:stop+1]
+                return pandas.Series(result_data, result_index, self._series._name)
+
+            return hpat_pandas_series_getitem_idx_slice_impl
+
+        if isinstance(idx, (int, types.Integer, types.UnicodeType, types.StringLiteral)):
+            def hpat_pandas_series_loc_impl(self, idx):
+                index = self._series.index.copy()
+                mask = numpy.empty(len(self._series._data), numpy.bool_)
```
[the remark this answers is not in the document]
1e-to: We cant do this for UnicodeString type

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -45,13 +49,198 @@
...
+
+                return pandas.Series(data=series._data[start_position:stop_position],
+                                     index=index[start_position:stop_position],
+                                     name=series._name)
+
+            return hpat_pandas_series_loc_slice_impl
+
+        if isinstance(idx, types.SliceType) and index_is_none:
+            def hpat_pandas_series_loc_slice_noidx_impl(self, idx):
+                max_slice = sys.maxsize
```
AlexanderKalistratov: Have you verified that `sys.maxsize` is the same value which `numba` is using to identified max slice value?

Do you have tests for `[start:]` and `[:]` cases?

1e-to: Tests for this cases added, functional loc callable and array delete

sdc/tests/test_series.py
```diff
@@ -1284,6 +1258,46 @@ def test_impl(A):
...
+        hpat_func = self.jit(test_impl)
+
+        all_data = [[1, 3, 5, 13, 22], [1, 3, 3, 13, 22]]
+        key = [1, 3, 18]
+        for index in all_data:
+            for n in key:
+                S = pd.Series([2, 4, 6, 6, 3], index)
+                pd.testing.assert_series_equal(hpat_func(S, n), test_impl(S, n))
+
+    @skip_sdc_jit('Not impl in old style')
```
AlexanderKalistratov: Please add tests also for decreasing index

1e-to: I added tests for decreasing index and they failed. Need to rewrite the algorithm

PR #365, opened 2019-12-04
Performance tests on series methods in workload

akharche:


sdc/tests/test_utils.py
```diff
@@ -43,8 +43,8 @@
 max_float64 = np.finfo('float64').max
 
 test_global_input_data_float64 = [
-    [1., np.nan, -1., 0., min_float64, max_float64, max_float64, min_float64],
```
shssf: What is the reason to delete `np.nan` here?

akharche: It is needed for astype test because Series.astype() can convert only finite values. I can add the same list without nan

PR #363, opened 2019-12-03
Dataframe.head()

1e-to:

PR #366, opened 2019-12-05
add reduce

Rubtsowa:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,38 @@
...
 
 from numba import types
 from numba.extending import (overload, overload_method, overload_attribute)
+from sdc.hiframes.pd_dataframe_ext import DataFrameType
 from numba.errors import TypingError
-
-from sdc.datatypes.hpat_pandas_dataframe_types import DataFrameType
-
-
-@overload_method(DataFrameType, 'count')
```
densmirn: Why did you remove the overload?

densmirn: Looks like this can be done in separate PR.

shssf: This overload_method `count` is not used. It is ok to remove it. It looks like implementation of `sdc_pandas_dataframe_reduce_columns` function is preparation for other PR #363 
not sure in it.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,38 @@
...
+        func_lines.append('  {} = {}.{}()'.format(d + '_O', d + '_S', name))
+    func_lines.append('  data = np.array(({},))'.format(
+        ", ".join(d + '_O' for d in data_args)))
+    func_lines.append('  index = hpat.str_arr_ext.StringArray(({},))'.format(
+        ', '.join('"{}"'.format(c) for c in saved_columns)))
+    func_lines.append('  return hpat.hiframes.api.init_series(data, index)')
+    loc_vars = {}
+    func_text = '\n'.join(func_lines)
+
+    exec(func_text, {'hpat': sdc, 'np': numpy}, loc_vars)
```
kozlov-alexey: @Rubtsowa I think we should stick to the new name:
```suggestion
    exec(func_text, {'sdc': sdc, 'np': numpy}, loc_vars)
```
and use sdc.hiframes instead hpat.hiframes in the generated function text too.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,38 @@
...
+    data_args = tuple('data{}'.format(i) for i in range(n_cols))
+    all_params = ['df']
+    for key, value in params:
+        all_params.append('{}={}'.format(key, value))
+    func_definition = 'def _reduce_impl({}):'.format(', '.join(all_params))
+    func_lines = [func_definition]
+    for i, d in enumerate(data_args):
+        line = '  {} = hpat.hiframes.api.init_series(hpat.hiframes.pd_dataframe_ext.get_dataframe_data(df, {}))'
+        func_lines.append(line.format(d + '_S', i))
+        func_lines.append('  {} = {}.{}()'.format(d + '_O', d + '_S', name))
```
kozlov-alexey: Why do we call the method of series without any arguments? 
You have to actually forward the arguments passed to DF call into the Series method. And also take care about how they actually map to each other (e.g. for DF.mean and Series.mean arguments are the same, but will it be so for all functions?)

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,64 +31,38 @@
...
-
-    return sdc_pandas_dataframe_count_impl
+from sdc.datatypes.hpat_pandas_series_functions import TypeChecker
+
+def sdc_pandas_dataframe_reduce_columns(df, name, params):
+    saved_columns = df.columns
+    n_cols = len(saved_columns)
+    data_args = tuple('data{}'.format(i) for i in range(n_cols))
+    all_params = ['df']
+    for key, value in params:
```
kozlov-alexey: A blank line, e.g. before a block that generates func text, wouldn't be that bad.
Also n_cols is used only once and probably can be removed at all.

sdc/datatypes/hpat_pandas_dataframe_functions.py
AlexanderKalistratov: Though arguments of df.mean and series.mean are the same, it can't be forwarded straight forward.
1. `axis` for series supports only `index (0)`
2. `numeric_only` is not support by series at all, and shouldn't be forwarded
3.  `level` at the moment we are not storing series in data frame. So there are no hierarchical index and level parameter doesn't make any sense for underlying series, but still there is sense for dataframe. on the other hand at the moment hierarchical index is not support for dataframe either.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mean.html 
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.mean.html

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,46 @@
...
-        raise TypingError("{} 'numeric_only' unsupported. Given: {}".format(_func_name, axis))
+    for key, value in params_s:
+        all_params.append('{}={}'.format(key, value))
+    ap = all_params.copy()
+    ap.pop(0)
+    par = '{}'.format(', '.join(ap))
+    func_definition = 'def _reduce_impl({}):'.format(', '.join(all_params))
+    func_lines = [func_definition]
+    for i, d in enumerate(data_args):
+        line = '  {} = sdc.hiframes.api.init_series(sdc.hiframes.pd_dataframe_ext.get_dataframe_data(df, {}))'
```
```suggestion
        line = '  {} = sdc.hiframes.api.init_series(sdc.hiframes.pd_dataframe_ext.get_dataframe_data(all_params[0], {}))'
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,46 @@
...
+    saved_columns = df.columns
+    data_args = tuple('data{}'.format(i) for i in range(len(saved_columns)))
+    all_params = ['df']
 
-    if not (isinstance(numeric_only, types.Omitted) or numeric_only is False):
-        raise TypingError("{} 'numeric_only' unsupported. Given: {}".format(_func_name, axis))
+    for key, value in params_s:
+        all_params.append('{}={}'.format(key, value))
+    ap = all_params.copy()
+    ap.pop(0)
```
densmirn: I vote for avoiding of using `pop`? Let's separate all parameters and the last one initially.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,46 @@
...
+    par = '{}'.format(', '.join(ap))
+    func_definition = 'def _reduce_impl({}):'.format(', '.join(all_params))
+    func_lines = [func_definition]
+    for i, d in enumerate(data_args):
+        line = '  {} = sdc.hiframes.api.init_series(sdc.hiframes.pd_dataframe_ext.get_dataframe_data(df, {}))'
+        func_lines.append(line.format(d + '_S', i))
+        func_lines.append('  {} = {}.{}({})'.format(d + '_O', d + '_S', name, par))
```
densmirn: `func_lines.append('  {} = {}.{}({})'.format(d + '_O', d + '_S', name, par))` -> `func_lines.append('  {}_O = {}_S.{}({})'.format(d, d, name, par))`

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,46 @@
...
 
-    if not (isinstance(level, types.Omitted) or level is None):
-        raise TypingError("{} 'level' unsupported. Given: {}".format(_func_name, axis))
+    saved_columns = df.columns
+    data_args = tuple('data{}'.format(i) for i in range(len(saved_columns)))
+    all_params = ['df']
 
-    if not (isinstance(numeric_only, types.Omitted) or numeric_only is False):
-        raise TypingError("{} 'numeric_only' unsupported. Given: {}".format(_func_name, axis))
+    for key, value in params_s:
```
kozlov-alexey: Why we use params_s (is it parameters of series function?) to to create df function signature?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,45 @@
...
-
-    _func_name = 'Method pandas.dataframe.count().'
-
-    if not isinstance(self, DataFrameType):
-        raise TypingError('{} The object must be a pandas.dataframe. Given: {}'.format(_func_name, self))
+from sdc.datatypes.hpat_pandas_series_functions import TypeChecker
 
-    if not (isinstance(axis, types.Omitted) or axis == 0):
-        raise TypingError("{} 'axis' unsupported. Given: {}".format(_func_name, axis))
+def sdc_pandas_dataframe_reduce_columns(df, name, params):
```
```suggestion
def sdc_pandas_dataframe_reduce_columns(df, name, series_call_params):
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,45 @@
...
-        raise TypingError("{} 'level' unsupported. Given: {}".format(_func_name, axis))
+    saved_columns = df.columns
+    data_args = tuple('data{}'.format(i) for i in range(len(saved_columns)))
+    all_params = ['df']
 
-    if not (isinstance(numeric_only, types.Omitted) or numeric_only is False):
-        raise TypingError("{} 'numeric_only' unsupported. Given: {}".format(_func_name, axis))
+    for key, value in params:
+        all_params.append('{}={}'.format(key, value))
+    ap = all_params.copy()
```
kozlov-alexey: I'm not sure the copy of all_params is needed. Also it would be good to have a comment here, to indicate that it relies on the fact we use series params to generate signature of DF method, i.e. something like:
```suggestion
    # This relies on parameters part of the signature of Series method called below being the same as for the corresponding DataFrame method
    series_call_params_str = '{}'.format(', '.join(all_params[1:]))
    func_definition = 'def _reduce_impl({}):'.format(', '.join(all_params))
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,45 @@
...
+    par = '{}'.format(', '.join(ap[1:]))
```
AlexanderKalistratov: That's not valid. Params which passed to dataframe method and params which pass to series methods are **different** params. You shouldn't pass `axis` or `numeric_only` params to series.

Also currently you always passing params with default values to series!

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,45 @@
...
+    func_definition = 'def _reduce_impl({}):'.format(', '.join(all_params))
```
AlexanderKalistratov: Please use fstring here and everywhere:
```python
func_definition = f'def _reduce_impl({', '.join(all_params)}):'
```

Or, better:
```python
func_params = ', '.join(all_params)
func_definition = f'def _reduce_impl({func_params)}):'
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -31,65 +31,45 @@
...
-        result_index = []
+    exec(func_text, {'sdc': sdc, 'np': numpy}, loc_vars)
+    _reduce_impl = loc_vars['_reduce_impl']
 
-        for dataframe_item in self._data:
-            item_count = dataframe_item.count()
-            item_name = dataframe_item._name
-            result_data.append(item_count)
-            result_index.append(item_name)
+    return _reduce_impl
```
AlexanderKalistratov: A small refactoring. But now it looks a little bit clearly:
```python
def _dataframe_reduce_columns_codegen(func_name, func_params, series_params, columns):
    result_name_list = []
    func_lines = [f'def {func_name}({func_params}):']
    for c in columns:
        result_d = f'result_{d}'
        func_lines += [f'  series_{d} = init_series(get_dataframe_data({all_params[0]}, {i}))',
                       f'  {result_d} = series_{d}.{func_name}({series_params})'
        result_name_list += result_d

    all_results = ', '.join(result_name_list)
    all_columns = ', '.join([f"'{c}'" for c in columns])

    func_lines += [f'  return pandas.Series([{all_results}], [{all_columns}])']

    global_vars = {'pandas': pandas, 'np': numpy, 
                   'init_series' : init_series, 
                   'get_dataframe_data': get_dataframe_data}

    return func_text, global_vars


def sdc_pandas_dataframe_reduce_columns(df, func_name, params):
    all_params = ['df']

    for key, value in params:
        all_params.append('{}={}'.format(key, value))
    ap = all_params.copy()
    par = '{}'.format(', '.join(ap[1:]))

    df_func_name = f'_df_{func_name}_impl'

    func_text, global_vars = _dataframe_reduce_columns_codegen(df_func_name, all_params, par, df.columns)
    loc_vars = {}

    exec(func_text, global_vars, loc_vars)
    _reduce_impl = loc_vars[df_func_name]

    return _reduce_impl
```

Please note. I haven't test it, so it is most likely full of bugs

I've split codegen part and eval part, so we can tests them separately and also each part become smaller and easier to understand.

sdc/__init__.py
```diff
@@ -60,10 +60,9 @@
     TODO: Needs to detect 'import Pandas' and align initialization according to it
     """
 
-    # sdc.config.numba_compiler_define_nopython_pipeline_orig = \
-    #     numba.compiler.DefaultPassBuilder.define_nopython_pipeline
-    # numba.compiler.DefaultPassBuilder.define_nopython_pipeline = \
-    #     sdc.datatypes.hpat_pandas_dataframe_pass.sdc_nopython_pipeline_lite_register
+    sdc.config.numba_compiler_define_nopython_pipeline_orig = numba.compiler.DefaultPassBuilder.define_nopython_pipeline
```
kozlov-alexey: I believe since @AlexanderKalistratov PR with rewrites this should be commented out. Please pay attention while rebasing.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -40,36 +40,97 @@
...
 from numba.errors import TypingError
+import sdc.datatypes.hpat_pandas_dataframe_types
 
 from sdc.datatypes.hpat_pandas_series_functions import TypeChecker
 
-def sdc_pandas_dataframe_reduce_columns(df, name, params):
 
-    saved_columns = df.columns
-    data_args = tuple('data{}'.format(i) for i in range(len(saved_columns)))
+def _dataframe_reduce_columns_codegen(func_name, func_params, series_params, columns):
```
kozlov-alexey: Can you please add a multi-line comment at the top of this function (not a docstring) with a short example of how func_text will look like (for example for mean and a DF with 1 column)?

AlexanderKalistratov:
@Rubtsowa please fix style issues:
> Hello @Rubtsowa! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`sdc/__init__.py`](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/__init__.py):
> 
> > [Line 63:121](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/__init__.py#L63): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (122 > 120 characters)
> > [Line 64:121](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/__init__.py#L64): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (143 > 120 characters)
> 
> * In the file [`sdc/datatypes/hpat_pandas_dataframe_functions.py`](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/datatypes/hpat_pandas_dataframe_functions.py):
> 
> > [Line 57:1](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/datatypes/hpat_pandas_dataframe_functions.py#L57): [E302](https://duckduckgo.com/?q=pep8%20E302) expected 2 blank lines, found 0
> 
> * In the file [`sdc/hiframes/pd_dataframe_ext.py`](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/hiframes/pd_dataframe_ext.py):
> 
> > [Line 1633:1](https://github.com/IntelPython/sdc/blob/c2a108b91c26ab77b51ccdb9a4126cce38490d45/sdc/hiframes/pd_dataframe_ext.py#L1633): [E402](https://duckduckgo.com/?q=pep8%20E402) module level import not at top of file
> 
> ##### Comment last updated at 2019-12-26 14:32:57 UTC

sdc/tests/test_dataframe.py
```diff
@@ -939,7 +939,7 @@ def test_impl(n):
         n = 11
         pd.testing.assert_series_equal(hpat_func(n), test_impl(n))
 
-    @skip_numba_jit
+    @skip_sdc_jit('SDC pipeline does not support arguments for Series.count()')
```
AlexanderKalistratov: Doesn't feel right that count stops to work with sdc pipeline.

sdc/datatypes/hpat_pandas_dataframe_functions.py
AlexanderKalistratov: I'm still seeing here docstring instead of comment

sdc/hiframes/pd_dataframe_ext.py
```diff
@@ -1628,3 +1628,6 @@ def _impl(df, path_or_buf=None, sep=',', na_rep='', float_format=None,
                       date_format, doublequote, escapechar, decimal)
 
     return _impl
+
+
+from sdc.datatypes.hpat_pandas_dataframe_functions import *
```
AlexanderKalistratov: It shouldn't be here

PokhodenkoSA: Why?

PR #367, opened 2019-12-05
Fix error when perf test results created without drivers.

PokhodenkoSA:
Jira SAT-2205.


sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -264,6 +264,8 @@ class TestResults:
     logger = setup_logging()
 
     def __init__(self, drivers=None):
+        if drivers is None:
+            drivers = []
         self.drivers = drivers
```
densmirn: You could simplify that:
`self.drivers = drivers or []`

PR #372, opened 2019-12-05
Import hooks for SDC implemented via overriding __import__

PokhodenkoSA:
POC for import hooks for SDC:
Import SDC only if both numba and pandas are imported.

### DIFF
--- a/numba_sdc/__init__.py
+++ b/numba_sdc/__init__.py
@@ -1,0 +1,28 @@
+import sys
+import builtins
+
+
+def _init_extension():
+    '''Register Pandas classes and functions with Numba.
+
+    This entry_point is called by Numba when it initializes.
+    '''
+    if 'pandas' in sys.modules:
+        import sdc
+    else:
+        old_import = builtins.__import__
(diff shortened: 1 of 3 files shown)


shssf:
@PokhodenkoSA Could you please move the code into `sdc/__init__`? Or it might be better into `sdc/config.py` because it called before `sdc/__init__.py`

PokhodenkoSA:
> @PokhodenkoSA Could you please move the code into `sdc/__init__`? Or it might be better into `sdc/config.py` because it called before `sdc/__init__.py`

When Numba import extension it imports module and run initialization function in this module. 
E.g. module sdc.config and function _init_extension. When Numba import sdc.config it is mechanics of Python than first sdc is imported and then sdc.config is imported. Then Numba calls sdc.config._init_extension().
It means that before calling sdc.config._init_extension() we have sdc already imported. It will be with any initialization function placed inside sdc.

But our goal is to defer loading of sdc. So I made that initialization function is not placed inside sdc, it is placed inside another package numba_sdc. So sdc is not imported, another module numba_sdc is imported. And importing numba_sdc does not imports sdc inevitable.

PR #375, opened 2019-12-06
Add sdc_overload, sdc_overload_method, sdc_overload_attribute decorators

Hardcode84:
We will need to add `inline='always'` (and possible other attributes) to all our `@overload_method`s after numba `@overload_method` refactoring merge. Add our own attribute wrapper so it can be made from one place.


AlexanderKalistratov:
Could you please do the same for `@overload_attribute` with replacing existing attributes and introduce `sdc_overload` without replacing existing overloads?

PR #382, opened 2019-12-08
Implement Series.str.find()

densmirn:


sdc/datatypes/hpat_pandas_stringmethods_functions.py
```diff
@@ -187,6 +189,65 @@ def hpat_pandas_stringmethods_{methodname}_impl(self{methodparams}):
...
+         returns :obj:`pandas.Series` object
+    """
+
+    ty_checker = TypeChecker('Method find().')
+    ty_checker.check(self, StringMethodsType)
+
+    if not isinstance(sub, (StringLiteral, UnicodeType)):
+        ty_checker.raise_exc(sub, 'str', 'sub')
+
+    accepted_types = (Integer, IntegerLiteral, NoneType, Omitted)
```
AlexanderKalistratov: If I'm not mistaken IntegerLiteral is inherited from Integer, so just `(Integer, NoneType, Omitted)` would be enough. But it is better to check

sdc/datatypes/hpat_pandas_stringmethods_functions.py
```diff
@@ -187,6 +189,65 @@ def hpat_pandas_stringmethods_{methodname}_impl(self{methodparams}):
...
+    if not isinstance(start, accepted_types) and start != 0:
+        ty_checker.raise_exc(start, 'None, int', 'start')
+
+    if not isinstance(end, accepted_types) and end is not None:
+        ty_checker.raise_exc(end, 'None, int', 'end')
+
+    def hpat_pandas_stringmethods_find_impl(self, sub, start=0, end=None):
+        if start != 0:
```
AlexanderKalistratov: Why we are not supporting `start` and `end`? UnicodeString doesn't?

sdc/datatypes/hpat_pandas_stringmethods_functions.py
AlexanderKalistratov: Also I'm not sure if it is better to explicitly add `IntegerLiteral` as supported type or not. On one hand it makes the code more clear, on another, it is confusing that in some places it is specified and in another is not.

sdc/datatypes/hpat_pandas_stringmethods_functions.py
densmirn: Numba 0.46 doesn't support str.find with extra parameters.

sdc/datatypes/hpat_pandas_stringmethods_functions.py
densmirn: You are right. `IntegerLiteral` is inherited of `Integer`.

PR #389, opened 2019-12-09
Fix series index and names issues pt.1

AlexanderKalistratov:
Fixing issues with index and names for Series and extending tests


sdc/tests/test_series.py
```diff
@@ -2977,6 +3011,51 @@ def test_impl(S):
         S = pd.Series([np.nan, 2., 3., np.inf])
         pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
 
+    def test_series_isna_index(self):
```
densmirn: You could combine the test with `test_series_isna_str_index` through the loop.

sdc/tests/test_series.py
```diff
@@ -2977,6 +3011,51 @@ def test_impl(S):
...
+        def test_impl(S):
+            return S.isna()
+        hpat_func = self.jit(test_impl)
+
+        # column with NA
+        S = pd.Series(data=[np.nan, 2., 3., np.inf], index=[3, 2, 1, 0])
+        pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
+
+    def test_series_isna_named(self):
```
densmirn: You could combine the test with `test_series_isna_str_named` through the loop.

AlexanderKalistratov:
@densmirn @kozlov-alexey @PokhodenkoSA please pay attention to the new `sdc_limitation` decorator. It uses `expectedFailure` and is used to mark tests which behavior is different from pandas due to numba or sdc limitation.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -903,21 +903,46 @@ def hpat_pandas_series_shift(self, periods=1, freq=None, axis=0, fill_value=None
...
+    fill_is_default = isinstance(fill_value, (types.Omitted, types.NoneType)) or fill_value is None
+    series_np_dtype = [numpy_support.as_dtype(self.data.dtype)]
+    fill_np_dtype = [numpy.float64 if fill_is_default else numpy_support.as_dtype(fill_value)]
+
+    common_dtype = numpy.find_common_type(series_np_dtype, fill_np_dtype)
+
+    if fill_is_default:
+        def hpat_pandas_series_shift_impl(self, periods=1, freq=None, axis=0, fill_value=None):
+            if axis != 0:
+                raise TypingError('Method shift(). Unsupported parameters. Given axis != 0')
```
kozlov-alexey: I think TypingError should be changed to ValueError (or whatever runtime exception Pandas raises when it gets wrong axis). I would use TypingError during Typing only.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1071,23 +1096,27 @@ def hpat_pandas_series_copy(self, deep=True):
...
+                return pandas.Series(data=self._data.copy(), name=self._name)
+            else:
+                return pandas.Series(data=self._data, name=self._name)
+        return hpat_pandas_series_copy_impl
+    else:
+        def hpat_pandas_series_copy_impl(self, deep=True):
+            if deep:
+                return pandas.Series(data=self._data.copy(), index=self._index.copy(), name=self._name)
+            else:
+                return pandas.Series(data=self._data, index=self._index.copy(), name=self._name)
```
kozlov-alexey: Why do we copy index here with deep=False? Is it only because we have no immutable indexes now?
Anyway, can we add a comment here for clarity.

sdc/tests/test_series.py
```diff
@@ -293,7 +294,7 @@ def test_impl():
 
     def test_create_series_index2(self):
```
kozlov-alexey: We might add docstrings for these tests to clarify what they are testing and what's the difference between them.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Oh, I see in the test below: "# Shallow copy of index is not supported yet". But still a comment here would be useful.

sdc/tests/test_series.py
```diff
@@ -3867,13 +4067,53 @@ def pyfunc(series, periods, freq, axis, fill_value):
...
+        pd.testing.assert_series_equal(cfunc(), ref_result)
+
+    def test_series_shift_0period_sdc(self):
+        def pyfunc():
+            series = pd.Series([6, 4, 3])
+            return series.shift(periods=0)
+
+        cfunc = self.jit(pyfunc)
+        ref_result = pyfunc()
+        pd.testing.assert_series_equal(cfunc(), ref_result.astype(np.float64))
```
kozlov-alexey: Not sure if that's really better, but I think it's good to have a comment in tests where we attempt to modify ref_result in some way or use less strict checks.
```suggestion
        # check_dtype=False due to SDC impl returns fixed dtype - float64
        pd.testing.assert_series_equal(cfunc(), ref_result, check_dtype=False)
```

sdc/tests/test_series.py
```diff
@@ -3867,13 +4067,53 @@ def pyfunc(series, periods, freq, axis, fill_value):
...
+        for data in datas:
+            for periods in [1, 2, -1]:
+                for fill_value in [-1, 0, 9.1, np.nan, -3.3, None]:
+                    with self.subTest(data=data, periods=periods, fill_value=fill_value):
+                        series = pd.Series(data)
+                        jit_result = cfunc(series, periods, freq, axis, fill_value)
+                        ref_result = pyfunc(series, periods, freq, axis, fill_value)
+                        pd.testing.assert_series_equal(jit_result, ref_result)
+
+    @sdc_limitation
```
kozlov-alexey: Can you please add a docstring here describing what the limitation is?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -903,21 +903,46 @@ def hpat_pandas_series_shift(self, periods=1, freq=None, axis=0, fill_value=None
     if not isinstance(axis, (types.Omitted, int, types.Integer)):
         raise TypingError('{} Unsupported parameters. Given axis: {}'.format(_func_name, axis))
 
+    fill_is_default = isinstance(fill_value, (types.Omitted, types.NoneType)) or fill_value is None
+    series_np_dtype = [numpy_support.as_dtype(self.data.dtype)]
+    fill_np_dtype = [numpy.float64 if fill_is_default else numpy_support.as_dtype(fill_value)]
+
+    common_dtype = numpy.find_common_type(series_np_dtype, fill_np_dtype)
```
kozlov-alexey: Does this work? I thought, numpy dtype has to be converted to Numba type instance via
`numpy_support.from_dtype(common_dtype)` to be used inside jitted function body.

sdc/tests/test_series.py
AlexanderKalistratov: Done

sdc/tests/test_series.py
AlexanderKalistratov: done

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: TypingError doesn't rise and cannot be caught in tests for unknown reason

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: done

sdc/tests/test_series.py
AlexanderKalistratov: added

sdc/tests/test_series.py
AlexanderKalistratov: done

sdc/tests/test_series.py
AlexanderKalistratov: added

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: done

PR #392, opened 2019-12-09
Implement Series.str.ljust()/rjust()

densmirn:


densmirn:
> I'm not sure if we really should create new file for each sample. Probably similar functionality into single file? Or create folder which unites all `str` methods?

We should have separate documentation for each functionality. So I think we need to have separated examples, but I agree to put the examples in separate folders.

PR #394, opened 2019-12-09
Add setitem

Rubtsowa:


sdc/tests/test_series.py
```diff
@@ -975,6 +975,7 @@ def test_impl(A):
         pd.testing.assert_series_equal(A1, A2)
 
     @skip_numba_jit
+    @skip_sdc_jit
```
densmirn: Can we get the test passed instead of skipping this one?

sdc/tests/test_series.py
```diff
@@ -1028,6 +1030,7 @@ def test_impl(A):
         self.assertEqual(hpat_func(S), test_impl(S))
 
     @skip_numba_jit
+    @skip_sdc_jit
```
densmirn: The same. It's not good idea to skip the tests.

examples/series_setitem.py
```diff
@@ -0,0 +1,48 @@
...
+def series_setitem():
+    numb = 0
+    series = pd.Series(np.arange(10, 0, -1))  # Series of 10, 9, ..., 1
+
+    # series[0] = numb
+    #
+    # series[2:6] = numb
+    #
+    # indices = pd.Series(np.asarray([1, 6, 7, 8, 9]))
+    # series[indices] = numb
```
densmirn: I would propose to add 3 examples: for num index, for slice, for series index. Then insert all the examples to docstring.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -145,32 +145,65 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+    -------
+    :class:`pandas.Series` or an element of the underneath type
+            object of :class:`pandas.Series`
+    """
 
     _func_name = 'Operator setitem().'
-    if not isinstance(self, SeriesType):
-        raise TypingError('{} The object must be a pandas.series. Given: {}'.format(_func_name, self))
+
+    ty_checker = TypeChecker('Operator setitem().')
```
densmirn: Is it correct to use brackets for `setitem()`?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -142,6 +142,95 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+            """
+            Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_setitem_for_series
+            """
+            super_index = idx._data
+            self._data[super_index] = value
+            return self
+
+        return hpat_pandas_series_setitem_idx_series_impl
+
+    raise TypingError('{} The index must be an Integer, Slice or a pandas.series. Given: {}'.format(_func_name, idx))
```
densmirn: Moreover I would propose to use `TypeCheker` here.

examples/series_setitem.py
densmirn: Additionally let's initialize Series with 5 elements, I think 10 is too many, because the result takes much space in the docstring,

sdc/tests/test_series.py
```diff
@@ -998,6 +998,7 @@ def test_impl(A):
         A = pd.Series(np.arange(n))
         self.assertEqual(hpat_func(A), test_impl(A))
 
+
```
densmirn: excess empty line

sdc/datatypes/hpat_pandas_series_functions.py
Rubtsowa: This quite correctly. Operator getitem also uses.

sdc/tests/test_series.py
```diff
@@ -4684,6 +4685,68 @@ def test_series_pct_change_impl(S, periods=1, fill_method='pad', limit=None, fre
...
+        pd.testing.assert_series_equal(result_ref, result)
+
+    def test_series_setitem_for_slice(self):
+        def test_impl(S, val):
+            S[2:] = val
+            return S
+
+        hpat_func = self.jit(test_impl)
+        S1 = pd.Series([0, 1, 2, 3, 4])
+        S2 = pd.Series([0, 1, 2, 3, 4])
```
densmirn: I propose to use method `copy()` here, because sometime somebody can change `S1` and forget to change `S2`.
`S2 = S1.copy()`

examples/series_setitem_slice.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+import pandas as pd
+
+from numba import njit
+
+
+@njit
+def series_setitem():
+    numb = 0
+    series = pd.Series(np.arange(5, 0, -1))  # Series of 10, 9, ..., 1
```
densmirn: Looks like we should correct comment according to the code line.

examples/series_setitem_series.py
```diff
@@ -0,0 +1,44 @@
...
+import numpy as np
+import pandas as pd
+
+from numba import njit
+
+
+@njit
+def series_setitem():
+    numb = 0
+    series = pd.Series(np.arange(5, 0, -1))  # Series of 10, 9, ..., 1
```
densmirn: Looks like we should correct comment according to the code line.

examples/series_setitem_int.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@njit
+def series_setitem():
+    numb = 0
+    series = pd.Series(np.arange(5, 0, -1))  # Series of 5, 4, 3, 2, 1
+
+    series[0] = numb
+
+    return series
```
densmirn: Let's specify the result as comment.

examples/series_setitem_series.py
```diff
@@ -0,0 +1,44 @@
...
+
+    indices = pd.Series(np.asarray([1, 3]))
+    series[indices] = numb
+
+    return series
```
densmirn: Let's specify the result as comment as it was done in other examples.

examples/series_setitem_slice.py
```diff
@@ -0,0 +1,43 @@
...
+
+    series[2:5] = numb
+
+    return series
```
densmirn: Let's specify the result as comment to let opportunity an user to reproduce that result.

sdc/datatypes/hpat_pandas_series_functions.py
densmirn: Is there other reasons that it is correct?

sdc/tests/test_series.py
densmirn: Maybe you need to use parameter `deep`: `S2 = S1.copy(deep=True)`

sdc/tests/test_series.py
```diff
@@ -4737,12 +4738,12 @@ def test_impl(S, ind, val):
 
         with self.assertRaises(TypingError) as raises:
             hpat_func(S, ind1, value1)
-        msg = 'TypingError: Operator setitem(). The object value'
+        msg = 'TypingError: Operator setitem. The object value'
```
densmirn: I think `TypingError: ` is excess in the message. Moreover let's provide full message (what's given, what's expected).

sdc/tests/test_series.py
```diff
@@ -4737,12 +4738,12 @@ def test_impl(S, ind, val):
...
         self.assertIn(msg, str(raises.exception))
 
         with self.assertRaises(TypingError) as raises:
             hpat_func(S, ind2, value2)
-        msg = 'Operator setitem(). The object idx'
+        msg = 'Operator setitem. The object idx'
```
densmirn: Let's provide full message (what's given, what's expected).

sdc/tests/test_series.py
Rubtsowa: This problem, because message will be not in 1 string,  this message create error.

sdc/tests/test_series.py
```diff
@@ -4779,7 +4779,8 @@ def test_impl(S, ind, val):
 
         with self.assertRaises(TypingError) as raises:
             hpat_func(S, ind2, value2)
-        msg = 'Operator setitem. The object idx' + '\n' + ' given: unicode_type' + '\n' + ' expected: int, Slice, Series'
+        msg = 'Operator setitem. The object idx' + '\n' + ' given: unicode_type' + '\n' \
```
densmirn: Why not avoid concatenation of string via `+`?
`msg = 'Operator setitem. The object idx\n given: unicode_type\n expected: int, Slice, Series'`

sdc/tests/test_series.py
```diff
@@ -4720,6 +4720,70 @@ def test_series_pct_change_impl(S, periods=1, fill_method='pad', limit=None, fre
...
+        hpat_func = self.jit(test_impl)
+        S = pd.Series([0, 1, 2, 3, 4, 5])
+        ind1 = 5
+        ind2 = '3'
+        value1 = 'ababa'
+        value2 = 101
+
+        with self.assertRaises(TypingError) as raises:
+            hpat_func(S, ind1, value1)
+        msg = 'Operator setitem. The object value' + '\n' + ' given: unicode_type' + '\n' + ' expected: self'
```
densmirn: Why not avoid concatenation of string via +?

sdc/tests/test_series.py
```diff
@@ -4774,13 +4774,12 @@ def test_impl(S, ind, val):
 
         with self.assertRaises(TypingError) as raises:
             hpat_func(S, ind1, value1)
-        msg = 'Operator setitem. The object value' + '\n' + ' given: unicode_type' + '\n' + ' expected: self'
+        msg = 'Operator setitem. The object value\n given: unicode_type\n expected: self'
```
densmirn: Let's create message template for code reusage:
```
msg_tmpl = 'Operator setitem. The object {}\n given: unicode_type\n expected: {}'
msg = msg_tmpl.format('value', 'self')
# ...
msg = msg_tmpl.format('idx', 'int, Slice, Series')
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -143,6 +143,102 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+    -------
+    :class:`pandas.Series` or an element of the underneath type
+            object of :class:`pandas.Series`
+    """
+    ty_checker = TypeChecker('Operator setitem.')
+    ty_checker.check(self, SeriesType)
+
+    if (isinstance(value, SeriesType) and not isinstance(self.dtype, value.dtype)) or \
+            not isinstance(self.dtype, type(value)):
+        ty_checker.raise_exc(value, 'self', 'value')
```
densmirn: What do you mean `'self'`? What does mean `self expected`?

examples/series_setitem_int.py
```diff
@@ -0,0 +1,43 @@
...
+
+from numba import njit
+
+
+@njit
+def series_setitem():
+    numb = 0
+    series = pd.Series(np.arange(5, 0, -1))  # Series of 5, 4, 3, 2, 1
+
+    series[0] = numb
```
kozlov-alexey: I think it'd be better to rename 'numb' to 'value' (since the former is a word itself)

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -142,6 +142,102 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+            2    3
+            3    0
+            4    1
+            dtype: int64
+
+    Intel Scalable Dataframe Compiler Developer Guide
+    *************************************************
+     Pandas Series operator :attr:`pandas.Series.set` implementation
+
+    Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_setitem_unsupported
```
kozlov-alexey: Please use a correct line to run the tests for setitem, i.e.
`python -m sdc.runtests -k sdc.tests.test_series.TestSeries.test_series_setitem*`

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -142,6 +142,102 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+
+    Parameters
+    ----------
+    series: :obj:`pandas.Series`
+           input series
```
kozlov-alexey: Please use the same indent

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -142,6 +142,102 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+    :class:`pandas.Series` or an element of the underneath type
+            object of :class:`pandas.Series`
+    """
+    ty_checker = TypeChecker('Operator setitem.')
+    ty_checker.check(self, SeriesType)
+
+    if (isinstance(value, SeriesType) and not isinstance(self.dtype, value.dtype)) or \
+            not isinstance(self.dtype, type(value)):
+        ty_checker.raise_exc(value, self.dtype, 'value')
+
```
kozlov-alexey: It would be easier to understand the options if value is checked against some type:
```suggestion
    if not ((isinstance(value, SeriesType) and isinstance(value.dtype, self.dtype))
               or isinstance(value, type(self.dtype)):
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -142,6 +142,102 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+            """
+            Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_setitem_for_series
+            """
+            super_index = idx._data
+            self._data[super_index] = value
+            return self
+
+        return hpat_pandas_series_setitem_idx_series_impl
+
+    ty_checker.raise_exc(idx, 'int, Slice, Series', 'idx')
```
kozlov-alexey: I think we should simply return None here, as this exception will not reach the end user anyway.

sdc/tests/test_series.py
```diff
@@ -4782,6 +4782,69 @@ def test_series_pct_change_impl(S, periods=1, fill_method='pad', limit=None, fre
         msg = 'Method pct_change(). The object periods'
         self.assertIn(msg, str(raises.exception))
 
+    def test_series_setitem_for_value(self):
```
kozlov-alexey: It seems that all three positive tests only cover integer series case, so floats and strings are not covered.
It would be good to add them too, you can write an aggregated test for diff data types/values, like here:
https://github.com/IntelPython/sdc/blob/abfa12bb622f3ca248c2513479487895b0e2d769/sdc/tests/test_series.py#L2176

sdc/datatypes/hpat_pandas_series_functions.py
Rubtsowa: I'm sorry, but this another statement. If to paint this stement, we get  

if (not isinstance(value, SeriesType) or not isinstance(value.dtype, self.dtype)) and not isinstance(value, type(self.dtype)):

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -141,6 +141,104 @@ def hpat_pandas_series_getitem_idx_series_impl(self, idx):
...
+           isinstance(value, type(self.dtype))):
+        ty_checker.raise_exc(value, self.dtype, 'value')
+
+    if isinstance(idx, types.Integer) or isinstance(idx, types.SliceType):
+        def hpat_pandas_series_setitem_idx_integer_impl(self, idx, value):
+            """
+            Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_setitem_for_value
+            Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_setitem_for_slice
+            """
+            self._data[idx] = value
```
kozlov-alexey: OK, I think I now why it doesn't work for StringArrays - operator.setitem for StringArray expects the size of 'value' to be equal to the value of _data[idx]. Hence, you'll need special handling for string data.

PR #395, opened 2019-12-09
Fix benchmark results collecting

PokhodenkoSA:
This PR also closes #393.
This PR make TestBase as base class for all performance tests classes. It enables result collection for all classes.

### DIFF
--- a/sdc/tests/tests_perf/test_perf_base.py
+++ b/sdc/tests/tests_perf/test_perf_base.py
@@ -6,19 +6,26 @@ from sdc.tests.tests_perf.test_perf_utils import *
 
 class TestBase(unittest.TestCase):
     iter_number = 5
+    test_results_class = TestResults
 
     @classmethod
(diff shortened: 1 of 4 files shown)


sdc/tests/tests_perf/test_perf_base.py
```diff
@@ -6,19 +6,26 @@
 
 class TestBase(unittest.TestCase):
     iter_number = 5
+    test_results_class = TestResults
 
     @classmethod
-    def setUpClass(cls):
+    def create_test_results(cls):
```
densmirn: Maybe does it make sense to move the class method to `TestResults` class as class method?

PokhodenkoSA: This method creates drivers and loads prev results. I think it should be decided outside the result class. So TestBase is good place for deciding it. I would like it method to stay here.

PR #408, opened 2019-12-11
Modifications of setup.py to allow building User Guide

samaid:
Allows building documentation using
::
    python setup.py build_user_doc


samaid:
Build failed because Sphinx is not present in the build environment. Need to rework setup.py so that it does not expect dependency on Sphinx while building SDC

PR #410, opened 2019-12-12
Extends setup.py with build_doc command

samaid:
1. Tested no Sphinx env
>python setup.py build_doc - failed with correct exception message

2. Tested Sphinx env
>python setup.py build_doc - completed HTML doc build

3. Tested no Sphinx env
>python setup.py install - no import Sphinx error message


setup.py
```diff
@@ -202,6 +180,9 @@ def readme():
 
 str_libs = np_compile_args['libraries']
 
+if not is_win:
```
Hardcode84: Is this really needed?

samaid: 1. It is definitely not needed for building documentation. This line (and a lot of similar lines) were inherited from existing setup.py
2. After reviewing all platform specifics (is_win flag) I can conclude that it comes from the need to link third party dependencies that I would like to get rid of by the release:
* MPI 
* Boost (filesystem, system, regex)
* OpenCV

@Vyacheslav-Smirnov, comments?

Vyacheslav-Smirnov: Yes, MPI and Boost are used
Opencv looks like is linked if installed on the system, I do not know if it is really used in SDC
In the conda recipe there is not dependency on opencv

samaid: We do not need to carry on these dependencies. No MPI, no Boost, no OpenCV.

docs/source/buildscripts/sdc_build_doc.py
```diff
@@ -0,0 +1,103 @@
...
+                sys.path.remove('.')
+                removed_flag = True
+
+            if './' in sys.path:
+                sys.path.remove('./')
+                removed_flag = True
+
+            if cwd in sys.path:
+                sys.path.remove(cwd)
+                removed_flag = True
```
densmirn: You could do that in the loop:
```
for pth in ['', '.', './'. cwd]:
    sys.path.remove(pth)
    removed_flag = True
```

setup.py
```diff
@@ -394,11 +364,19 @@ def run(self):
...
 
+# Custom build commands
+#
+# These commands extend standard setuptools build procedure
+#
+sdc_build_commands = versioneer.get_cmdclass()
+sdc_build_commands['build_doc'] = SDCBuildDoc
 sdc_build_commands.update({'style': style})
+sdc_version = versioneer.get_version()
+sdc_release = 'Alpha (' + versioneer.get_version() + ')'
```
densmirn: You could use f-strings in such cases:
`'Alpha (' + versioneer.get_version() + ')'` -> `f'Alpha ({versioneer.get_version()})'` or `'Alpha ({})'.format(versioneer.get_version())`

docs/source/buildscripts/sdc_build_doc.py
PokhodenkoSA: Correct replacement:
```
for pth in ['', '.', './'. cwd]:
    if pth in sys.path:
        sys.path.remove(pth)
        removed_flag = True
```
`sys.path.remove(x)` raises `ValueError: list.remove(x): x not in list`

setup.py
```diff
@@ -27,8 +28,8 @@
 from setuptools import setup, Extension, find_packages, Command
 import platform
 import os
-from distutils.command import build
-from distutils.spawn import spawn
+import sys
```
PokhodenkoSA: I do not see where this `sys` is used in this file.

PokhodenkoSA:
This PR requires install `sdc` with `pip install .` or `pip install sdc`. 
Installing with `pip install -e .` (for development) will not work because `cwd` is removed from `sys.path`. For developers it is more convenient to use `pip install -e .` and run `cd docs && make html`.

Vyacheslav-Smirnov:
I will continue work on this PR

setup.py
```diff
@@ -394,11 +363,19 @@ def run(self):
...
 
+# Custom build commands
+#
+# These commands extend standard setuptools build procedure
+#
+sdc_build_commands = versioneer.get_cmdclass()
+sdc_build_commands['build_doc'] = SDCBuildDoc
 sdc_build_commands.update({'style': style})
+sdc_version = versioneer.get_version()
+sdc_release = 'Alpha (' + versioneer.get_version() + ')'
```
PokhodenkoSA:
```suggestion
sdc_release = 'Alpha ({})'.format(versioneer.get_version())
```

PR #414, opened 2019-12-15
Add series.rolling() and rolling.min()

densmirn:


sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+from sdc.datatypes.hpat_pandas_series_rolling_types import SeriesRollingType
+from sdc.utils import sdc_overload_method
+
+
+@register_jitable
+def arr_finite_min(arr):
+    """Calculate minimum of finite values"""
+    finite_arr = arr[numpy.isfinite(arr)]
+    if len(finite_arr) == 0:
+        return numpy.nan
```
AlexanderKalistratov: It doesn't looks correct to me. Have you test it on data where you have numbers and +- `inf`? without `nan`? Like this: [1, 2, inf], [1, 2, -inf], [inf, -inf]?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for i in range(min(win, length)):
+            arr_range = input_arr[:i + 1]
+            finite_values_num = numpy.isfinite(arr_range).sum()
+            if finite_values_num < minp:
+                output_arr[i] = numpy.nan
+            else:
+                output_arr[i] = rolling_func(arr_range)
+
+        for i in range(min(win, length), length):
```
AlexanderKalistratov: prange

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+        win = self._window
+        minp = self._min_periods
+
+        input_series = self._data
+        input_arr = input_series._data
+        length = len(input_arr)
+        out_type = input_arr.dtype if nan_out_type == True else output_type # noqa
+        output_arr = numpy.empty(length, dtype=out_type)
+
+        for i in range(min(win, length)):
```
AlexanderKalistratov: I'm not sure. If `win` is big enough it is better to put here `prange`

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+            arr_range = input_arr[:i + 1]
+            finite_values_num = numpy.isfinite(arr_range).sum()
+            if finite_values_num < minp:
```
AlexanderKalistratov: Why do we need this check? Wouldn't `arr_finite_min` check finitness?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+                output_arr[i] = numpy.nan
+            else:
+                output_arr[i] = rolling_func(arr_range)
+
+        for i in range(min(win, length), length):
+            arr_range = input_arr[i + 1 - win:i + 1]
+            finite_values_num = numpy.isfinite(arr_range).sum()
+            if finite_values_num < minp:
```
AlexanderKalistratov: Same question as above - do we need this check?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+            else:
+                output_arr[i] = rolling_func(arr_range)
+
+        return pandas.Series(output_arr, input_series._index, name=input_series._name)
+
+    return impl
+
+
+hpat_pandas_rolling_series_min_impl = register_jitable(
+    gen_hpat_pandas_series_rolling_impl(arr_finite_min, float64))
```
AlexanderKalistratov: Looks very interesting. Do you think we can use something like that to generate `DataFrame` methods? All this `min`, `max`, `mean`...

AlexanderKalistratov: Does jittible inlines?

sdc/datatypes/hpat_pandas_series_rolling_types.py
```diff
@@ -0,0 +1,104 @@
...
+# *****************************************************************************
+
+from numba import cgutils, types
+from numba.datamodel import StructModel
+from numba.extending import (intrinsic, make_attribute_wrapper,
+                             models, register_model)
+from numba.typing.templates import signature
+
+
+class SeriesRollingType(types.Type):
```
AlexanderKalistratov: I wonder if `DataFrame` rolling really should be a different type

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: I checked it on such input: `[1., np.inf, np.inf, -1., 0., np.inf, np.NINF, np.NINF]` that was passed.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: Done!

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: Done!

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: According to the documentation:
```
min_periods : int, default None
    Minimum number of observations in window required to have a value (otherwise result is NA).
```
we need to count number of finite values to decide how to get a result (call a function or just take a `nan`). This function is generic, so the function shouldn't depend on called method. As I saw `series.rolling.min` calculate minimum using finite values.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: According the documentation on `rolling`, I think so.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: I would say we can, at the least can try. What do you mean "Does jittible inlines?"?

sdc/datatypes/hpat_pandas_series_rolling_types.py
densmirn: Do you mean we need to create common type, e.g. `RollingType`? Perhaps you are right. Let me do that.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
AlexanderKalistratov: This check could be just
```suggestion
            if i <= (minp - 1):
```
You don't need to check for finitness

sdc/datatypes/hpat_pandas_series_rolling_types.py
densmirn: We somehow should separate `SeriesRolling` and `DataFrameRolling` types to be able to overload  their methods separately.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: Unit test `test_series_rolling_min` fails with this approach.

sdc/datatypes/hpat_pandas_series_rolling_types.py
AlexanderKalistratov: But we know data type of member `data` (either series or dataframe). Isn't it enough?
Let's discuss it personally

densmirn: Do you want I make some related changes in this PR?

AlexanderKalistratov: No, i think it's ok for now

sdc/datatypes/hpat_pandas_series_rolling_functions.py
AlexanderKalistratov: But why it fails? Why do we need this extra check for finitness here?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -0,0 +1,145 @@
...
+from sdc.datatypes.common_functions import TypeChecker
+from sdc.datatypes.hpat_pandas_series_rolling_types import SeriesRollingType
+from sdc.utils import sdc_overload_method
+
+
+@register_jitable
+def arr_finite_min(arr):
+    """Calculate minimum of finite values"""
+    finite_arr = arr[numpy.isfinite(arr)]
+    if len(finite_arr) == 0:
```
AlexanderKalistratov: What are we actually checking here? Shouldn't it be something like:
```python
if len(finite_arr) != len(arr):
```

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: As I understand "required to have a value" from the docs means "required to be a finite". So I count number of finite values to compare it with `min_periods` and in further decide how to get the result.

AlexanderKalistratov:
@densmirn it looks like I'm incorecctly understands meaning of parameter `min_periods`. Let's discuss it personally

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: We check if array is empty, then the minimum should be a `nan`. `series.rolling.min` works this way.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
AlexanderKalistratov: Most likely you are right. Still, I don't like the idea to check for finitness twice.

densmirn: Me too, but when we will add another methods which will be called inside of the generic rolling function, e.g. `sum()`, checking of finitness will be happened once (of course if the method `sum` doesn't require to check finitness :) ). May be we can discuss it offline :)

densmirn: I mean it's ok to check finitness in this case twice, because firstly we check it because of `min_periods` requires it, secondly we check it because the method `min` requires that.

PR #416, opened 2019-12-15
Add SDC examples execution in public CI

Vyacheslav-Smirnov:
- New option test-mode=exmples is added to test.py and used


buildscripts/test.py
```diff
@@ -196,6 +199,56 @@
 
         format_print(f'Tests for {package_type} packages are PASSED')
 
+    # Execute examples
+    if test_mode == 'examples':
+        format_print(f'Run SDC examples')
```
densmirn: You don't need f-string here.

buildscripts/test.py
```diff
@@ -196,6 +199,56 @@
...
+        sdc_examples = os.path.join(sdc_src, 'examples')
```
densmirn: I would advise to use `pathlib` module instead of `os` as a newer approach.

buildscripts/test.py
Vyacheslav-Smirnov: Ok, will remove.

buildscripts/test.py
Vyacheslav-Smirnov: I would like to leave os.path.join here due to changing this to pathlib require to make changes in all other places.
I think this should be done in separate PR.

densmirn: I'm not sure it will be done in separate PR. It was like a recommendation, so you we can skip it in this PR.

PR #419, opened 2019-12-17
Overload Series.rolling.count()

densmirn:


sdc/tests/test_rolling.py
```diff
@@ -486,43 +498,24 @@ def test_impl(series, window, min_periods, center,
...
-
-            # TODO: fix the issue when window = 0
-            for window in range(1, len(series) + 2):
-                for min_periods in range(window + 1):
-                    ref_result = test_impl(series, window, min_periods)
-                    jit_result = hpat_func(series, window, min_periods)
-                    pd.testing.assert_series_equal(ref_result, jit_result)
+            all_data = test_global_input_data_float64
+            indices = [list(range(len(data)))[::-1] for data in all_data]
+            for data, index in zip(all_data, indices):
```
AlexanderKalistratov: Please use `subTest` for this cycle. So, in case test would break we would know with which parameters

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -36,6 +36,12 @@
 from sdc.utils import sdc_overload_method
 
 
+@register_jitable
+def arr_nonnan_count(arr):
+    """Count non-NaN values"""
+    return len(arr) - numpy.isnan(arr).sum()
```
AlexanderKalistratov: Does it correctly handles `inf`?

sdc/tests/test_rolling.py
```diff
@@ -486,43 +498,24 @@ def test_impl(series, window, min_periods, center,
...
+                series = pd.Series(data, index, name='A')
+
+                # TODO: fix the issue when window = 0
+                for window in range(1, len(series) + 2):
+                    for min_periods in range(window + 1):
```
AlexanderKalistratov: Or, even put subTest here

AlexanderKalistratov:
I'm not sure it is a good idea to have single test which tests everything

AlexanderKalistratov:
@kozlov-alexey what do you thing about combining different tests?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
densmirn: Yes, it does. Test is passed.

densmirn:
> I'm not sure it is a good idea to have single test which tests everything

Let'm me split the test.

sdc/tests/test_rolling.py
```diff
@@ -486,6 +486,24 @@ def test_impl(series, window, min_periods, center,
...
+
+        hpat_func = self.jit(test_impl)
+
+        all_data = test_global_input_data_float64
+        indices = [list(range(len(data)))[::-1] for data in all_data]
+        for data, index in zip(all_data, indices):
+            series = pd.Series(data, index, name='A')
+            for window in range(len(series) + 2):
+                for min_periods in range(window + 1):
+                    with self.subTest(window=window, min_periods=min_periods):
```
kozlov-alexey: It might be useful to pass data and index into subTest too, to see the Series data on which the tests fails.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -89,12 +95,101 @@ def impl(self):
...
+    def impl(self):
+        win = self._window
+
+        input_series = self._data
+        input_arr = input_series._data
+        length = len(input_arr)
+        out_type = input_arr.dtype if nan_out_type == True else output_type  # noqa
+        output_arr = numpy.empty(length, dtype=out_type)
+
+        for i in prange(min(win, length)):
```
kozlov-alexey: A comment here regarding the size of the part of the array in window would be useful. Also we might call min(win, length) somehow.

AlexanderKalistratov: @kozlov-alexey not sure that you exactly mean by
>Also we might call min(win, length) somehow.

?

And by
>A comment here regarding the size of the part of the array in window would be useful.

too...

kozlov-alexey: @AlexanderKalistratov  I just suggested to use a new variable storing min(win, length), since it's used twice, but it's already done as I see.

sdc/tests/test_rolling.py
```diff
@@ -497,13 +516,13 @@ def test_impl(series, window, min_periods):
...
         for data, index in zip(all_data, indices):
             series = pd.Series(data, index, name='A')
-
             # TODO: fix the issue when window = 0
             for window in range(1, len(series) + 2):
                 for min_periods in range(window + 1):
-                    ref_result = test_impl(series, window, min_periods)
-                    jit_result = hpat_func(series, window, min_periods)
-                    pd.testing.assert_series_equal(ref_result, jit_result)
+                    with self.subTest(window=window, min_periods=min_periods):
```
AlexanderKalistratov: But what about passing series to subTest here?

sdc/tests/test_rolling.py
```diff
@@ -516,13 +535,13 @@ def test_impl(series, window, min_periods):
...
```
AlexanderKalistratov: And here?

PR #420, opened 2019-12-17
Helps to see which skipped tests are passing now

PokhodenkoSA:
Uncomment this lines to see which skipped tests are passing now.
You will see that skipped tests are now have status `expected failure` but some tests may became `unexpected success`.
You can unskip `unexpected success` tests because they are passing now.


AlexanderKalistratov:
@shssf please pay attention to this PR. Is it clear how it would work?

Hardcode84:
Maybe better to use env var instead of commenting/uncommenting code?

PokhodenkoSA:
> Maybe better to use env var instead of commenting/uncommenting code?

1. @Vyacheslav-Smirnov will you use it in CI?
2. SDC_TEST_EXPECTED_FAILURE - is it good enough name?

PR #421, opened 2019-12-17
Overload series.rolling.sum()

densmirn:


sdc/tests/test_rolling.py
```diff
@@ -545,6 +545,29 @@ def test_impl(series, window, min_periods):
...
+        indices = [list(range(len(data)))[::-1] for data in all_data]
+        for data, index in zip(all_data, indices):
+            series = pd.Series(data, index, name='A')
+            for window in range(0, len(series) + 3, 2):
+                for min_periods in range(0, window + 1, 2):
+                    with self.subTest(series=series, window=window,
+                                      min_periods=min_periods):
+                        ref_result = test_impl(series, window, min_periods)
+                        jit_result = hpat_func(series, window, min_periods)
+                        pd.testing.assert_series_equal(ref_result, jit_result)
```
kozlov-alexey: I think we should keep jitted result first and result_ref second for consistency to other test files.

examples/series/rolling/series_rolling_sum.py
```diff
@@ -0,0 +1,39 @@
...
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_rolling_sum():
+    series = pd.Series([4, 3, 5, 2, 6])  # Series of 4, 3, 5, 2, 6
+    out_series = series.rolling(3).sum()
+
+    return out_series  # Expect series of NaN, NaN, 12.0, 10.0, 13.0
```
kozlov-alexey: Maybe better would be to say "Expect the series with the sum of last 3 elements of input series on each position"?

densmirn: As I caught we shouldn't describe how we get the result here (it should be described in the documentation), here we should specify a result of the call which can be reproducible by an user. Let me keep the comment unchanged? :)

sdc/tests/test_rolling.py
densmirn: I agree with that, let me make such change.

PR #430, opened 2019-12-20
Fix lnk to docs in README.rst

PokhodenkoSA:


PokhodenkoSA:
This PR do not change code. Only README.rst. So build failing is not a problem caused by the PR.
Admins, could you merge this PR?

PR #432, opened 2019-12-20
Fix windows issue with conda build

Vyacheslav-Smirnov:
pywin32 version which is installed with conda-build conda-verify vc vs2015_runtime vs2015_win-64 package is changed from 223 to 227
227 version sets some paths in PATH variable thus PATH overflows
Set pywin32=223 to fix it


PokhodenkoSA:
This is temporary solution.

PR #433, opened 2019-12-20
Overload series.rolling.cov()

densmirn:


sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -84,19 +84,28 @@
...
 def arr_nonnan_count(arr):
     """Count non-NaN values"""
     return len(arr) - numpy.isnan(arr).sum()
 
 
 @register_jitable
-def arr_corr(x, y):
-    """Calculate correlation of values"""
+def arr_cov(x, y, ddof):
+    """Calculate covariance of values"""
```
kozlov-alexey: Since it's a separate function, I think it would be good to either check len(x) == len(y), or directly state that both arrays should have the same size, e.g.
```suggestion
    """Calculate covariance of 1D arrays x and y of the same size"""
```

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -355,6 +364,70 @@ def hpat_pandas_series_rolling_count(self):
...
+            other_arr = main_arr
+        else:
+            other_arr = other._data
+
+        other_arr_length = len(other_arr)
+        length = max(main_arr_length, other_arr_length)
+        output_arr = numpy.empty(length, dtype=float64)
+
+        def calc_cov(main, other, ddof, minp):
+            # align arrays `main` and `other` by size and finiteness
```
kozlov-alexey: If implementation follows some pattern in pandas sources it might be good to add a reference to them here. Not insisting though.

sdc/tests/test_rolling.py
```diff
@@ -578,6 +578,85 @@ def test_impl(series, window, min_periods):
...
+                for min_periods in range(0, window, 2):
+                    with self.subTest(series=series, window=window,
+                                      min_periods=min_periods):
+                        jit_result = hpat_func(series, window, min_periods)
+                        ref_result = test_impl(series, window, min_periods)
+                        pd.testing.assert_series_equal(jit_result, ref_result)
+
+    @skip_sdc_jit('Series.rolling.cov() unsupported Series index')
+    @unittest.expectedFailure
+    def test_series_rolling_cov_issue_floating_point_rounding(self):
```
kozlov-alexey: Can you please add a docstring explaining what is the problem with rounding and on which side it is (Numba or Numpy)?

PR #441, opened 2019-12-23
Add UNEXPECTED SUCCEESS for examples CI

PokhodenkoSA:


buildscripts/test.py
```diff
@@ -234,20 +237,26 @@
...
         for item in failed_examples:
             format_print(f' - {item}: FAILED', new_block=False)
         for item in expected_failures:
             format_print(f' - {item}: EXPECTED FAILED', new_block=False)
+        for item in unexpected_success:
+            format_print(f' - {item}: UNEXPECTED SUCCESS', new_block=False)
 
-        sys.exit(0 if total_failed == 0 else -1)
+        success = (total_failed + total_unexpected_success) == 0
+        sys.exit(0 if success else -1)
```
Vyacheslav-Smirnov: Do we really need to fail build if some "expected failed" example passed?
I want to suggest to just print this case.

PokhodenkoSA: Failed build will attract attention that we should remove example from `expected_failure_list`. Or we will face situation when it become OK, than after some time become FAIL and we will not catch this jumps.

Vyacheslav-Smirnov:
Need to choose which one should be merged first - this one or https://github.com/IntelPython/sdc/pull/447

PR #448, opened 2019-12-23
Move SDC master to Numba master

AlexanderKalistratov:


AlexanderKalistratov:
@kozlov-alexey 
>so when these tests are supported in new-pipeline, running it in new-pipeline will fail

But that's the main reason to use `expectedFailure`. When this functionality would be fixed we'd know about it. So if it would for some reason pass on new pipeline we can replace it with different decorator

PR #451, opened 2019-12-24
Reimplement Series.cov using prange

AlexanderKalistratov:


AlexanderKalistratov:
@kozlov-alexey adding test which reproduce the issue here https://github.com/IntelPython/sdc/pull/465

PR #457, opened 2019-12-26
Introduce config options which enables parallel/inline for overloads

AlexanderKalistratov:
Introduce the way to configure default behavior of sdc_overloads:

1. Config parameter `config_use_parallel_overloads` and environment variable `SDC_AUTO_PARALLEL` which config if `parallel` option is enabled or disabled for all sdc overloads. Default is `False`
2. Config parameter `config_inline_overloads` and environment variable `SDC_AUTO_INLINE` which config is `inline` option is enabled for all sdc overloads. Default is `False`

### DIFF
--- a/sdc/config.py
+++ b/sdc/config.py
@@ -30,7 +30,6 @@ This is a set of configuration variables in SDC initialized at startup
 
 
 import os
-from distutils import util as distutils_util
 
 try:
     import pyarrow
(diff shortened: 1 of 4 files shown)


AlexanderKalistratov:
@PokhodenkoSA @kozlov-alexey this is important PR. It blocks performance investigation & writing perf tests by team members. Could you please review it ASAP?

sdc/config.py
```diff
@@ -47,7 +46,17 @@
...
+def strtobool(val):
+    '''Convert a string to True or False.'''
+    val = val.lower()
+    if val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    else:
+        return True
+
+
+config_transport_mpi_default = strtobool(os.getenv('SDC_CONFIG_MPI', 'True'))
```
PokhodenkoSA: Why `distutils_util`'s function is not good?

PR #468, opened 2020-01-09
Overload DataFrame.rolling() + DataFrame.rolling.min()

densmirn:


sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -0,0 +1,141 @@
...
+         returns :obj:`pandas.DataFrame` object
+"""
+
+
+def params2list(params):
+    """Convert parameters from dict to list"""
+    return ['{}={}'.format(k, v) for k, v in params.items()]
+
+
+def sdc_pandas_dataframe_rolling_apply(method_name, self, args=None, kws=None):
```
AlexanderKalistratov: Could you please separate code generation and receiving python object part?

So we would have function which returns generated function code as text and global vars needed to it?

sdc/datatypes/hpat_pandas_dataframe_rolling_types.py
```diff
@@ -0,0 +1,48 @@
...
+# OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+from numba.extending import intrinsic, register_model
+from sdc.datatypes.hpat_pandas_rolling_types import (
+    gen_hpat_pandas_rolling_init, RollingType, RollingTypeModel)
+
+
+class DataFrameRollingType(RollingType):
```
AlexanderKalistratov: I would say this type is redundant. Let's ask someone else opinion.
@kozlov-alexey @PokhodenkoSA

densmirn: I would say we are able to separate overloads of Series and DataFrame methods through separating types `DataFrameRollingType` and `SeriesRollingType`.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -39,8 +39,12 @@
 from numba.errors import TypingError
 from sdc.hiframes.pd_dataframe_type import DataFrameType
 
+from sdc.datatypes.hpat_pandas_dataframe_rolling_types import _hpat_pandas_df_rolling_init
+from sdc.datatypes.hpat_pandas_rolling_types import (
+    gen_sdc_pandas_rolling, sdc_pandas_rolling_docstring_tmpl)
 from sdc.datatypes.hpat_pandas_series_functions import TypeChecker
```
kozlov-alexey: Hasn't TypeChecker been moved to common_functions?

densmirn: It has been, but accidentally it has been copied back by somebody. Anyway I didn't change the code in this PR :)

sdc/datatypes/hpat_pandas_rolling_types.py
```diff
@@ -0,0 +1,241 @@
...
+                context.nrt.incref(builder, self, rolling.data)
+
+            return rolling._getvalue()
+
+        return sig, _codegen
+
+    return _hpat_pandas_rolling_init
+
+
+def gen_sdc_pandas_rolling(initializer, ty):
```
kozlov-alexey: It's not very clear from the name of the function what it does. I would suggest:
```suggestion
def gen_sdc_pandas_rolling_overload_body(ctor, type):
```

Also the docstring seems to be very obscure. Maybe it would be better to state that it generates code of the overloaded method using associated DataType and constructor.

sdc/datatypes/hpat_pandas_dataframe_rolling_types.py
kozlov-alexey: @AlexanderKalistratov To me it also looks like this new class is the only way to have different implementations to be called for df.rolling() and series.rolling() calls. Maybe I'm not getting something, but it really looks like a proper design.

sdc/tests/test_rolling.py
```diff
@@ -400,107 +401,164 @@ def test_impl2(df, df2, w, c):
             pd.testing.assert_frame_equal(hpat_func(*args), test_impl2(*args))
             pd.testing.assert_frame_equal(hpat_func(*args), test_impl2(*args))
 
-    @skip_sdc_jit('Series.rolling.min() unsupported exceptions')
-    def test_series_rolling_unsupported_values(self):
-        def test_impl(series, window, min_periods, center,
+    def _get_assert_equal(self, obj):
```
kozlov-alexey: This looks like a very handy function. Should we maybe move it to test_utils and use it all other tests as well?

PR #471, opened 2020-01-10
Loc with slice

1e-to:


sdc/tests/test_series.py
```diff
@@ -1295,6 +1295,65 @@ def test_impl(A):
...
+    def test_series_slice_loc_start_stop(self):
+        def test_impl(A, n, k):
+            return A.loc[n:k]
+        hpat_func = self.jit(test_impl)
+
+        all_data = [[1, 3, 5, 13, 22], [1, 3, 3, 13, 22], [22, 13, 5, 3, 1], [100, 3, 0, -3, -3]]
+        key = [1, 3, 18]
+        for index in all_data:
+            for n in key:
+                for k in key[::-1]:
```
densmirn: Please use `itertools.product` for such cases. Also let's use `unittest.subTest` here to see which parameters break functionality in case of failure.

sdc/tests/test_series.py
```diff
@@ -1295,6 +1295,65 @@ def test_impl(A):
...
+
+
+    @skip_sdc_jit('Not impl in old style')
+    def test_series_slice(self):
+        def test_impl(A):
+            return A.loc[1:5]
+        hpat_func = self.jit(test_impl)
+
+        S = pd.Series([2, 4, 6, 6, 3], [1, 3, 5, 13, 22])
+        pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
```
densmirn: Looks like test `test_series_slice_loc_start_stop` covers the same functionality as tests `test_series_slice_nonidx_nostart` and `test_series_slice`. So I propose to remove excess tests to not load CI.

sdc/tests/test_series.py
```diff
@@ -1295,6 +1295,50 @@ def test_impl(A):
...
+        key = [1, 3, 18]
+        for index in all_data:
+            k = index[0]
+            for n in key:
+                with self.subTest(index=index, start=k, stop=n):
+                    S = pd.Series([2, 4, 6, 6, 3], index)
+                    pd.testing.assert_series_equal(hpat_func(S, k, n), test_impl(S, k, n))
+
+    @skip_sdc_jit('Not impl in old style')
+    def test_series_slice_loc_start_stop(self):
```
densmirn: Looks like test `test_series_slice_loc_start_stop` completely covers test `test_series_slice_loc_stop`. So we can remove test `test_series_slice_loc_stop`. Please correct me if I am wrong.

1e-to: test_series_slice_loc_stop tests functional only for stop position, but we have limitation for functional, therefore syntax [ : n ] not support in its implementation. User must write start element, if he want to get slice with beginning in first element

densmirn: Ok, is there a logical difference between tests `test_series_slice_loc_start_stop` and `test_series_slice_loc_stop`? As I see both tests cover `start = 1, 22, 100`.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -122,10 +122,66 @@ def hpat_pandas_series_iat_impl(self, idx):
...
+                        max_diff = max(max_diff, index[i] - index[i - 1])
+                        min_diff = min(min_diff, index[i] - index[i - 1])
+
+                if max_diff*min_diff < 0:
+                    raise ValueError("Index must be monotonic increasing or decreasing")
+
+                if max_diff > 0:
+                    start_position = start_position_inc
+                    stop_position = stop_position_inc
+                    if idx_start <= index[0] and idx_stop < index[0] and start_position == 0 and stop_position == 0:
```
AlexanderKalistratov: Wouldn't
```python
if idx_stop < index[0]:
```
be enough?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -122,10 +122,66 @@ def hpat_pandas_series_iat_impl(self, idx):
...
+                        return pandas.Series(data=series._data[:0], index=series._index[:0], name=series._name)
+                else:
+                    start_position = start_position_dec
+                    stop_position = stop_position_dec if idx.stop != max_int64 else len(index)
+                    if idx_start >= index[0] and idx_stop > index[0] and start_position == 0 and stop_position == 0:
```
AlexanderKalistratov:
```python
if idx_stop > index[0]
```
?

PR #472, opened 2020-01-10
Overload df.rolling.apply()

densmirn:


AlexanderKalistratov:
I think we should add support of `args`

densmirn:
> I think we should add support of `args`

Are you sure it's supported?

AlexanderKalistratov:
> > I think we should add support of `args`
> 
> Are you sure it's supported?

Looks like it's not

PR #473, opened 2020-01-10
update test.py

samir-nasibli:
* extend tests runner with the new arg --benchmark-num-threads-list

### DIFF
--- a/buildscripts/test.py
+++ b/buildscripts/test.py
@@ -83,6 +83,8 @@ if __name__ == '__main__':
     parser.add_argument('--channel-list', default=None, help='List of channels to use: "-c <channel> -c <channel>"')
     parser.add_argument('--benchmark-argv', default='sdc.tests.tests_perf',
                         help='Run performance testing for all or a sigle one"')
+    parser.add_argument('--benchmark-num-threads-list', nargs='+',
+                        help='List of values for NUMBA_NUM_THREADS env variable', type=int)
 
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


buildscripts/test.py
```diff
@@ -267,6 +270,10 @@
 
     # Benchmark tests
     if test_mode == 'benchmark':
+        if benchmark_num_threads_list is None:
+            msg = ("List of values for NUMBA_NUM_THREADS is not defined; "
+                   "Please use --benchmark-num-threads-list option")
+            raise ValueError(msg)
```
densmirn: Isn't it possible to define number of threads automatically?

samir-nasibli: It is possible, of course. Another questions why is this needed? And what algorithm could you suggest?

densmirn: Obviously it can be needed to avoid manual setting up number of threads for each DUT you use for testing. If you think it's possible let's add the feature. Moreover I propose to define default number of threads in case of `--benchmark-num-threads-list` is not used, e.g. it can be max number of cores on DUT that can be defined automatically as well.

samir-nasibli: It makes no sense to set as the default value the maximum number of cores on the machine.
Since the meaning of these benchmark tests is scalability determination for the tested functions, for this it is necessary to specify a list of the numbers of threads.

A manual way to set the number of threads is a necessity, in this case user for a specific DUT determines a specific set of threads for an experiment and then analyzes the result.

Of course, it could be possible an automatically define this list, if `--benchmark-num-threads-list`  is not used. But in this case, I would like to clarify the algorithm by which you would like this list to be generated automatically.

Vyacheslav-Smirnov: @densmirn, now we do not have strong algorithm for calculating thread list thus I suggest to pass it via cmd arguments for now. Later we can change this approach if we want.

PR #475, opened 2020-01-11
Overload df.rolling.kurt()

densmirn:


sdc/tests/test_rolling.py
```diff
@@ -565,6 +568,21 @@ def func(x, q):
...
                         assert_equal(jit_result, ref_result)
 
+    def _test_rolling_kurt(self, obj):
+        def test_impl(obj, window, min_periods):
+            return obj.rolling(window, min_periods).kurt()
+
+        hpat_func = self.jit(test_impl)
+        assert_equal = self._get_assert_equal(obj)
+
+        for window in range(4, len(obj) + 1):
```
kozlov-alexey: Why do we start from 4? The less window values are still handled by pandas.

densmirn: Yes, less window values are handled by pandas, but not for this method.

PR #476, opened 2020-01-11
Overload df.rolling.max()

densmirn:


sdc/tests/test_rolling.py
```diff
@@ -565,6 +565,22 @@ def func(x, q):
...
                         assert_equal(jit_result, ref_result)
 
+    def _test_rolling_max(self, obj):
+        def test_impl(obj, window, min_periods):
+            return obj.rolling(window, min_periods).max()
+
+        hpat_func = self.jit(test_impl)
+        assert_equal = self._get_assert_equal(obj)
+
+        # TODO: fix the issue when window = 0
```
kozlov-alexey: Is this issue still applicable? Maybe remove all these TODOs and add single failing test for zero window case, so that we won't to forget to fix it later?

densmirn: I found out that Python implementation crashes if `window = 0`. JIT implementation works correctly. So I will make this comment more clear.

PR #482, opened 2020-01-12
Overload df.rolling.std()

densmirn:


sdc/tests/test_rolling.py
```diff
@@ -769,6 +769,28 @@ def test_impl(obj, window, min_periods):
...
+                    jit_result = hpat_func(obj, window, min_periods, ddof)
+                    ref_result = test_impl(obj, window, min_periods, ddof)
+                    assert_equal(jit_result, ref_result)
+
+    def _test_rolling_std_exception_unsupported_ddof(self, obj):
+        test_impl = rolling_std_usecase
+        hpat_func = self.jit(test_impl)
+
+        with self.assertRaises(TypingError) as raises:
+            hpat_func(obj, 3, 2, '1')
```
kozlov-alexey: It's not obvious what parameter value is invalid, maybe it would be better to use variables, like?
```
        window, min_periods, invalid_ddof = 3, 2, '1'
        with self.assertRaises(TypingError) as raises:
            hpat_func(obj, window, min_periods, invalid_ddof)
```

PR #484, opened 2020-01-13
Df isin

Rubtsowa:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -660,3 +704,140 @@ def count_overload(df, axis=0, level=None, numeric_only=False):
...
+                   'get_dataframe_data': get_dataframe_data}
+
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _reduce_impl = loc_vars[df_func_name]
+
+    return _reduce_impl
+
+
+@overload_method(DataFrameType, 'isin')
```
AlexanderKalistratov: WTF?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -279,6 +280,49 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
+def sdc_pandas_dataframe_apply_columns(df, func_name, params, ser_params):
+    all_params = ['df']
+    ser_par = []
+
+    for key, value in params.items():
+        all_params.append('{}={}'.format(key, value))
+    for key, value in ser_params.items():
+        ser_par.append('{}={}'.format(key, value))
+
+    s_par = '{}'.format(', '.join(ser_par[:]))
```
```suggestion
    s_par = ', '.join(ser_par)
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -279,6 +280,49 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
+
+
+def sdc_pandas_dataframe_apply_columns(df, func_name, params, ser_params):
+    all_params = ['df']
+    ser_par = []
+
+    for key, value in params.items():
+        all_params.append('{}={}'.format(key, value))
+    for key, value in ser_params.items():
+        ser_par.append('{}={}'.format(key, value))
```
```
def params2list(params):
    return ['{}={}'.format(k, v) for k, v in params.items()]

all_params = ['df'] + params2list(params)
ser_par = params2list(ser_params)
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -660,3 +704,140 @@ def count_overload(df, axis=0, level=None, numeric_only=False):
...
+
+def sdc_pandas_dataframe_isin_iter_codegen(df, values, func_name, ser_param):
+    all_params = ['df', 'values']
+
+    df_func_name = f'_df_{func_name}_impl'
+    func_text, global_vars = _dataframe_apply_columns_codegen(func_name, all_params, ser_param, df.columns)
+
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _reduce_impl = loc_vars[df_func_name]
```
```suggestion
    _apply_impl = loc_vars[df_func_name]
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -660,3 +704,140 @@ def count_overload(df, axis=0, level=None, numeric_only=False):
...
+        # func_lines += [
+        #     f'  series_{c} = pandas.Series(get_dataframe_data({all_params[0]}, {i}))',
+        #     f'  result = []',
+        #     f'  for i in series_{c}._index:',
+        #     f'    if series_{c}[i] == values[i]:',
+        #     f'      result.append(True)',
+        #     f'    else:',
+        #     f'      result.append(False)',
+        #     f'  {result_c} = pandas.Series(result, index=series_{c}._index)'
+        # ]
```
densmirn: Let's remove commented code.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -660,3 +704,140 @@ def count_overload(df, axis=0, level=None, numeric_only=False):
...
+    df_func_name = f'_df_{func_name}_impl'
+    func_text, global_vars = _dataframe_apply_columns_codegen(func_name, all_params, ser_param, df.columns)
+
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _reduce_impl = loc_vars[df_func_name]
+    return _reduce_impl
+
+
+def sdc_pandas_dataframe_isin_ser_codegen(df, values, func_name):
```
densmirn: Could you please separate code generation and receiving python object part?

So we would have function which returns generated function code as text and global vars needed to it.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -660,3 +704,140 @@ def count_overload(df, axis=0, level=None, numeric_only=False):
...
+                   'get_dataframe_data': get_dataframe_data}
+
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _reduce_impl = loc_vars[df_func_name]
+
+    return _reduce_impl
+
+
+def sdc_pandas_dataframe_isin_df_codegen(df, values, func_name):
```
densmirn: Could you please separate code generation and receiving python object part? You could look such change in https://github.com/IntelPython/sdc/pull/468.

sdc/tests/test_dataframe.py
```diff
@@ -1103,6 +1110,28 @@ def test_impl(df):
...
+            return df.isin(vals)
+
+        hpat_func = self.jit(test_impl)
+        n = 11
+        df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n)**2})
+        pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
+
+    def test_isin_set1(self):
+        def test_impl(df):
+            vals = values = {1, 2, 5, 7, 8}
```
densmirn: Please remove unused variables.

sdc/tests/test_dataframe.py
```diff
@@ -1103,6 +1110,28 @@ def test_impl(df):
         df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n)**2})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    @skip_sdc_jit
+    def test_isin_ser1(self):
+        def test_impl(df):
+            vals = pd.Series([2, 3, 4])
+            return df.isin(vals)
```
densmirn: Could we pass `vals` as a parameter to the `test_impl`? if it's so let's combine all the tests to a single one through `unittest.subTest`

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -977,3 +974,141 @@ def pct_change_overload(df, periods=1, fill_method='pad', limit=None, freq=None)
...
+    result_name = []
+    joined = ', '.join(all_params)
+    func_lines = [f'def _df_{func_name}_impl({joined}):']
+    for i, c in enumerate(columns):
+        result_c = f'result_{c}'
+        func_lines += [
+            f'  series_{c} = pandas.Series(get_dataframe_data({all_params[0]}, {i}))',
+            f'  result = []',
+            f'  for i in range(len(series_{c}._data)):',
+            f'    if series_{c}._data[i] == values._data[i]:',
```
kozlov-alexey: Why are you comparing elements of the same indexes from series and values, is this correct?
Most likely the check should be 
`"if series_{c}._data[i] in values._data:'"`

densmirn:
> `.idea/workspace.xml` and other unnecessary files should be removed.

Do you exactly need to keep `.pyc` files in the repo?

AlexanderKalistratov:
As far as I remember, the implementation wasn't handling case with index correctly. Assuming it is fixed, why there are still no tests for it?

sdc/tests/test_dataframe.py
```diff
@@ -1339,6 +1339,7 @@ def test_impl(df, df2):
         pd.testing.assert_frame_equal(hpat_func(df, df2), test_impl(df, df2))
 
     @unittest.skip("not work with index")
+    @skip_sdc_jit
```
densmirn: Please add a message about why did you skip that.

densmirn: > Please add a message about why did you skip that.

Moreover the test was skipped before, `@skip_sdc_jit` doesn't make sense.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -487,10 +488,11 @@ def sdc_pandas_dataframe_apply_columns(df, func_name, params, ser_params):
     all_params = ['df']
     ser_par = []
 
-    for key, value in params.items():
-        all_params.append('{}={}'.format(key, value))
-    for key, value in ser_params.items():
-        ser_par.append('{}={}'.format(key, value))
+    def params2list(params):
```
kozlov-alexey: I

kozlov-alexey: I think you can simply use kwsparams2list from sdc_typing_utils.

sdc/tests/test_dataframe.py
```diff
@@ -1454,6 +1454,16 @@ def test_impl(df):
 
         pd.testing.assert_series_equal(sdc_func(df), test_impl(df))
 
+    @unittest.skip('DF.getitem unsupported idx as a tuple')
```
kozlov-alexey: @Rubtsowa, @densmirn 
I think this test is no longer needed, since it's tested with _test_df_getitem_tuple_idx, am I right?

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
...
+        hpat_func(df, df2)
         pd.testing.assert_frame_equal(hpat_func(df, df2), test_impl(df, df2))
 
-    @unittest.skip("needs dict typing in Numba")
+    @skip_sdc_jit
+    def test_isin_df2(self):
+        def test_impl(df, df2):
+            return df.isin(df2)
+
+        cfunc = self.jit(test_impl)
```
kozlov-alexey: Please use consistent names across all the tests, e.g test_impl, hpat_func, etc.

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
         df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n)**2})
         df2 = pd.DataFrame({'A': np.arange(n), 'C': np.arange(n)**2})
         df2.A[n // 2:] = n
+        hpat_func(df, df2)
```
kozlov-alexey: What's the purpose of this line?

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
...
         pd.testing.assert_frame_equal(hpat_func(df, df2), test_impl(df, df2))
 
-    @unittest.skip("needs dict typing in Numba")
+    @skip_sdc_jit
+    def test_isin_df2(self):
```
kozlov-alexey: Please make all tests have descriptive names, so that it's clear what the test actually verifies, e.g. if the purpose of this one is to verify isin works with dfs with indexes, you could name it test_isin_df_index_str.

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
...
+        def test_impl(df, df2):
+            return df.isin(df2)
+
+        cfunc = self.jit(test_impl)
+        df = pd.DataFrame({'num_legs': [2, 4], 'num_wings': [2, 0]}, index=['falcon', 'dog'])
+        df2 = pd.DataFrame({'num_legs': [8, 2], 'num_wings': [0, 2]}, index=['spider', 'falcon'])
```
kozlov-alexey: Can you extend the data of dfs so that more combinations of results are tested?
That is for index='falcon' you test all values of row in df are in corresponding values in df2 row, but that's only one of the possible results, can you add those that lead to isin False as well? Not sure if that's not already tested with other tests though.

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
...
+    def test_isin(self):
+        def test_impl(df, vals):
+            return df.isin(vals)
+
+        hpat_func = self.jit(test_impl)
+        n = 11
+        df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n) ** 2})
+        df2 = pd.DataFrame({'A': np.arange(n), 'C': np.arange(n) ** 2})
+        df2.A[n // 2:] = n
+        values = [{1, 2, 5, 7, 8}, [2, 3, 4], pd.DataFrame({}), df2]
```
kozlov-alexey: Combining values of types that have separate implementations is not good. I would suggest separate this test into two: first for list and set which fall into iter_impl and second that tests isin_df.

sdc/tests/test_dataframe.py
```diff
@@ -1473,30 +1482,67 @@ def test_impl(df, df2):
...
         df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n)**2})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
-    @skip_numba_jit
-    def test_isin_list1(self):
+    @skip_sdc_jit
+    def test_isin_ser1(self):
         def test_impl(df):
-            vals = [2, 3, 4]
+            vals = pd.Series([2, 3, 4])
```
kozlov-alexey: There's only one True in result. I think it's better to use bigger data with duplicate values, e.g.
vals = pd.Series([2, 3, 4, 3, 4, 5, 8, 49])
and probably passed series as argument to have all input data in one place (outside the jitted func).

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1396,6 +1398,243 @@ def pct_change_overload(df, periods=1, fill_method='pad', limit=None, freq=None)
...
+
+    global_vars = {'pandas': pandas,
+                   'numba': numba,
+                   'numpy': numpy,
+                   'get_dataframe_data': get_dataframe_data}
+
+    return func_text, global_vars
+
+
+def sdc_pandas_dataframe_isin_ser(df, values, func_name):
```
kozlov-alexey: The functions:
    sdc_pandas_dataframe_isin_dict
    sdc_pandas_dataframe_isin_ser
    sdc_pandas_dataframe_isin_df
look almost identical and differ only in the applied function code generator.
I think you can apply the same approach as was used for getitem, see: [hpat_pandas_dataframe_functions.py](https://github.com/IntelPython/sdc/blob/ec7e4f7fd3325391285b4856c092fb24b957e36f/sdc/datatypes/hpat_pandas_dataframe_functions.py#L1367)

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1396,6 +1398,243 @@ def pct_change_overload(df, periods=1, fill_method='pad', limit=None, freq=None)
     return sdc_pandas_dataframe_apply_columns(df, name, params, ser_par)
 
 
+def sdc_pandas_dataframe_isin_iter_codegen(df, values, func_name, ser_param):
+    all_params = ['df', 'values']
```
kozlov-alexey: These values 'df' and 'values' are actually args of produced functions that must match the overload arguments (also df and values), so it's not good to have them hardcoded in any other function except the top level overload. I would suggest passing them into codegen in the same was as done for groupby impls: [hpat_pandas_groupby_functions.py](https://github.com/IntelPython/sdc/blob/ec7e4f7fd3325391285b4856c092fb24b957e36f/sdc/datatypes/hpat_pandas_groupby_functions.py#L412)

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1396,6 +1398,243 @@ def pct_change_overload(df, periods=1, fill_method='pad', limit=None, freq=None)
...
+    df_func_name = f'_df_{func_name}_impl'
+    func_text, global_vars = _dataframe_apply_columns_codegen(func_name, all_params, ser_param, df.columns)
+
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _apply_impl = loc_vars[df_func_name]
+    return _apply_impl
+
+
+def sdc_pandas_dataframe_isin_dict_codegen(func_name, df_type, values, all_params, columns):
```
kozlov-alexey: Can we not pass columns into this function, since we pass df_type anyway?

PR #486, opened 2020-01-14
Add perf tests for corr

Rubtsowa:


Rubtsowa:
median       min       max   compile    boxing
name        nthreads type      size
series_corr 1.0      JIT       10000000.0  0.119788  0.119502  0.119835  0.319196  0.000269
                     Reference 10000000.0  0.106864  0.105005  0.115487       NaN       NaN
            28.0     JIT       10000000.0  0.087989  0.087780  0.088321  0.330122  0.000292
                     Reference 10000000.0  0.105538  0.103525  0.119165       NaN       NaN

PR #494, opened 2020-01-14
Overload df.rolling.corr()

densmirn:


examples/dataframe/rolling/dataframe_rolling_corr.py
```diff
@@ -0,0 +1,32 @@
+
```
kozlov-alexey: Seems the top of the copyright header was erased.

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -78,43 +82,175 @@
...
-    method_params = args + ['{}={}'.format(k, k) for k in kwargs]
-    method_params_as_str = ', '.join(method_params)
-
-    impl_params = ['self'] + args + ['{}={}'.format(k, v) for k, v in kwargs.items()]
+    rolling_params = df_rolling_params_codegen()
+    method_kws = {k: k for k in kwargs}
+    impl_params = ['self'] + args + params2list(kwargs)
     impl_params_as_str = ', '.join(impl_params)
 
+    data_columns = {col: idx for idx, col in enumerate(self.data.columns)}
```
kozlov-alexey: Maybe this way?
```suggestion
    data_columns = dict(zip(self.data.columns, range(len(self.data.columns))))
```

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -78,43 +82,175 @@
...
+    other_columns = {col: idx for idx, col in enumerate(other.columns)}
+    common_columns = [col for col in data_columns if col in other_columns]
+    all_columns = [col for col in data_columns]
```
kozlov-alexey: Does the order matter? Can we just use intersection and union of sets, i.e.
```suggestion
    self_columns_set, other_columns_set = set(data_columns), set(other_columns)
    common_columns = self_columns_set.intersection(other_columns_set)
    all_columns = self_columns_set.union(other_columns_set)
```

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
densmirn: I don't see principle difference between these ways. @kozlov-alexey what is the purpose of the change in your opinion?

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
densmirn: Nice remark, but unit tests showed that the columns order matters.

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -125,6 +263,32 @@ def gen_df_rolling_method_impl(method_name, self, args=None, kws=None):
     return _impl
 
 
+@sdc_overload_method(DataFrameRollingType, 'corr')
+def sdc_pandas_dataframe_rolling_corr(self, other=None, pairwise=None):
+
+    ty_checker = TypeChecker('Method rolling.corr().')
+    ty_checker.check(self, DataFrameRollingType)
+
+    accepted_other = (bool, Omitted, NoneType, DataFrameType, SeriesType)
```
kozlov-alexey:
```suggestion
    accepted_other = (Omitted, NoneType, DataFrameType, SeriesType)
```

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -134,6 +298,24 @@ def sdc_pandas_dataframe_rolling_min(self):
...
+    'limitations_block':
+    """
+    Limitations
+    -----------
+    DataFrame elements cannot be max/min float/integer. Otherwise SDC and Pandas results are different.
+    """,
+    'extra_params':
+    """
+    other: :obj:`Series`
+        Other Series.
```
kozlov-alexey: Probably it should be:
```
    other: :obj:`Series` or :obj:`DataFrame`
        Other Series or DataFrame.
```

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
kozlov-alexey: I just thought that it looks better than dict comprehension, since there's no need to swap keys and values. I don't mind keeping it as-is though.

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -78,43 +82,177 @@
 """
 
 
-def df_rolling_method_codegen(method_name, self, args=None, kws=None):
+def params2list(params):
```
kozlov-alexey: We could move this function into common_functions.py and use it in all other DF generation functions (e.g. in sdc_pandas_dataframe_reduce_columns, sdc_pandas_dataframe_apply_columns).

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -125,6 +263,32 @@ def gen_df_rolling_method_impl(method_name, self, args=None, kws=None):
...
+    nan_other = isinstance(other, (Omitted, NoneType)) or other is None
+    kws = {'other': 'None', 'pairwise': 'None'}
+
+    if nan_other:
+        return gen_df_rolling_method_without_other_impl('corr', self, kws=kws)
+
+    if isinstance(other, DataFrameType):
+        return gen_df_rolling_method_with_other_df_impl('corr', self, other, kws=kws)
+
+    return gen_df_rolling_method_impl('corr', self, kws=kws)
```
kozlov-alexey: @densmirn I think this can be a bit improved by always passing the other type into the real generation function (e.g. df_rolling_method_codegen) and inside it selecting a branch based on the other type (which can be None, Omitted, DataFrame and Series). This way we can join the functions: df_rolling_method_without_other_codegen and df_rolling_method_codegen, which are almost the same into one, e.g.

```
def df_rolling_method_codegen(method_name, self, other, args=None, kws=None):
    args = args or []
    kwargs = kws or {}

    impl_params = ['self'] + args + params2list(kwargs)
    impl_params_as_str = ', '.join(impl_params)

    impl_name = f'_df_rolling_{method_name}_impl'
    func_lines = [f'def {impl_name}({impl_params_as_str}):']

    if not other:
        if 'pairwise' in kwargs:
            func_lines += [
                '  if pairwise is None:',
                '    _pairwise = True',
                '  else:',
                '    _pairwise = pairwise',
                '  if _pairwise:',
                '    raise ValueError("Method rolling.corr(). The object pairwise\\n expected: False")'
            ]

        method_params = args + ['{}={}'.format(k, k) for k in kwargs if k != 'other']
    else:
        method_params = args + ['{}={}'.format(k, k) for k in kwargs]
    func_lines += df_rolling_method_tail_codegen(method_params, self.data.columns, method_name)

    global_vars = {'pandas': pandas, 'get_dataframe_data': get_dataframe_data}

    return func_text, global_vars
```
What do you think?

densmirn: Done!

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -78,45 +82,149 @@
...
+    func_lines += [f'  return pandas.DataFrame({{{data}}})']
+    func_text = '\n'.join(func_lines)
+
+    global_vars = {'numpy': numpy, 'pandas': pandas, 'float64': float64,
+                   'get_dataframe_data': get_dataframe_data}
+
+    return func_text, global_vars
+
+
+def df_rolling_method_tail_codegen(method_params, df_columns, method_name):
```
kozlov-alexey: @densmirn I think this function can really be incorporated into df_rolling_method_codegen, since it's used only once now. So that df_rolling_method_codegen and df_rolling_method_with_other_df_codegen have similar look.

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -78,45 +82,149 @@
...
+    args = args or []
+    kwargs = kws or {}
+
+    impl_params = ['self'] + args + params2list(kwargs)
+    impl_params_as_str = ', '.join(impl_params)
+
+    impl_name = f'_df_rolling_{method_name}_impl'
+    func_lines = [f'def {impl_name}({impl_params_as_str}):']
+
+    if 'other' in kwargs and not with_other:
```
kozlov-alexey: From what I see below 'other' will always be in kwargs, am I wrong? Can this part of the check be omitted?
Also I still think it's better to pass other as a type here (but not as a boolean variable that indicates that other has certain types) and make the decision here rather than on the call site, e.g.
```suggestion
    other_is_none_or_default = isinstance(other, (Omitted, NoneType)) or other is None
    if other_is_none_or_default:
```

kozlov-alexey: OK, it looks like 'other' wont be in kwargs for other rolling methods. So with my suggestions we reduce the code a bit, but actually make it more complex. Let's keep the original version then, but keep consistent naming of gen functions, e.g. df_rolling_method_other_df_codegen, df_rolling_method_other_none_codegen, etc.

PR #497, opened 2020-01-14
Adding support for StringArrays comparison operators

kozlov-alexey:


sdc/sdc_autogenerated.py
```diff
@@ -1516,3 +1510,267 @@ def _series_operator_eq_common_impl(self, other):
...
+        return None
+
+    if operands_are_arrays:
+        def _sdc_str_arr_operator_lt_impl(self, other):
+            if len(self) != len(other):
+                raise ValueError("Mismatch of String Arrays sizes in operator.lt")
+            n = len(self)
+            out_list = [False] * n
+            for i in numba.prange(n):
+                out_list[i] = out_list[i] = (self[i] < other[i]
```
densmirn: Please apply such fix in all the places.
```suggestion
                out_list[i] = (self[i] < other[i]
```

sdc/str_arr_ext.py
```diff
@@ -1438,3 +1433,13 @@ def append_string_array_to(result, pos, A):
         j += 1
 
     return i
+
+
+@numba.njit(no_cpython_wrapper=True)
+def make_str_arr_from_list(str_list):
+
+    data_total_chars = np.sum(np.asarray([len(s) for s in str_list]))
```
densmirn: Why not avoid converting list to the array and just get sum from the list of lengths? Something like that:
`data_total_chars = sum(len(s) for s in str_list)`

sdc/sdc_autogenerated.py
kozlov-alexey: Thank you, fixed it.

sdc/str_arr_ext.py
kozlov-alexey: @densmirn I tried that, but it doesn't work in nopython mode (see [Numba python builtin functions](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#built-in-functions). 
Also the line
`data_total_chars = sum(len(s) for s in str_list)`
requires the use of generators in a closure call, which is not supported either.

But you're probably right and there's no need for conversion here, so I changed it to explicit prange.

buildscripts/autogen_sources.py
```diff
@@ -116,9 +116,10 @@
     imports_start_line, import_end_line = min(imports_line_numbers), max(imports_line_numbers)
     import_section_text = ''.join(module_text_lines[imports_start_line: import_end_line + 1])
 
-    # read function templates for arithmetic and comparison operators from tempaltes module
-    func_template_arithmetic_binop = inspect.getsource(templates_module.sdc_pandas_series_operator_binop)
-    func_template_comparison_binop = inspect.getsource(templates_module.sdc_pandas_series_operator_comp_binop)
+    # read function templates for arithmetic and comparison operators from templates module
+    template_series_arithmetic_binop = inspect.getsource(templates_module.sdc_pandas_series_operator_binop)
+    template_series_comparison_binop = inspect.getsource(templates_module.sdc_pandas_series_operator_comp_binop)
+    template_str_arr_comparison_binop = inspect.getsource(templates_module.sdc_str_arr_operator_comp_binop)
```
AlexanderKalistratov: Shouldn't we also add some str arithmetic operations?

sdc/sdc_autogenerated.py
```diff
@@ -1516,3 +1509,267 @@ def _series_operator_eq_common_impl(self, other):
...
+    if operands_are_arrays:
+        def _sdc_str_arr_operator_lt_impl(self, other):
+            if len(self) != len(other):
+                raise ValueError("Mismatch of String Arrays sizes in operator.lt")
+            n = len(self)
+            out_list = [False] * n
+            for i in numba.prange(n):
+                out_list[i] = (self[i] < other[i]
+                               and not (str_arr_is_na(self, i) or str_arr_is_na(other, i)))
+            return out_list
```
kozlov-alexey: This compiles and work in some cases, but should not be merged, since returning list instead of np.array of bools forbids  using StringArray comparisons as masks in numpy operations. Instead the core issue with incorrect IR should be fixed.

AlexanderKalistratov:
@kozlov-alexey so...?

buildscripts/autogen_sources.py
kozlov-alexey: Done, added operator.add and operator.mul for string series (as a separate overload though, not to complicate the common template)

sdc/sdc_autogenerated.py
kozlov-alexey: The Numba issue is investigated here: https://github.com/numba/numba/issues/5157

kozlov-alexey:
@AlexanderKalistratov I had to return back astype call for one specialization, cause when numpy calculates // and % in integer domain the result is actually different from pandas. I think we should rely on astype making a copy of desired dtype.

AlexanderKalistratov:
Merge conflicts, binary file added & pep issues

AlexanderKalistratov:
Please comment on https://github.com/IntelPython/sdc/pull/535

kozlov-alexey:
> Merge conflicts, binary file added & pep issues

Done.

AlexanderKalistratov:
@kozlov-alexey could you please fix the rest of pep issues?

kozlov-alexey:
> @kozlov-alexey could you please fix the rest of pep issues?

@AlexanderKalistratov I actually fixed it, don't know why PEP still points to it, see below:
[sdc_typing_utils.py](https://github.com/IntelPython/sdc/blob/bb758174a1e47867e4a105c86085430140c43ff4/sdc/utilities/sdc_typing_utils.py#L158)

PR #498, opened 2020-01-15
Loc array

1e-to:


sdc/tests/test_series.py
```diff
@@ -1257,6 +1257,16 @@ def test_impl(A):
...
         np.testing.assert_array_equal(hpat_func(S), test_impl(S))
 
+    @skip_sdc_jit('Not impl in old style')
+    def test_series_loc_array(self):
+        def test_impl(A, n):
+            return A.loc[n]
+        hpat_func = self.jit(test_impl)
+
+        S = pd.Series([1, 2, 4, 8, 6, 0], [1, 2, 4, 0, 6, 0])
+        n = np.array([0, 4, 2])
```
densmirn: Maybe to extend the test by passing parameter `n` both as a list and an array.

1e-to: added

PR #501, opened 2020-01-15
impl loc callable

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -149,6 +151,22 @@ def hpat_pandas_series_loc_impl(self, idx):
...
+                if len(tmp) > 0:
+                    new_series = new_series.append(tmp)
+                else:
+                    new_series = new_series.append(pandas.Series(numpy.array([numpy.nan]), numpy.array([res[i]])))
+
+            return new_series
+
+        return hpat_pandas_series_loc_callable_impl
+
         raise TypingError('{} The index must be an Number, Slice, String, List, Array or a callable.\
```
densmirn: Looks like we need to remove the exception raising.

sdc/tests/test_series.py
```diff
@@ -1257,6 +1257,23 @@ def test_impl(A):
         S = pd.Series([2, 4, 6], ['1', '3', '5'])
         np.testing.assert_array_equal(hpat_func(S), test_impl(S))
 
+    @skip_parallel
+    @skip_sdc_jit('Not impl in old style')
+    def test_series_loc_callable(self):
+        def test_impl(S):
+            return S.loc[(lambda a: a)]
```
densmirn: Do you exactly need round brackets around of lambda?

densmirn: Maybe to complicate lambda, e.g. `a` -> `a ** 2`?

sdc/tests/test_series.py
```diff
@@ -1257,6 +1257,23 @@ def test_impl(A):
...
+        hpat_func = self.jit(test_impl)
+        S = pd.Series([0, 6, 4, 7, 8], [0, 6, 66, 6, 8])
+        pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
+
+    @unittest.skip('Loc callable return float Series')
+    def test_series_loc_callable2(self):
```
densmirn: What do you think about merging of `test_series_loc_callable` and `test_series_loc_callable2`?

densmirn: Why not do that?

1e-to: Because implementation has limit and always return float Series. In that case pandas return int Series, and we return float Series. I wrote about it in notes in loc. 
Test skips with expected failure

PR #507, opened 2020-01-15
Pct change perf test

Rubtsowa:


sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -241,6 +241,15 @@ def usecase_series_nunique(input_data):
     return finish_time - start_time, res
 
 
+def usecase_series_pct_change(input_data):
+    start_time = time.time()
+    res = input_data.pct_change(periods=1, limit=None, freq=None)
```
PokhodenkoSA: If we call function with specific parameters it should be also shown in usecase name.

sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -241,6 +241,15 @@ def usecase_series_nunique(input_data):
     return finish_time - start_time, res
 
 
+def usecase_series_pct_change(input_data):
```
PokhodenkoSA:
```suggestion
def usecase_series_pct_change_periods_1_limit_None_freq_None(input_data):
```

PR #511, opened 2020-01-16
pd.DataFrame.index impl

akharche:
Index attribute + unboxing


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -48,7 +46,67 @@
...
+    """
+
+    ty_checker = TypeChecker(f'Attribute index.')
+    ty_checker.check(df, DataFrameType)
+
+    def sdc_pandas_dataframe_index_impl(df):
+        indent = 4 * ' '
+        func_args = ['df']
+
+        func_definition = [f'def sdc_pandas_dataframe_index_impl({", ".join(func_args)}):']
```
AlexanderKalistratov: Why do we need code generation here?

akharche: We need the number of rows = the length of the column array to generate RangeIndex when it is not set explicitly

AlexanderKalistratov: Ok. And why do we need code generation?

akharche: Improved

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -48,7 +46,60 @@
...
+    """
+
+    ty_checker = TypeChecker(f'Attribute index.')
+    ty_checker.check(df, DataFrameType)
+
+    if isinstance(df.index, types.NoneType) or df.index is None:
+        empty_df = not df.columns
+
+        def hpat_pandas_df_index_none_impl(df):
+            if empty_df == True:  # nq
```
densmirn: Sorry for misleading, you need to replace `# nq` with `# noqa`

densmirn: Don't you want to try it in a single line?
`df_len = 0 if empty_df == False else len(get_dataframe_data(df, 0))`

But you need to check that it works.

sdc/tests/test_dataframe.py
```diff
@@ -544,6 +544,50 @@ def test_impl(n):
...
         self.assertEqual(count_parfor_REPs(), 0)
 
+    @skip_sdc_jit
+    def test_index_attribute(self):
+        index_to_test = [[1, 2, 3, 4, 5],
+                         [.1, .2, .3, .4, .5],
+                         ['a', 'b', 'c', 'd', 'e']]
+        n = 5
+        A = np.ones(n)
+        B = np.random.ranf(n)
```
densmirn: Please set up seed for the random generation here and in other places.

sdc/tests/test_dataframe.py
```diff
@@ -544,6 +544,50 @@ def test_impl(n):
         self.assertEqual(count_array_REPs(), 0)
         self.assertEqual(count_parfor_REPs(), 0)
 
+    @skip_sdc_jit
+    def test_index_attribute(self):
```
densmirn: I propose to add generic method to test `df.index`, something like that:
```
def _test_df_index(self, df):
    def test_impl(df):
        return df.index

    np.testing.assert_array_equal(sdc_func(df), test_impl(df))
```
then use it in 3 places, e.g.:
```
@skip_sdc_jit
def test_df_index_attribute_empty(self):
    n = 5
    A = np.ones(n)
    B = np.random.ranf(n)
    df = pd.DataFrame({'A': A, 'B': B})

    self._test_df_index(df)
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -48,7 +46,59 @@
...
+       -----------
+       df: :obj:`pandas.DataFrame`
+           input arg
+       Returns
+       -------
+       :obj: `numpy.array`
+           return the index of DataFrame
+    """
+
+    ty_checker = TypeChecker(f'Attribute index.')
```
densmirn: Using of f-string is excess here.

PR #515, opened 2020-01-16
Series perf tests

1e-to:


sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -217,11 +217,12 @@ def {func_name}(self):
     ('idxmax', '', [10 ** 9]),
     ('idxmin', '', [10 ** 9]),
     ('isna', '', [2 * 10 ** 7]),
-    ('max', '', [10 ** 9]),
+    ('max', '', [10 ** 8]),
+    ('max', 'skipna=False', [10 ** 8]),
```
AlexanderKalistratov: Shouldn't we also add parameters 'skipna=False' for `mean`, `median`, `idxmax`, `idxmin`, `var`, `cumsum` and so on?
@PokhodenkoSA

sdc/tests/tests_perf/test_perf_series.py
```diff
@@ -234,14 +235,25 @@ def {func_name}(self):
     ('value_counts', '', [3 * 10 ** 5]),
     ('var', '', [5 * 10 ** 8]),
     ('unique', '', [10 ** 5]),
+    ('isnull', '', [10 ** 7]),
+    ('notna', '', [10 ** 7]),
+    ('var', '', [10 ** 8]),
```
AlexanderKalistratov: `var` added twice

sdc/tests/tests_perf/test_perf_series.py
PokhodenkoSA: I think we should check that it influence on scaling first. Then we should add this cases to show difference.

sdc/tests/tests_perf/test_perf_series.py
PokhodenkoSA: It is better to sort lines alphabetically.

PR #517, opened 2020-01-17
Add perf.test on df.rolling.min()

densmirn:
Additionally add unit test on df.rolling.min() to save figured out issue
when tested DataFrame has more than 19 columns.


sdc/tests/tests_perf/test_perf_df_rolling.py
```diff
@@ -0,0 +1,136 @@
...
+
+# python -m sdc.runtests sdc.tests.tests_perf.test_perf_df_rolling.TestDFRollingMethods
+class TestDFRollingMethods(TestBase):
+    # more than 19 columns raise SystemError: CPUDispatcher() returned a result with an error set
+    max_columns_num = 19
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.total_data_length = {
```
Rubtsowa: I think this dictionary can move outside of class.

sdc/tests/tests_perf/test_perf_df_rolling.py
```diff
@@ -0,0 +1,136 @@
...
+            record['test_type'] = 'Python'
+            self._test_python(pyfunc, record, test_data)
+            self.test_results.add(**record)
+
+    def _test_df_rolling_method(self, name):
+        usecase = gen_df_rolling_usecase(name)
+        test_name = f'DataFrame.rolling.{name}'
+        self._test_case(usecase, name, test_name=test_name)
+
+    def test_df_rolling_min(self):
```
Rubtsowa: This is function can be obtained using template for tests. For example, how in test_perf_series.py

sdc/tests/tests_perf/test_perf_df_rolling.py
densmirn: @Rubtsowa what advantages do you think you proposal have?

Rubtsowa: Compact code. out of class we can pass all necessary parameters in 1 line.

sdc/tests/tests_perf/test_perf_df_rolling.py
densmirn: I tried to implement such generator:
```
test_params = {
    'min': {'rolling': None, 'method': ''}
}

for name, params in test_params.items():
    test_name = f'test_df_rolling_{name}'
    test_text = f"""
    def {test_name}(self):
        self._test_df_rolling_method("{name}", rolling_params={params['rolling']}, method_params={params['method']})
    """

    loc_vars = {}
    exec(test_text, {}, loc_vars)
    _df_rolling_test = loc_vars[test_name]

    setattr(TestDFRollingMethods, test_name, _df_rolling_test)
```

My opinion the generator is excess because actually there are no real advantages from that, e.g. we we don't even shorten the code.

sdc/tests/tests_perf/test_perf_df_rolling.py
densmirn: I agree to move it out of the class if it will be used there.

sdc/tests/tests_perf/test_perf_df_rolling.py
Rubtsowa: if add in test_params in values "data_length" then will not need  to conect with list "total_data_length" in class. Also definition parameters will in one place that convenient.

densmirn: I agree with that, but I don't like how the generator looks (it looks harder than tests itself) and I don't see real advantages of that.

densmirn: @Rubtsowa in conclusion, I propose to implement generic test generator in separate PR which can be used for generating most the performance tests. In this case we could use it here as well.

PR #518, opened 2020-01-17
Add gen ser str test perf

Rubtsowa:


sdc/tests/tests_perf/test_perf_series_str.py
```diff
@@ -36,78 +36,32 @@
...
 
-def usecase_series_len(input_data):
-    start_time = time.time()
-    res = input_data.str.len()
-    finish_time = time.time()
+test_global_input_data_unicode_kind1 = [
+    'ascii',
+    '12345',
+    '1234567890',
+]
```
1e-to: This is already in another file, you can simply import

PR #519, opened 2020-01-17
Use prange in apply

PokhodenkoSA:
`args` are not supported still.
This PR makes possible to use func() which converts type. It is implemented using get_call_type().
Also it uses prange to make `apply()` parallel.

### DIFF
--- a/sdc/datatypes/pandas_series_functions/apply.py
+++ b/sdc/datatypes/pandas_series_functions/apply.py
@@ -26,7 +26,8 @@
 
 import numpy
 import pandas
-import numba
+from numba import prange, types
+from numba.targets.registry import cpu_target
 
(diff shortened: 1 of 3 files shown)


AlexanderKalistratov:
@densmirn please consider. It looks like that *args now working

AlexanderKalistratov:
@PokhodenkoSA any performance tests?

densmirn:
> @densmirn please consider. It looks like that *args now working

Thanks for the notice :)

PokhodenkoSA:
> @PokhodenkoSA any performance tests?

image

PokhodenkoSA:
> Please fix limitation block in docstring according to the fact that `args` is supported.

Fixed. Actually `args` is not supported still.

PR #521, opened 2020-01-19
Removed mpi dependency

dmitrii-zagornyi:


AlexanderKalistratov:
@dmitrii-zagornyi have you checked that perf tests still working?

dmitrii-zagornyi:
> @dmitrii-zagornyi have you checked that perf tests still working?
Perf tests before my changes not working too. The error is the same and looks like my changes not affected perf tests.

PR #527, opened 2020-01-21
Add docs for 13 series func

1e-to:


examples/series/series_index.py
```diff
@@ -0,0 +1,39 @@
...
+# *****************************************************************************
+
+import numpy as np
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_index():
+    series = pd.Series(np.arange(5))
```
akharche: Data and index are equal in this example, so the return expectation is not obvious. May be it is better to set index explicitly?
`series = pd.Series(np.arange(5), index=[1, 2, 3, 4, 5])`

examples/series/series_shape.py
```diff
@@ -0,0 +1,39 @@
...
+import numpy as np
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_shape():
+    series = pd.Series(np.arange(10))
+
+    return series.shape  # Expect series of 9, 8, 7, 6
```
akharche:
```suggestion
    return series.shape  # Expect (10,)
```

examples/series/series_std.py
```diff
@@ -0,0 +1,39 @@
...
+import numpy as np
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_std():
+    series = pd.Series(np.arange(10))
+
+    return series.std()  # Expect series of 9, 8, 7, 6
```
akharche: I am not sure that Expectation is right

examples/series/series_var.py
```diff
@@ -0,0 +1,39 @@
...
+import numpy as np
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_var():
+    series = pd.Series(np.arange(10))
+
+    return series.var()  # Expect value: 9.166666666666666
```
AlexanderKalistratov: Not sure we can specify it with such precision. In case of parallel execution value could be different
@PokhodenkoSA @Hardcode84

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -641,6 +670,35 @@ def hpat_pandas_series_nsmallest_impl(self, n=5, keep='first'):
...
+    ********************************************
+
+    Pandas API: pandas.Series.nlargest
+
+    Examples
+    --------
+    .. literalinclude:: ../../../examples/series/series_nlargest.py
+       :language: python
+       :lines: 27-
+       :caption: Return the smallest n elements.
```
AlexanderKalistratov: largest n elements?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1278,41 +1530,78 @@ def hpat_pandas_series_len_impl(self):
...
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+
+    Pandas API: pandas.Series.astype
+
+    Examples
+    --------
+    .. literalinclude:: ../../../examples/series/series_astype.py
+       :language: python
+       :lines: 27-
```
akharche: You added Expected result before the function, may be it is better to shift to 41- line and not to show the expected result because it will be in the output?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1381,15 +1670,46 @@ def hpat_pandas_series_astype_no_modify_impl(self, dtype, copy=True, errors='rai
...
+
+    .. note::
+
+        Parameter axis is currently unsupported by Intel Scalable Dataframe Compiler
+
+    Examples
+    --------
+    .. literalinclude:: ../../../examples/series/series_shift.py
+       :language: python
+       :lines: 27-
```
akharche: same as 'astype'

sdc/datatypes/hpat_pandas_series_functions.py
1e-to: This is for developers who look at the code and not at the documentation.
But in the future, maybe in all functions, these comments will be removed

akharche: I guess that this code `:lines: 27-` may show "Expected result". Code for docs starts on 41 line

examples/series/series_var.py
AlexanderKalistratov: Lets specify it like '9.16666...'

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -612,19 +645,19 @@ def hpat_pandas_series_nsmallest(self, n=5, keep='first'):
     Returns
     -------
     :obj:`series`
-         returns :obj:`series`
+        returns :obj:`series`
```
PokhodenkoSA: I think that with new docs generation system it is not necessary to copy parameters and returns description from Pandas docs.
@samaid do you agree?

AlexanderKalistratov: @PokhodenkoSA so what the decision?

PokhodenkoSA: @samaid agree to remove it.

AlexanderKalistratov:
@akharche are you ok with this PR or not ok?

AlexanderKalistratov:
@PokhodenkoSA are you ok with this PR?

akharche:
@AlexanderKalistratov My comments have been ignored. I didn't see the generated docs of this PR, may be it's ok, so I don't mind to merge it

AlexanderKalistratov:
@1e-to please respond to Angelina

sdc/datatypes/hpat_pandas_series_functions.py
1e-to: It sounds like a good idea, but I think it’s worth doing in another PR, for all such examples.

PR #529, opened 2020-01-22
Implement Series.map()

PokhodenkoSA:


buildscripts/test.py
```diff
@@ -214,7 +214,10 @@
         failed_examples = []
         expected_failures = []
         unexpected_success = []
-        expected_failures_list = []
+        expected_failures_list = [
+            'series/series_map_dict.py',
```
PokhodenkoSA: @Vyacheslav-Smirnov could you propose to make file path which will work also on Windows?

Vyacheslav-Smirnov: Good question)
Looks like need to duplicate example with windows path ('series\series_map_dict.py')
or change a logic for these lists

PR #531, opened 2020-01-22
Cleaning up in test_series.py

kozlov-alexey:


sdc/tests/test_series.py
```diff
@@ -1322,7 +1272,26 @@ def test_impl(A, n):
                     S = pd.Series([2, 4, 6, 6, 3], index)
                     pd.testing.assert_series_equal(hpat_func(S, n), test_impl(S, n))
 
-    @unittest.expectedFailure
+    '''
```
densmirn: [PEP8](https://www.python.org/dev/peps/pep-0008/#string-quotes): For triple-quoted strings, always use double quote characters to be consistent with the docstring convention in PEP 257.

Please fix it here and in some places below.

kozlov-alexey: Sure, changed in all places.

AlexanderKalistratov:
@kozlov-alexey travis has failed

AlexanderKalistratov:
please add name to all getitem tests. As far as I see not all cases are covered with names

kozlov-alexey:
@AlexanderKalistratov 
I fixed travis and added names in tests for getitem/loc/iloc. Can be merged once CI passes.

AlexanderKalistratov:
@kozlov-alexey it seems that travis still failing?

kozlov-alexey:
@AlexanderKalistratov Yes, it seems I've been fooled by our tests (they passed on my local machine in old pipeline run). Now I see that for the rest 3 failing tests the reason is the same as for the others I skipped:
"During: resolving callee type: type(CPUDispatcher(<function get_series_data at 0x7f91f9506200>))"
Which is probably related to migration to latest Numba, so I have just skipped them.

AlexanderKalistratov:
@kozlov-alexey conflict

AlexanderKalistratov:
@kozlov-alexey cumulative sum for strings failed

PR #532, opened 2020-01-23
Add SDC_AUTO_PARALLEL=1 in benchmark tests

PokhodenkoSA:
SDC_AUTO_PARALLEL=1 is necessary to activate parallelization in overloads.
I have find a place where to set SDC_AUTO_PARALLEL=1. Please, propose, better place if you know.

It is also possible to set env var using features of internal CI. Please, show how to do it in internal CI.


PokhodenkoSA:
> It is possible to set environment parameters in internal CI. Build configuration settings - parameters - parameter type: environment.

Thanks.

PR #535, opened 2020-01-24
Numpy astype parallel

1e-to:


sdc/tests/tests_perf/array_bench.py
```diff
@@ -0,0 +1,30 @@
...
+def ref(a, b): 
+    start_time = time.time()
+    res = a.astype(np.int64)
+    finish_time = time.time()
+    print(res)
+    return finish_time - start_time  
+
+
+print(sdc(x, y))
+print(ref(x, y))
```
PokhodenkoSA:
```suggestion
print(ref(x, y))

```

sdc/tests/tests_perf/array_bench.py
```diff
@@ -0,0 +1,30 @@
+from numba import jit
+import numpy as np
+import time
+import random
+
+import pandas as pd
+from sdc.numpy_parallel_methods.astype import *
```
PokhodenkoSA: `import *` is bad style. Please, name all imported things explicitly.

sdc/tests/test_sdc_numpy.py
```diff
@@ -0,0 +1,171 @@
...
+
+        cases = [[5, 2, 0, 333, -4], [3.3, 5.4, np.nan]]
+        cases_type = [np.float64, np.int64]
+        for i in cases:
+            for j in cases_type:
+                a = np.array(i)
+                with self.subTest(data=i, type=j):
+                    np.testing.assert_array_equal(sdc_func(a, j), ref_impl(a, j))
+
+    @unittest.skip('Numba cant unbox literal value as type')
```
PokhodenkoSA:
```suggestion
    @unittest.skip('Numba can't unbox literal value as type')
```

sdc/tests/test_sdc_numpy.py
```diff
@@ -0,0 +1,171 @@
...
+
+        sdc_func = self.jit(sdc_impl)
+
+        cases = [[5, 2, 0, 333, -4], [3.3, 5.4, 3.4]]
+        cases_type = ['float64', 'int64']
+        for i in cases:
+            for j in cases_type:
+                a = np.array(i)
+                with self.subTest(data=i, type=j):
+                    np.testing.assert_array_equal(sdc_func(a, j), ref_impl(a, j))
```
PokhodenkoSA: Rename, please, `i` -> `case`, `j`-> `type_`

sdc/tests/test_sdc_numpy.py
```diff
@@ -0,0 +1,171 @@
...
+        def sdc_impl(a):
+            return astype(a, 'float64')
+
+        sdc_func = self.jit(sdc_impl)
+
+        cases = [[2, 3, 0], [4., 5.6, np.nan]]
+        for i in cases:
+            a = np.array(i)
+            with self.subTest(data=i):
+                np.testing.assert_array_equal(sdc_func(a), ref_impl(a))
```
PokhodenkoSA: Rename, please, `i` -> `case`

sdc/tests/test_sdc_numpy.py
```diff
@@ -0,0 +1,171 @@
...
+        def sdc_impl(a, t):
+            return astype(a, t)
+
+        sdc_func = self.jit(sdc_impl)
+
+        cases = [['a', 'cc', 'd'], ['3.3', '5', '.4'], ['¡Y', 'tú quién ', 'te crees']]
+        cases_type = [np.float64, np.int64]
+        for i in cases:
+            for j in cases_type:
+                a = np.array(i)
```
PokhodenkoSA: Why this source array is created inside `j` cycle and not in `i` cycle?

sdc/tests/test_sdc_numpy.py
```diff
@@ -0,0 +1,171 @@
...
+# -*- coding: utf-8 -*-
+import gc
+import glob
+import numpy as np
+import pandas as pd
+import platform
+import pyarrow.parquet as pq
+import re
+import sdc
+import unittest
```
PokhodenkoSA: Please, check that all this modules should be imported. I see that `re` is not used, is it?
The same with `gc`, `glob`, `platform`, `pq`.

sdc/numpy/sdc_numpy_modified_functions.py
```diff
@@ -0,0 +1,135 @@
...
+from sdc.datatypes.common_functions import TypeChecker
+from sdc.str_arr_ext import (StringArrayType, pre_alloc_string_array)
+from sdc.utils import sdc_overload
+
+
+def astype(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
+    pass
+
+
+@sdc_overload(astype, jit_options={'parallel': True})
```
PokhodenkoSA: @AlexanderKalistratov should we force `parallel=True` or should we rely on `sdc_overload`?

sdc/numpy/sdc_numpy_modified_functions.py
```diff
@@ -0,0 +1,135 @@
...
+def sdc_astype_overload(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
+    """
+    Intel Scalable Dataframe Compiler Developer Guide
+    *************************************************
+    Parallel analogue of numpy 'astype' implementation.
+
+    .. only:: developer
+
+       Test: python -m sdc.runtests -k sdc.tests.test_sdc_numpy.TestArrays.test_astype*
```
PokhodenkoSA:
```suggestion
       Test: python -m sdc.runtests sdc.tests.test_sdc_numpy -k astype
```

AlexanderKalistratov:
let's name new module like `functions` and probably submodule `numpy_like` or `numpy_tools` or no submodule

@PokhodenkoSA @kozlov-alexey what's your opinion?

AlexanderKalistratov:
@Hardcode84 and what's your opinion?

sdc/numpy/sdc_numpy_modified_functions.py
AlexanderKalistratov: we should rely on `sdc_overload`. No `jit_options` should be added

sdc/numpy/sdc_numpy_modified_functions.py
```diff
@@ -0,0 +1,135 @@
...
+from sdc.str_arr_ext import (StringArrayType, pre_alloc_string_array)
+from sdc.utils import sdc_overload
+
+
+def astype(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
+    pass
+
+
+@sdc_overload(astype, jit_options={'parallel': True})
+def sdc_astype_overload(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
```
AlexanderKalistratov: We don't need to completely reproduce numpy interface by the way

sdc/numpy/sdc_numpy_modified_functions.py
```diff
@@ -0,0 +1,135 @@
...
+
+            for i in range(arr_len):
+                item = self[i]
+                data[i] = str(item)  # TODO: check NA
+
+            return data
+
+        return sdc_astype_number_to_string_impl
+
+    if (isinstance(self, types.Array) and isinstance(dtype, types.functions.NumberClass)):
```
AlexanderKalistratov: I think you can unify these two cases - NumberClass and StringLiteral. It is different only by the way you extracting `other_numba_dtype`, implementation is the same

sdc/numpy/sdc_numpy_modified_functions.py
PokhodenkoSA: Yes. It is possible to use `@numba.extending.register_jitable` for creating reused functions.

AlexanderKalistratov: We need some kind of `sdc_register_jitable`. Also `register_jitable` doesn't care about types, and this is not what we want from `astype`. Our `astype` should do nothing if input and target types are the same

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+from sdc.str_arr_ext import (StringArrayType, pre_alloc_string_array)
+from sdc.utils import sdc_overload
+
+
+def astype(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
+    pass
+
+
+@sdc_overload(astype)
+def sdc_astype_overload(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
```
AlexanderKalistratov: Please remove unused parameters

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if not isinstance(self, types.Array):
+        return None
+
+    if not isinstance(dtype, (types.StringLiteral, types.UnicodeType, types.Function, types.functions.NumberClass)):
+        ty_checker.raise_exc(dtype, 'string or type', 'dtype')
+
+    if ((isinstance(dtype, types.Function) and dtype.typing_key == str) or
+        (isinstance(dtype, types.StringLiteral) and dtype.literal_value == 'str')):
+        def sdc_astype_number_to_string_impl(self, dtype, order='K', casting='unsafe', subok=True, copy=True):
```
AlexanderKalistratov: Here too

sdc/__init__.py
```diff
@@ -45,6 +45,7 @@
 import sdc.datatypes.hpat_pandas_series_rolling_functions
 import sdc.datatypes.hpat_pandas_seriesgroupby_functions
 import sdc.datatypes.hpat_pandas_stringmethods_functions
+import sdc.functions.numpy_like
```
AlexanderKalistratov: I don't think we need it here

AlexanderKalistratov:
@PokhodenkoSA I'm fine with this PR, but what we are going to do with perf tests?

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+        numba_self = self.dtype
+
+    if (isinstance(self, types.Array) and isinstance(dtype, (types.StringLiteral, types.UnicodeType))):
+        other_numpy_dtype = numpy.dtype(dtype.literal_value)
+        other_numba_dtype = numpy_support.from_dtype(other_numpy_dtype)
+        numba_self = self.dtype
+
+    def sdc_astype_number_impl(self, dtype):
+        if other_numba_dtype == numba_self:
+            return self
```
kozlov-alexey: I think we should still return a copy here, not the original array, as otherwise we would change the semantics of  the call itself and a user of astype won't be able to know whether modification are allowed on the returned data (if a new array was created) or not (if a reference to the old one was returned)

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+
+| This file contains SDC numpy modified functions, rewritten according to concurrency
+
+"""
+
+import numba
+import numpy
+
+from numba.errors import TypingError
+from numba import types
```
kozlov-alexey: It would be probably better to combine these lines into one

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+def astype(self, dtype):
+    pass
+
+
+@sdc_overload(astype)
+def sdc_astype_overload(self, dtype):
+    """
+    Intel Scalable Dataframe Compiler Developer Guide
+    *************************************************
+    Parallel analogue of numpy 'astype' implementation.
```
kozlov-alexey: That's not actually true, since we're changing the interface.
I would name it "Parallel replacement of numpy.astype"

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+            return data
+
+        return sdc_astype_number_to_string_impl
+
+    if (isinstance(self, types.Array) and isinstance(dtype, types.functions.NumberClass)):
+        other_numba_dtype = dtype.instance_type
+        numba_self = self.dtype
+
+    if (isinstance(self, types.Array) and isinstance(dtype, (types.StringLiteral, types.UnicodeType))):
+        other_numpy_dtype = numpy.dtype(dtype.literal_value)
```
kozlov-alexey: I think this should throw if dtype is actually types.UnicodeType since it won't have literal_value. You should provide different implementations for these. And also add tests that check dtype as non-literal string.

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,112 @@
...
+    if ((isinstance(dtype, types.Function) and dtype.typing_key == str) or
+        (isinstance(dtype, types.StringLiteral) and dtype.literal_value == 'str')):
+        def sdc_astype_number_to_string_impl(self, dtype):
+            num_chars = 0
+            arr_len = len(self)
+
+            # Get total chars for new array
+            for i in prange(arr_len):
+                item = self[i]
+                num_chars += len(str(item))  # TODO: check NA
```
kozlov-alexey: It appears that for unicode strings len returns number of chars, but not number of bytes, and allocating StringArray we need to know total number of bytes, not chars in it. Not sure if that's applicable for this case, but still it might be better to use get_utf8_size everywhere, just for consistency.

PR #536, opened 2020-01-27
Refactoring Series.setitem to support pd.Series as value and index based assignment

kozlov-alexey:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -434,56 +441,289 @@ def hpat_pandas_series_setitem(self, idx, value):
...
+
+                if idx_is_series == True:  # noqa
+                    _idx = idx._data
+                else:
+                    _idx = idx
+
+                if value_is_series == True:  # noqa
+                    _value = value._data
+                else:
+                    _value = value
```
densmirn: Don't you like shorter option?
```python
_idx = idx._data if idx_is_series == True else idx  # noqa
_value = value._data if value_is_series == True else value  # noqa
```

sdc/tests/test_series.py
```diff
@@ -5382,158 +5379,119 @@ def test_impl(S, val):
...
-            S[ind] = val
+        }
+        dtype_to_values = {
+            'int' :    50,
+            'float' :  np.nan,
+            'string' : 'bird'
+        }
+
+        idx = slice(2, None)
+        for dtype in dtype_to_data:
```
densmirn: Let's use method `items()` of `dict` here in other tests below because of using `dict[key]`.

sdc/tests/test_series.py
```diff
@@ -6175,6 +6133,596 @@ def test_impl(S):
...
+            return S
+        hpat_func = self.jit(test_impl)
+
+        S = pd.Series(['cat', '', 'bbb', '', 'a', None, 'a', '', None, 'b'])
+        idx, value = 0, 'dog'
+        result = hpat_func(S, idx, value)
+        pd.testing.assert_series_equal(result, S)
+
+    @skip_numba_jit('Requires StringArray support of operator.eq')
+    def test_series_setitem_idx_str_scalar(self):
```
densmirn: You could separate common part of tests `test_series_setitem_idx_str_scalar` and `test_series_setitem_idx_int_scalar1` something like this:
```python
    def _test_series_setitem_idx_scalar(self, S, idx):
        """ Verifies Series.setitem for scalar idx operand and integer Series with index of matching dtype"""
        def test_impl(A, i, value):
            A[i] = value

        values_to_test = [-100,
                          np.array([5, 55, 555]),
                          pd.Series([5, 55, 555])]

        for value in values_to_test:
            with self.subTest(value=value):
                S1 = S.copy(deep=True)
                S2 = S.copy(deep=True)
                hpat_func(S1, idx, value)
                test_impl(S2, idx, value)
                pd.testing.assert_series_equal(S1, S2)

    def test_series_setitem_idx_str_scalar(self):
        idx = 'a'
        S = pd.Series(np.arange(5), index=['a', 'a', 'c', 'd', 'a'])
        self._test_series_setitem_idx_scalar(S, idx)

    def test_series_setitem_idx_int_scalar1(self):
        idx = 8
        S = pd.Series(np.arange(5), index=[8, 8, 9, 8, 5])
        self._test_series_setitem_idx_scalar(S, idx)
```

Maybe you could do that for other tests if it's possible.

sdc/tests/test_series.py
```diff
@@ -6175,6 +6133,596 @@ def test_impl(S):
...
+
+    def test_series_setitem_idx_int_scalar2(self):
+        """ Verifies Series.setitem for scalar integer idx operand and integer Series with
+            index of non matching dtype (i.e. set along positions, not index)"""
+        def test_impl(A, i, value):
+            A[i] = value
+        hpat_func = self.jit(test_impl)
+
+        n = 11
+        np.random.seed(0)
```
densmirn: Look like you don't need to set up seed in this test.

kozlov-alexey: Right, thanks! I removed it here and in other places where there are no calls to np.random.

sdc/tests/test_series.py
kozlov-alexey: Changed here and in the test below.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Thank you, it's better this way.

sdc/tests/test_series.py
kozlov-alexey: Yes, thank you, good suggestion. I added _test_series_setitem for this.

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: Fixed.

sdc/tests/test_series.py
```diff
@@ -6252,6 +6189,482 @@ def test_impl(A, B):
...
+        """ Common function used by setitem tests to compile and run setitem on provided data"""
+        def test_impl(A, i, value):
+            A[i] = value
+        hpat_func = self.jit(test_impl)
+
+        for series_data in all_data:
+            for series_index in indexes:
+                S = pd.Series(series_data, series_index, dtype=dtype)
+                for idx in idxs:
+                    for value in values:
```
densmirn: You could use `itertools.product` to combine the loops in the future. Now we can skip this.

sdc/tests/test_utils.py
```diff
@@ -193,3 +193,30 @@ def skip_inline(msg_or_func):
         wrapper = unittest.expectedFailure
     # wrapper = lambda f: f  # disable skipping
     return wrapper(func) if func else wrapper
+
+
+def take_k_elements(k, data):
+    random_idx = np.arange(len(data))
+    np.random.shuffle(random_idx)
+    return np.asarray(data).take(random_idx[:k])
```
densmirn: Should we set up seed here and in below function? Maybe pass seed to the functions as an argument.

sdc/tests/test_series.py
kozlov-alexey: Will be done in #582

sdc/tests/test_utils.py
kozlov-alexey: Yes, I'll add it.

PR #541, opened 2020-01-29
Scale min/max(skipna=True)

PokhodenkoSA:
I think nanmax should be moved as separate function like in Elena's astype(). Do you agree?


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4114,7 +4114,25 @@ def hpat_pandas_series_max_impl(self, axis=None, skipna=None, level=None, numeri
...
+                    break
+            else:
+                return result
+
+            for i in prange(length):
+                value = input_arr[i]
+                if not numpy.isnan(value):
+                    result = max(result, input_arr[i])
+
+            return result
```
1e-to: Could it be better?
`mask = numpy.isnan(self._data)
data_no_nan = self._data[~mask]
return data_no_nan.max()`

PokhodenkoSA: It should be benchmarked.

PokhodenkoSA: My version:
image
@1e-to version:
image

PokhodenkoSA: > Could it be better?
> `mask = numpy.isnan(self._data) data_no_nan = self._data[~mask] return data_no_nan.max()`
```python
            mask = numpy.isnan(self._data)
            data_no_nan = self._data[~mask]
            return data_no_nan.max()
```

AlexanderKalistratov: It is expectedly that single parfor is faster. But Elena's version shows poor scalability. This is most likely due to `isnan` is not parallelized.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4031,6 +4031,28 @@ def hpat_pandas_series_min_impl(self, axis=None, skipna=None, level=None, numeri
...
+
+    # set first element
+    result = numpy.nan
+    for i in range(length):
+        value = a[i]
+        if not numpy.isnan(value):
+            result = value
+            break
+    else:
+        return result
```
AlexanderKalistratov: doesn't looks good. In case if all values are `nan` this function wouldn't scale at all.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4031,6 +4031,28 @@ def hpat_pandas_series_min_impl(self, axis=None, skipna=None, level=None, numeri
     return hpat_pandas_series_min_impl
 
 
+@numba.extending.register_jitable(parallel=True)
+def numpy_nanmax(a):
+    length = len(a)
+
+    # set first element
+    result = numpy.nan
```
AlexanderKalistratov: That wouldn't work for ints

sdc/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: @AlexanderKalistratov I will check `isnan`. I like this small code :)
What is `_data[~mask]`. Is it parallelized?

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: There are utility functions in numpy which returns maximum/minimum values for different types:
```
min_float64 = np.finfo('float64').min
max_float64 = np.finfo('float64').max
min_int64 = np.iinfo('int64').min
max_int64 = np.iinfo('int64').max
```
And for the rest. We need to implement utility function which would return min/max value based on variable type.
Please also note that we can't the same value for all int types - result type should be of the same type as array. I think so.

For floats we should use `nan`. I think

sdc/datatypes/hpat_pandas_series_functions.py
AlexanderKalistratov: I think it is. But I'm not sure.

The small code looks better, but it uses intermediate array and it would be slower anyway and uses more memory..
I'm afraid this time we should sacrifice readability to achieve better performance

sdc/datatypes/hpat_pandas_series_functions.py
PokhodenkoSA: I will improve tests to catch it.

sdc/functions/numpy_like.py
```diff
@@ -0,0 +1,58 @@
...
+| This file contains SDC numpy modified functions, rewritten according to concurrency
+
+"""
+
+import numpy
+from numba import prange
+from numba.extending import register_jitable
+
+
+@register_jitable(parallel=True, inline='always')
```
AlexanderKalistratov: So, does inline actually works?

sdc/functions/numpy_like.py
```diff
@@ -97,3 +97,30 @@ def sdc_astype_number_impl(self, dtype):
...
         return sdc_astype_number_impl
+
+
+def nan_min_max_factory(reduce_op):
+    def impl(a):
+        length = len(a)
+
+        # set first element
+        result = None
+        for i in range(length):
```
AlexanderKalistratov: I still don't like this solution.

1. It is not scalable
2. It most likely would prevent loops fusing
3. It is big and ugly
4. It is unnecessary for ints

sdc/functions/numpy_like.py
```diff
@@ -97,3 +97,30 @@ def sdc_astype_number_impl(self, dtype):
...
+            if not numpy.isnan(v):
+                result = v
+                break
+        else:
+            return numpy.nan
+
+        for i in prange(length):
+            v = a[i]
+            if not numpy.isnan(v):
+                result = reduce_op(result, v)
```
AlexanderKalistratov: I like it, but let's check it doesn't affect performance

sdc/functions/numpy_like.py
AlexanderKalistratov: For max case you could set initial value to min_int, min_float. It is one line and you wouldn't need to do this loop. Though, in case of floats you need to somehow detect no value case

PokhodenkoSA:
image

AlexanderKalistratov:
@PokhodenkoSA conflict

PR #552, opened 2020-01-30
Add decorator sdc_register_jitable

PokhodenkoSA:
This PR adds decorator `sdc_register_jitable` which wraps `numba.extending.register_jitable` decorator.
`sdc_register_jitable` takes default parameters for `parallel` and `inline` from `SDC_AUTO_PARALLEL` and `SDC_AUTO_INLINE` environment variables.

I am planning to use it in #541 and #535 and other PRs for improving scalability.

### DIFF
--- a/sdc/utils.py
+++ b/sdc/utils.py
@@ -594,7 +594,7 @@ def sdc_overload(func, jit_options={}, parallel=None, strict=True, inline=None):
 
 def patched_register_jitable(*args, **kwargs):
     """
-    register_jitable patched according this:
+    register_jitable patched according to this:
     https://github.com/numba/numba/issues/5142#issuecomment-579704346
     """
     def wrap(fn):
--- a/sdc/utils.py
+++ b/sdc/utils.py
@@ -46,6 +46,7 @@ from sdc.config import (config_use_parallel_overloads, config_inline_overloads)
 from enum import Enum
 import types as pytypes
 from numba.extending import overload, overload_method, overload_attribute
+from numba.extending import register_jitable
 
 
 # int values for types to pass to C code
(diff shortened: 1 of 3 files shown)


PokhodenkoSA:
Replaced all `register_jitable` in SDC with `sdc_register_jitable`.
Most `@register_jitable` was used in @densmirn series rolling functions. @densmirn is it ok to replace them?

densmirn:
> Replaced all `register_jitable` in SDC with `sdc_register_jitable`.
> Most `@register_jitable` was used in @densmirn series rolling functions. @densmirn is it ok to replace them?

I think so.

sdc/utils.py
```diff
@@ -591,6 +592,20 @@ def sdc_overload(func, jit_options={}, parallel=None, strict=True, inline=None):
     return overload(func, jit_options=jit_options, strict=strict, inline=inline)
 
 
+def sdc_register_jitable(*args, **kwargs):
+    updated_kwargs = kwargs.copy()
+    updated_kwargs['parallel'] = updated_kwargs.get('parallel', config_use_parallel_overloads)
+    updated_kwargs['inline'] = updated_kwargs.get('inline', 'always' if config_inline_overloads else 'never')
```
AlexanderKalistratov: So, does it work?

AlexanderKalistratov: I'm 90% sure it doesn't

PokhodenkoSA: Yes. If I use e.g. `Always` it prints this:
image

PokhodenkoSA: @AlexanderKalistratov could you propose how to check precisely?

AlexanderKalistratov: The best way is to check IR with enabled/disable option

AlexanderKalistratov: It seems, it doesn't work
https://github.com/numba/numba/issues/5142

PokhodenkoSA: I have applied proposal patch from the issue.

PR #559, opened 2020-02-01
Overload df.getitem for str idx

densmirn:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -927,6 +928,121 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
...
+    """Generate main code lines for df.getitem"""
+    try:
+        col_idx = self.columns.index(idx)
+    except ValueError:
+        func_lines = ['  raise KeyError']
+    else:
+        col_data = f'data_{col_idx}'
+        func_lines = [
+            f'  {col_data} = get_dataframe_data(self, {col_idx})',
+            f'  return pandas.Series({col_data}, index=self._index, name=idx)'
```
akharche: What if self._index is None? Is it ok for Series constructor?
Attribute index is supported for DataFrames, it handles 'no index' cases
```suggestion
            f'  return pandas.Series({col_data}, index=self.index, name=idx)'
```

AlexanderKalistratov: @akharche no. self._index is better

AlexanderKalistratov: It is ok to provide None as index to series

AlexanderKalistratov: Otherwise we would create index on each getitem. This is definitely something we don't want to happen

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -927,6 +929,86 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
...
+    exec(func_text, global_vars, loc_vars)
+    _impl = loc_vars['_df_getitem_str_literal_idx_impl']
+
+    return _impl
+
+
+def gen_df_getitem_unicode_idx_impl(self):
+    df_is_empty = False if self.columns else True
+
+    def _df_getitem_unicode_idx_impl(self, idx):
```
AlexanderKalistratov: I'd suggest this:
```python
def _df_getitem_unicode_idx_impl(self, idx):
    idx_ = literally(idx)
    return self[idx]
```

PR #562, opened 2020-02-03
Update SDC build instruction in Readme.rst

Vyacheslav-Smirnov:
Build instruction in Readme.rst is not reflect to current SDC build process.
 - Add step of specific commit Numba build
 - Correct packages versions
 - Add missed packages required for Numba build


README.rst
```diff
@@ -47,23 +50,33 @@ Building on Linux with conda-build
 ::
 
     PYVER=<3.6 or 3.7>
+    NUMPYVER=<1.16 or 1.17>
     conda create -n CBLD python=$PYVER conda-build
     source activate CBLD
-    git clone https://github.com/IntelPython/sdc
+    git clone https://github.com/IntelPython/sdc.git
```
PokhodenkoSA: Change also line 71.

README.rst
```diff
@@ -47,23 +50,33 @@ Building on Linux with conda-build
...
+    conda build --python $PYVER --numpy $NUMPYVER -c file://<path_to_sdc>/numba_build -c numba -c defaults -c conda-forge --override-channels buildscripts/sdc-conda-recipe
 
 Building on Linux with setuptools
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ::
 
     PYVER=<3.6 or 3.7>
-    conda create -n SDC -q -y -c numba -c defaults -c conda-forge python=$PYVER numba pandas scipy mpich pyarrow=0.15.1 gcc_linux-64 gxx_linux-64
+    NUMPYVER=<1.16 or 1.17>
+    conda create -n SDC -q -y -c numba -c defaults -c intel -c conda-forge python=$PYVER numpy=$NUMPYVER pandas=0.25.3 scipy pyarrow=0.15.1 gcc_linux-64 gxx_linux-64 tbb-devel llvmlite=0.31.0rc1=py*_0
```
PokhodenkoSA: llvmlite is a dependency of numba. Why not remove it from the list?

README.rst
```diff
@@ -85,23 +98,35 @@ Building on Windows with conda-build
...
     conda activate SDC
     git clone https://github.com/IntelPython/sdc.git
     cd sdc
     set INCLUDE=%INCLUDE%;%CONDA_PREFIX%\Library\include
     set LIB=%LIB%;%CONDA_PREFIX%\Library\lib
-    %CONDA_PREFIX%\Library\bin\mpivars.bat quiet
+    # Build Numba
+    git clone https://github.com/numpy/numpy.git
+    cd numba
+    git checkout ef119bcd1733ff49d71bdf2da8a66e91bb704f83
```
PokhodenkoSA: This magic constant is commit number. Could you, please, create variable with name explaining what is specific with this commit?

README.rst
```diff
@@ -85,23 +98,35 @@ Building on Windows with conda-build
...
+    python setup.py install
+    # Build Intel SDC
+    cd ..\
```
PokhodenkoSA:
```suggestion
    cd ..
```

README.rst
Vyacheslav-Smirnov: We need specific llvmlite version for Numba

README.rst
Vyacheslav-Smirnov: Do not think it is required to create a variable for checkouting Numba hash
I just mentioned  that
"Intel SDC uses Numba ``ef119bcd1733ff49d71bdf2da8a66e91bb704f83`` commit from master branch as Numba package for build and run.
That is why it is required to build specified Numba first. Build steps are described below."
Do you think there should be described features which we are using?

README.rst
Vyacheslav-Smirnov: Will update

README.rst
Vyacheslav-Smirnov: Also here we build numba via setuptools thus need to install its dependencies

README.rst
```diff
@@ -47,23 +50,33 @@ Building on Linux with conda-build
...
     source activate SDC
-    git clone https://github.com/IntelPython/sdc
-    cd sdc
+    # Build Numba
+    git clone https://github.com/numba/numba.git
+    cd numba
+    git checkout ef119bcd1733ff49d71bdf2da8a66e91bb704f83
+    python setup.py install
     # build SDC
+    cd ../
```
PokhodenkoSA:
```suggestion
    cd ..
```

README.rst
PokhodenkoSA: When we will update Numba version we also will have to check new dependencies and modify this line. Is there more convenient way to install numba dependencies?

Also, I think we now could use newer version of llvmlite.

README.rst
Vyacheslav-Smirnov: Updated

Vyacheslav-Smirnov:
When we updating Numba we should always check the documentation for update.
I just want to make it more clear and relevant to current situation - in Numba recipe it is defined llvmlite=0.31.0, thus the user who wants to build Numba using setuptools should install the same llvmlite (for example, conda resolver may install less or more version if version is not specified and Numba can be failed to build)

AlexanderKalistratov:
@PokhodenkoSA ping

PokhodenkoSA:
@Vyacheslav-Smirnov could you please describe improvements of this PR in PR description? It is not clear for me what is the reason of this PR.

Vyacheslav-Smirnov:
@PokhodenkoSA, PR description updated.

PR #563, opened 2020-02-04
Move warming up from all benchmarks suites to get_times() function.

PokhodenkoSA:
So Python branches also will be warmed up.


sdc/tests/tests_perf/test_perf_utils.py
```diff
@@ -182,6 +182,10 @@ def calc_compilation(pyfunc, *args, iter_number=5):
 
 def get_times(f, *args, iter_number=5):
     """Get time of boxing+unboxing and internal execution"""
+
+    # Warming up
+    f(*args)
```
Hardcode84: Move this as close to iter loop as possible (after `exec_times` and `boxing_times` creation), LGTM otherwise

PR #566, opened 2020-02-04
Overload df.getitem with slice idx

densmirn:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -929,81 +929,95 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
     return sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns)
 
 
-def df_getitem_str_literal_idx_main_codelines(self, idx):
+def df_getitem_slice_idx_main_codelines(self, idx):
```
AlexanderKalistratov: So... this should work not only for slice, write? For any idx. E.g. for array of booleans (filter)?

sdc/tests/test_dataframe.py
```diff
@@ -1244,43 +1248,62 @@ def test_impl(df):
...
         sdc_func = self.jit(test_impl)
+        df = gen_df_int_cols(test_global_input_data_float64)
 
-        for df in [gen_df(test_global_input_data_float64), pd.DataFrame()]:
-            with self.subTest(df=df):
-                with self.assertRaises(KeyError):
-                    sdc_func(df, 'ABC')
+        pd.testing.assert_series_equal(sdc_func(df), test_impl(df))
+
+    @unittest.skip('DF.getitem unsupported idx as a tuple')
```
AlexanderKalistratov: Are you planning to add this functionality later?

AlexanderKalistratov:
Looks very good

sdc/datatypes/hpat_pandas_dataframe_functions.py
densmirn: Yes, it should work only for slice.

sdc/tests/test_dataframe.py
densmirn: Now I'm not able to get values of a tuple as a literal. So I cannot get data of DF via `get_frame_data`. I will support it when will be able.

sdc/datatypes/hpat_pandas_dataframe_functions.py
densmirn: I avoided such code generation for `StringLiteral` and `Unicode` types.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -929,81 +929,95 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
...
-            f'  {col_data} = get_dataframe_data(self, {col_idx})',
-            f'  return pandas.Series({col_data}, index=self._index, name=idx)'
+        func_lines = ['  index = self._index']
+
+    results = []
+    for i, col in enumerate(self.columns):
+        res_data = f'result_data_{i}'
+        func_lines += [
+            f'  data_{i} = get_dataframe_data(self, {i})',
+            f'  {res_data} = pandas.Series(data_{i}[idx], index=index[idx], name="{col}")'
```
kozlov-alexey: It might be better to compute index[idx] once and just use it in these lines, to avoid repetitive calls.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -929,81 +929,95 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
...
 
 
-def df_getitem_str_literal_idx_codegen(self, idx):
+def df_getitem_str_slice_codegen(self, idx):
     """
-        def _df_getitem_str_literal_idx_impl(self, idx):
-          data_0 = get_dataframe_data(self, 0)
-          return pandas.Series(data_0, index=self._index, name=idx)
+    def _df_getitem_slice_idx_impl(self, idx):
+      index = self._index
```
kozlov-alexey: This comment is a bit incorrect, since index is defined this way only if it's not None.
```suggestion
      index = ...
```

Otherwise we might add overload for DataFrame.index similar to Series.index and use it in a generated func.

densmirn: The idea was to provide just an example of generated implementation in docstring. So I will update the docstring to say about that.
I don't mind the fact that we might add such overload, but I vote for doing it in separate PR.

densmirn: @kozlov-alexey what do you think about that?

sdc/tests/test_dataframe.py
densmirn: I gonna support it in separate PR, at the least try to do that.

sdc/datatypes/hpat_pandas_dataframe_functions.py
densmirn: I gonna check it works for any `idx` in separate PR.

sdc/datatypes/hpat_pandas_dataframe_functions.py
kozlov-alexey: @densmirn I don't mind, let's do it in a separate one.

PR #567, opened 2020-02-04
Fix astype in operators series

1e-to:


sdc/sdc_autogenerated.py
```diff
@@ -308,7 +309,7 @@ def _series_operator_sub_common_impl(self, other):
                     result_data[:] = self._data - other._data
 
                     if none_or_numeric_indexes == True:  # noqa
-                        result_index = left_index.astype(numba_index_common_dtype)
+                        result_index = left_index.astype(numba_n_dtypeleft_index, )
```
kozlov-alexey: It's probably a typo here, and moreover np.astype call still remains in this and below lines

sdc/sdc_autogenerated.py
```diff
@@ -414,7 +415,7 @@ def _series_operator_mul_scalar_impl(self, other):
             def _series_operator_mul_none_indexes_impl(self, other):
 
                 if (len(self._data) == len(other._data)):
-                    result_data = self._data.astype(numpy.float64)
+                    result_data = astype(self._data, numpy.float64)
                     result_data = result_data * other._data
```
kozlov-alexey: We might have it as
```suggestion
                    result_data *= other._data
```

PR #573, opened 2020-02-06
Overload df.getitem with idx of tuple

densmirn:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -978,14 +1003,52 @@ def _df_getitem_slice_idx_impl(self, idx):
...
+    key_error = any(i not in self.columns for i in literal_idx)
 
-    loc_vars = {}
-    exec(func_text, global_vars, loc_vars)
-    _impl = loc_vars['_df_getitem_slice_idx_impl']
+    if self.columns and not key_error:
+        func_lines += df_getitem_tuple_idx_main_codelines(self, literal_idx)
+    else:
+        # raise KeyError if input DF is empty or idx is invalid
+        func_lines += ['  raise KeyError']
```
```suggestion
        func_lines += ['  raise KeyError("Index is not in the DataFrame")']
```

sdc/tests/test_dataframe.py
```diff
@@ -1236,6 +1236,17 @@ def test_impl(df, start, end):
...
         pd.testing.assert_frame_equal(jit_result, ref_result)
 
+    @skip_sdc_jit('DF.getitem unsupported Series name')
+    def _test_df_getitem_tuple_idx(self, df):
+        def test_impl(df):
+            # pd.df.getitem does not support idx as a tuple
+            return df[['A', 'C']]
+
+        # SDC pd.df.getitem does not support idx as a list
+        sdc_func = self.jit(lambda df: df[('A', 'C')])
```
kozlov-alexey: I think we should state this as a limitation in our documentation too.
Also it might look better if you write it this way:
```
        test_impl = lambda df: df[['A', 'C']]
        sdc_func = self.jit(lambda df: df[('A', 'C')])
```
or use do_jit parameter, e.g.
 ```
       # Pandas implementation does not support idx as a tuple
        # but SDC implementation only supports tuple, so adjust tested code
        def test_impl(df, do_jit):
            if do_jit == True:  # noqa
                return df[('A', 'C')]
            else:
                return df[['A', 'C']]

        pd.testing.assert_frame_equal(sdc_func(df, True), test_impl(df, False))
```

File: sdc/datatypes/hpat_pandas_dataframe_functions.py (hunk 1 above)
Comment (kozlov-alexey) [inline sdc/datatypes/hpat_pandas_dataframe_functions.py]: The exception should really have a message, maybe something like "Column is not in the DataFrame" would fit?

File: sdc/datatypes/hpat_pandas_dataframe_functions.py [hunk 3]
```diff
@@ -978,14 +1003,52 @@ def _df_getitem_slice_idx_impl(self, idx):
...
+
+    func_text = '\n'.join(func_lines)
+    global_vars = {'pandas': pandas, 'numpy': numpy,
+                   'get_dataframe_data': get_dataframe_data}
+
+    return func_text, global_vars
+
+
+def gen_df_getitem_impl_generator(codegen, impl_name):
```
kozlov-alexey: This looks as a very generic function, that we could use everywhere else, so it might be useful to move it to sdc_typing_utils. So it's responsibility would be to call exec on provided func_code and return back created func object.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1018,8 +1081,11 @@ def _df_getitem_unicode_idx_impl(self, idx):
 
         return _df_getitem_unicode_idx_impl
 
+    if isinstance(idx, types.Tuple):
```
kozlov-alexey: This should probably be a more strict check, e.g.
```suggestion

    if (isinstance(idx, types.Tuple)
        and all([isinstance(item, types.StringLiteral) for item in idx])):
```

PR #574, opened 2020-02-06
Fix astype numpy-like unliteral

1e-to:


sdc/tests/test_sdc_numpy.py
```diff
@@ -80,7 +62,7 @@ def ref_impl(a):
             return a.astype('float64')
 
         def sdc_impl(a):
-            return astype(a, 'float64')
+            return numpy_like.astype(a, 'float64')
```
AlexanderKalistratov: Do we need this test? Isn't it covered by `test_astype_to_num`?

sdc/tests/test_sdc_numpy.py
```diff
@@ -95,7 +77,7 @@ def ref_impl(a):
             return a.astype(np.int64)
 
         def sdc_impl(a):
-            return astype(a, np.int64)
+            return numpy_like.astype(a, np.int64)
```
AlexanderKalistratov: And this one?

sdc/tests/test_sdc_numpy.py
1e-to: This tests the parameter that is passed from the jitted region

PR #575, opened 2020-02-06
capitalize, casefold, title, swapcase with use prange & work with None

Rubtsowa:


sdc/hiframes/api.py
```diff
@@ -570,7 +570,10 @@ def get_nan_mask(arr):
 def get_nan_mask_overload(arr):
 
     def get_nan_mask_via_isna_impl(arr):
-        return np.array([isna(arr, i) for i in np.arange(len(arr))])
+        res = np.empty(len(arr), dtype=np.bool_)
```
densmirn: `len(arr)` could be assigned to a separate variable because of twice usage.

sdc/tests/test_series.py
```diff
@@ -3168,6 +3167,14 @@ def test_series_str2str(self):
                 pd.testing.assert_series_equal(hpat_func(S), test_impl(S),
                                                check_names=check_names)
 
+    def test_series_capitalize_str_with_none(self):
```
densmirn: Looks like you could combine tests `test_series_capitalize_str_with_none` and `test_series_capitalize_str` somehow. That also can be applied for other added tests.

PR #578, opened 2020-02-06
Scale Series.count()

PokhodenkoSA:
image


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -5138,8 +5138,11 @@ def hpat_pandas_series_count_impl(self, level=None):
         Returns number of unique elements in the object
         Test: python -m sdc.runtests sdc.tests.test_series.TestSeries.test_series_count
         """
-        data_no_nan = self._data[~numpy.isnan(self._data)]
-        return len(data_no_nan)
+        result = 0
```
AlexanderKalistratov: Lets have special case for ints which simply return len

PR #579, opened 2020-02-06
Scale Series.median(skipna=True/False)

PokhodenkoSA:
image
Bad performance.


sdc/functions/numpy_like.py
```diff
@@ -97,3 +99,32 @@ def sdc_astype_number_impl(self, dtype):
...
+
+@sdc_overload(nanmedian)
+def np_nanmedian(a):
+    if not isinstance(a, types.Array):
+        return
+    isnan = get_isnan(a.dtype)
+
+    def nanmedian_impl(a):
+        # Create a temporary workspace with only non-NaN values
+        temp_arry = np.empty(a.size, a.dtype)
```
AlexanderKalistratov: Result of median always should be float

sdc/functions/numpy_like.py
```diff
@@ -97,3 +99,32 @@ def sdc_astype_number_impl(self, dtype):
...
+
+def nanmedian(a):
+    pass
+
+
+@sdc_overload(nanmedian)
+def np_nanmedian(a):
+    if not isinstance(a, types.Array):
+        return
+    isnan = get_isnan(a.dtype)
```
AlexanderKalistratov: Looks good. Let's use it everywhere

AlexanderKalistratov: After some investigation, i see few issues with it:

1. It return `False` for ints, but the function itself doesn't inline
2. Even if it would inline, there is no dead code elimination pass after inline pass. Which means, the condition would not be eliminated
3. Even if dead code elimination would be added in previous versions it works only with conditions like `if True:` and `if False:` and doesn't work with conditions like `if not False`.

Probably llvm would be able to fix all this issues in it's own IR. Probably not. Need to check and measure

Still, let's stick to the strategy use this function

PR #580, opened 2020-02-07
Argmin/argmax + nanargmin/nanargmax

1e-to:
image


sdc/functions/numpy_like.py
```diff
@@ -97,3 +112,193 @@ def sdc_astype_number_impl(self, dtype):
...
+    if isinstance(dtype, types.Float):
+        max_ref = max_dtype_float_val(dtype)
+
+    if not isinstance(self, types.Array):
+        return None
+
+    if isinstance(dtype, types.Number):
+        def sdc_nanargmin_impl(self):
+            res = max_ref
+            position = max_
```
AlexanderKalistratov: What is `max_`?

sdc/functions/numpy_like.py
```diff
@@ -97,3 +112,193 @@ def sdc_astype_number_impl(self, dtype):
...
+            length = len(self)
+            for i in prange(length):
+                if min(res, self[i]) == self[i]:
```
AlexanderKalistratov: That looks weird.
```python
if not isnan(self[i]):
    res = min(res, self[i])
    if res == self[i]:
        position = min(position, i)
```

AlexanderKalistratov:
Python version is 10000 times faster on one thread? Am I getting it right?

AlexanderKalistratov:
@PokhodenkoSA I can't find previous measurements of argmax/argmin anywhere. Do we have it?

1e-to:
> Python version is 10000 times faster on one thread? Am I getting it right?

Yes, it is, perhaps the python finds and returns the first nan by function min (x,y)

> I can't find previous measurements of argmax/argmin anywhere. Do we have it?

Now we use argmin that named "Numba" at the table of results

sdc/functions/numpy_like.py
```diff
@@ -97,3 +112,193 @@ def sdc_astype_number_impl(self, dtype):
...
+    if not isinstance(self, types.Array):
+        return None
+
+    if isinstance(dtype, types.Number):
+        def sdc_argmin_impl(self):
+            res = max_ref
+            position = max_
+            length = len(self)
+            for i in prange(length):
+                if not isnan(self[i]):
```
AlexanderKalistratov: Let's try it this way:
```python
result_is_nan = False
if not isnan(self[i]):
    if not result_is_nan:
        res = min(res, self[i])
        if res == self[i]:
            position = min(position, i)
else:
    position = min(position, i)
    result_is_nan = True
```

AlexanderKalistratov:
> > Python version is 10000 times faster on one thread? Am I getting it right?
> 
> Yes, it is, perhaps the python finds and returns the first nan by function min (x,y)

Can we test also without `nans`?

sdc/functions/numpy_like.py
AlexanderKalistratov: No, we need different position for nan result and for not nan result:
```python
result_is_nan = False
nanposition = max_
if not isnan(self[i]):
    if not result_is_nan:
        res = min(res, self[i])
        if res == self[i]:
            position = min(position, i)
else:
    nanposition = min(nanposition, i)
    result_is_nan = True

File: sdc/functions/numpy_like.py [hunk 4]
```diff
@@ -130,7 +150,278 @@ def sdc_astype_number_impl(self, dtype):
...
+        def sdc_nanargmax_impl(self):
+            chunks = get_chunks(len(self))
+            arr_res = numpy.empty(shape=len(chunks), dtype=dtype)
+            arr_pos = numpy.empty(shape=len(chunks), dtype=numpy.int64)
+            for i in prange(len(chunks)):
+                chunk = chunks[i]
+                res = min_ref
+                pos = max_int64
+                for j in range(chunk.start, chunk.stop):
+                    if max(res, self[j]) == self[j]:
```
AlexanderKalistratov: Since we are using chunks now, can't we simplify it just to:
```python
if self[j] > res:
    res = self[j]
    pos = j
```

AlexanderKalistratov:
Could you please think if you can generalize argmin/argmax implementation in the same way it was done here: https://github.com/IntelPython/sdc/pull/541

@PokhodenkoSA has details

AlexanderKalistratov:
any performance numbers?

1e-to:
> any performance numbers?
Skipna = True and False has equal results:
image

sdc/tests/tests_perf/test_perf_numpy.py
```diff
@@ -85,6 +85,24 @@ def _test_case(self, cases, name, total_data_length, data_num=1, input_data=test
         CE(type_='Numba', code='data.astype(np.int64)', jitted=True),
         CE(type_='SDC', code='sdc.functions.numpy_like.astype(data, np.int64)', jitted=True),
     ], usecase_params='data'),
+    TC(name='nanargmin', size=[10 ** 7], call_expr=[
```
PokhodenkoSA: Why did not you add cases for Numba for `np.nanargmin` and `np.nanargmax`?

1e-to: Because its not exist. Numba dont have this methods

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2975,22 +2975,63 @@ def hpat_pandas_series_idxmax(self, axis=None, skipna=True):
...
+            # return numpy.argmax(self._data)
+            if skipna is None:
+                _skipna = True
+            else:
+                _skipna = skipna
+
+            if _skipna:
+                return numpy_like.nanargmax(self._data)
+
+            return numpy_like.argmax(self._data)
```
densmirn: I would create variable `none_index = isinstance(self.index, types.NoneType) or self.index is None`, then make common implementations `hpat_pandas_series_idxmax_impl` and `hpat_pandas_series_idxmax_impl` for both cases with index and without index. In the implementations I would add something like that:
```python
if none_index == True:  # noqa
    return result
else:
    self._index[int(result)]
```

The same is applicable for `idxmin`.

sdc/functions/numpy_like.py
```diff
@@ -131,7 +151,165 @@ def sdc_astype_number_impl(self, dtype):
...
+                    res = initial_result
+                    pos = max_int64
+                    for j in range(chunk.start, chunk.stop):
+                        if reduce_op(res, self[j]) == self[j]:
+                            if not isnan(self[j]):
+                                if res == self[j]:
+                                    pos = min(pos, j)
+                                else:
+                                    pos = j
+                                    res = self[j]
```
densmirn: Let's decrease indentations:
```python
if reduce_op(res, self[j]) != self[j]:
    continue
if isnan(self[j]):
    continue
if res == self[j]:
    pos = min(pos, j)
else:
    pos = j
    res = self[j]
```

sdc/functions/numpy_like.py
```diff
@@ -131,7 +151,165 @@ def sdc_astype_number_impl(self, dtype):
...
+                    arr_res[i] = res
+                    arr_pos[i] = pos
+
+                general_res = initial_result
+                general_pos = max_int64
+                for i in range(len(chunks)):
+                    if reduce_op(general_res, arr_res[i]) == arr_res[i]:
+                        if general_res == arr_res[i]:
```
densmirn: Let's decrease indentations.

sdc/tests/test_sdc_numpy.py
```diff
@@ -240,6 +240,66 @@ def sdc_impl():
...
         np.testing.assert_array_equal(sdc_func(), ref_impl())
 
+    def test_argmin(self):
+        def ref_impl(a):
+            return np.argmin(a)
+
+        def sdc_impl(a):
+            return numpy_like.argmin(a)
+
+        sdc_func = self.jit(sdc_impl)
```
densmirn: Let's combine that to:
```python
@self.jit
def sdc_impl(a):
    return numpy_like.argmin(a)
```

This change is acceptable for other tests.

AlexanderKalistratov:
@1e-to conflict

PR #581, opened 2020-02-07
numpy-like copy

1e-to:
image


sdc/functions/numpy_like.py
```diff
@@ -97,3 +101,48 @@ def sdc_astype_number_impl(self, dtype):
...
+    Intel Scalable Dataframe Compiler Developer Guide
+    *************************************************
+    Parallel replacement of numpy.copy.
+
+    .. only:: developer
+       Test: python -m sdc.runtests sdc.tests.test_sdc_numpy -k copy
+
+    """
+
+    dtype = self.dtype
```
PokhodenkoSA: It could be after if.

sdc/tests/test_sdc_numpy.py
```diff
@@ -157,20 +172,57 @@ def sdc_impl(a):
...
+            return np.copy(a)
+
+        def sdc_impl():
+            a = np.array([True, False, True])
+            return numpy_like.copy(a)
+
+        sdc_func = self.jit(sdc_impl)
+        np.testing.assert_array_equal(sdc_func(), ref_impl())
+
+    @unittest.skip('Numba have not string array')
```
PokhodenkoSA:
```suggestion
    @unittest.skip('Numba doesn't have string array')
```

PokhodenkoSA: Why we have this tests?

1e-to: This is so that we add functionality if it becomes possible to create a string array in the jit region

sdc/tests/test_sdc_numpy.py
```diff
@@ -188,6 +188,57 @@ def sdc_impl(a):
             with self.subTest(data=case):
                 np.testing.assert_array_equal(sdc_func(a), ref_impl(a))
 
+    def test_copy(self):
+        def ref_impl(a):
+            return np.copy(a)
+
+        def sdc_impl(a):
+            return numpy_like.copy(a)
```
densmirn: If you want you could do something like that [sdc/tests/test_dataframe.py#L1240-L1247](https://github.com/IntelPython/sdc/blob/master/sdc/tests/test_dataframe.py#L1240-L1247).

PR #582, opened 2020-02-09
Fix for DataFrame str column created as List of Optional values

kozlov-alexey:


sdc/str_arr_ext.py
```diff
@@ -671,12 +671,16 @@ def construct_string_array(context, builder):
...
-    if isinstance(args[0], types.UniTuple):
-        assert args[0].dtype == string_type
 
-    if isinstance(args[0], types.Tuple):
-        for i in args[0]:
-            assert i.dtype == string_type or i.dtype == types.StringLiteral
+    arg = args[0]
+    if isinstance(arg, (types.UniTuple, types.List)):
+        assert (arg.dtype == string_type
+                or (isinstance(arg, types.optional) and arg.type == string_type))
```
densmirn: I'm not sure, but shouldn't we check `arg.dtype` on `optional`?

sdc/str_arr_ext.py
```diff
@@ -688,12 +692,22 @@ def str_arr_from_list(in_list):
...
 
         A = pre_alloc_string_array(n_strs, total_chars)
-        for i in range(n_strs):
-            A[i] = in_list[i]
+
+        for i in numba.prange(n_strs):
+            if nan_mask[i]:
+                A[i] = ''
+            else:
+                A[i] = in_list[i]
```
densmirn: Seems It looks better in a single line: `A[i] = '' if nan_mask[i] else in_list[i]`

sdc/tests/test_dataframe.py
```diff
@@ -1663,6 +1664,26 @@ def test_impl():
...
+                        'B': ['a', 'b', 'd', 'a', '', 'c', 'b'],
+                        'C': [2, 1, np.nan, 1, np.nan, 2, 1]
+            })
+
+            total_elements = 0
+            column_ids = (0, 1, 2)
+            for col_id in literal_unroll(column_ids):
+                total_elements += len(get_dataframe_data(df, col_id))
+            return total_elements
+        hpat_func = self.jit(test_impl)
```
densmirn: You could decorate tested function directly:
```
@self.jit
def hpat_func():
    pass
```

sdc/str_arr_ext.py
```diff
@@ -671,12 +671,16 @@ def construct_string_array(context, builder):
...
-            assert i.dtype == string_type or i.dtype == types.StringLiteral
+    arg = args[0]
+    if isinstance(arg, (types.UniTuple, types.List)):
+        assert (arg.dtype == string_type
+                or (isinstance(arg, types.optional) and arg.type == string_type))
+
+    # TODO: check whether below works at all, otherwise it might be removed
+#     if isinstance(arg, types.Tuple):
+#         for i in arg:
+#             assert i.dtype == string_type or i.dtype == types.StringLiteral
```
AlexanderKalistratov: This should work. But shouldn't we also check for None type? Or optional... I'm not sure

sdc/str_arr_ext.py
kozlov-alexey: @densmirn Yes, thank you! I''l fix that.

sdc/str_arr_ext.py
kozlov-alexey: Right, thanks!

sdc/tests/test_dataframe.py
kozlov-alexey: Fixed

sdc/str_arr_ext.py
kozlov-alexey: @AlexanderKalistratov No, actually I double-checked and it doesn't:
```
Invalid use of Function(<built-in function getitem>) with argument(s) of type(s): (Tupl
e(unicode_type, unicode_type, none, unicode_type, none, unicode_type, unicode_type), in
t64)
```
as tuple aggregates different types in this case, and applying getitem in a loop fail at:
```
    def str_arr_from_sequence(in_list):
        <source elided>
        for i in numba.prange(n_strs):
            s = in_list[i]
```
So this func only works for a UniTuple of unicode strings case now. I'll add a note about this.

sdc/str_arr_ext.py
```diff
@@ -670,33 +670,43 @@ def construct_string_array(context, builder):
...
 def impl_string_array_single(context, builder, sig, args):
-    if isinstance(args[0], types.UniTuple):
-        assert args[0].dtype == string_type
 
-    if isinstance(args[0], types.Tuple):
-        for i in args[0]:
+    arg = args[0]
+    if isinstance(arg, (types.UniTuple, types.List)):
+        assert (arg.dtype == string_type
+                or (isinstance(arg.dtype, types.optional) and arg.type == string_type))
```
densmirn: Shouldn't `arg.type == string_type` be replaced with `arg.dtype.type == string_type` and `types.optional` with `types.Optional` as above?

AlexanderKalistratov:
@densmirn are you ok with this PR?

PR #585, opened 2020-02-10
Series.sum optimization

1e-to:
image


AlexanderKalistratov:
@1e-to tests for bool series fails. We need to implement sum/nansum (and all other functions) for bools too.

densmirn:
> Wouldn't you like to move implementations of `sum` and `nansum` from overloads and call it there?

The implementations looks the same.

AlexanderKalistratov:
> Wouldn't you like to move implementations of `sum` and `nansum` from overloads and call it there?

didn't get what do you want to do

densmirn:
> > Wouldn't you like to move implementations of `sum` and `nansum` from overloads and call it there?
> 
> didn't get what do you want to do

I meant something like this:
```python
def gen_sum_number_impl():
    """Generate sum number implementation."""
    def _sum_number_impl(self):
        length = len(self)
        result = 0
        for i in prange(length):
            if self[i]:
                result += self[i]

        return result

    return _sum_number_impl


@sdc_overload(sum)
def sdc_sum_overload(self):
    # ...
    if isinstance(dtype, (types.Boolean, bool)):
        return gen_sum_number_impl()
    # ...


@sdc_overload(nansum)
def sdc_nansum_overload(self):
    # ...
    if isinstance(dtype, (types.Boolean, bool)):
        return gen_sum_number_impl()
    # ...
```

sdc/functions/numpy_like.py
```diff
@@ -201,6 +201,20 @@ def sdc_isnan_float_impl(self):
     ty_checker.raise_exc(dtype, 'int or float', 'self.dtype')
 
 
+def gen_sum_bool_impl():
+    """Generate sum bool implementation."""
+    def _sum_bool_impl(self):
+        length = len(self)
+        result = 0
+        for i in prange(length):
+            if self[i]:
```
AlexanderKalistratov: Why do you need this `if self[i]` condition?

1e-to: Without it also works, fixed

PR #587, opened 2020-02-11
Overload df.getitem with bool array idx

densmirn:


AlexanderKalistratov:
@kozlov-alexey could you please review it?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1285,7 +1333,12 @@ def _df_getitem_unicode_idx_impl(self, idx):
 
         return gen_df_getitem_bool_series_idx_impl(self, idx)
 
-    ty_checker.raise_exc(idx, 'str', 'idx')
+    if isinstance(idx, types.Array) and isinstance(idx.dtype, types.Boolean):
```
kozlov-alexey: I would propose to move it at the top, just after we checked that self is of DataFrameType, because this check should really intercept wrong types (for which we do not want compilation to happen). Otherwise we might compile (or at least try to compile) some specialization we do not want to compile (e.g. if some of the arg types can be type casted to another one).

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1059,53 +1059,32 @@ def sdc_pandas_dataframe_drop_impl(df, _func_name, args, columns):
...
+    """Generate code lines to get length of DF"""
+    if self.columns:
+        return ['  length = len(get_dataframe_data(self, 0))']
 
-    return func_lines
+    return ['  length = 0']
 
 
-def df_index_codelines(self):
+def df_index_codelines(self, with_length=False):
```
kozlov-alexey: I think it's better for this function to do only index lines generation, so that we just make df_length_codelines call not inside it, but where this func is called with with_lenght=True. This simplifies things, as on the call site you see exactly, that you generate line for lenght, then line for index. 
Alternatively we could use a format string and pass the actual var name that will hold df's len to it as a second argument, i.e.
```
def df_index_codelines(self, length):
...
        func_lines += [f'  res_index = numpy.arange({length})']
```

PR #589, opened 2020-02-11
Scale Series.fillna

1e-to:
image


AlexanderKalistratov:
Looks good, but lets move implementation into `numpy_like`. I know that numpy doesn't have such function, but it could be useful to keep implementation for arrays separately

PR #595, opened 2020-02-12
Create test case without exec

PokhodenkoSA:
This PR shows that it is not necessary to create text function and call exec. It is possible to define function inside function with values captured from context.

### DIFF
--- a/sdc/tests/tests_perf/generator.py
+++ b/sdc/tests/tests_perf/generator.py
@@ -76,27 +76,18 @@ def gen_call_expr(test_case, prefix):
 
 
 def gen_test(test_case, prefix):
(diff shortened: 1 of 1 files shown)


AlexanderKalistratov:
So?

PokhodenkoSA:
> So?

Once @Rubtsowa said that she was tired of generating functions through text. So this is the solution.

AlexanderKalistratov:
@PokhodenkoSA can we merge this PR? Or this is just an illustration?

PokhodenkoSA:
> @PokhodenkoSA can we merge this PR? Or this is just an illustration?

I think we should merge it. It is not only illustration.

PR #596, opened 2020-02-13
Add Boolean handling to Series.value_counts

akharche:


sdc/tests/test_series.py
```diff
@@ -2590,6 +2590,23 @@ def test_impl(S):
                     result = hpat_func(S).sort_index()
                     pd.testing.assert_series_equal(result, result_ref)
 
+    @skip_sdc_jit
```
densmirn: Could you add a reason why is the test skipped: `@skip_sdc_jit('The reason')`

sdc/tests/test_series.py
```diff
@@ -2590,6 +2590,23 @@ def test_impl(S):
...
+    def test_series_value_counts_boolean(self):
+        def test_impl(S):
+            return S.value_counts(sort=True)
+
+        input_data = [[True, False, True, True, False]]
+
+        hpat_func = self.jit(test_impl)
+
+        for data_to_test in input_data:
```
densmirn: Do you exactly need the loop with a single iteration?

PR #598, opened 2020-02-13
Add test data without nan

Rubtsowa:


sdc/tests/test_utils.py
```diff
@@ -59,6 +59,13 @@
 ]
 
 
+test_global_input_data_float64_without_nan_inf = [
```
densmirn: 1. Is it used anywhere?
2. I would name it `test_global_input_data_float64_finite`
3. The first internal list includes other ones. So I would leave only the first one.

Rubtsowa:
This changes are equivalent test_global_input_data_float64[0].

PR #600, opened 2020-02-13
Take Series scalable

1e-to:
WIP version of the series take, terrible compilation time, need to explore other options
image

### DIFF
--- a/sdc/functions/numpy_like.py
+++ b/sdc/functions/numpy_like.py
@@ -68,6 +68,10 @@ def nansum(self):
     pass
 
 
+def take(self, indices):
+    pass
+
+
 @sdc_overload(astype)
 def sdc_astype_overload(self, dtype):
     """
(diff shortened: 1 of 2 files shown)


sdc/functions/numpy_like.py
```diff
@@ -238,6 +242,40 @@ def sdc_isnan_float_impl(self):
...
+        return None
+
+    dtype = self.dtype
+    isnan = get_isnan(dtype)
+    if isinstance(dtype, (types.Number, types.Boolean)):
+        def sdc_take_impl(self, indices):
+            length = len(self)
+            length_res = len(indices)
+            res = numpy.empty(length_res, dtype=dtype)
+            for i in range(length_res):
```
AlexanderKalistratov: prange?

AlexanderKalistratov: why not list comprehension?

PR #601, opened 2020-02-13
Implement initial version of rewrite for df.getitem as attr

densmirn:


sdc/rewrites/dataframe_getitem_attribute.py
```diff
@@ -0,0 +1,87 @@
...
+                if not isinstance(obj, DataFrameType):
+                    continue
+                if expr.attr in obj.columns:
+                    getattrs.add(expr)
+
+            return len(getattrs) > 0
+
+        def apply(self):
+            new_block = self.block.copy()
+            new_block.clear()
```
AlexanderKalistratov: That's a bad idea. Block may contain a lot of functionality, including jumps and so on. You remove it all and add single statement.

densmirn: I add all statements from `self.block` to `new_block` additionally rewriting some instructions:
```python
for inst in self.block.body:
    # ...
    new_block.append(inst)
```

@AlexanderKalistratov could you confirm it should be fixed somehow?

AlexanderKalistratov: @densmirn you call `new_block.clear()` which erases all statements from original block. Isn't it?

AlexanderKalistratov: Ok, I get it.

sdc/rewrites/dataframe_getitem_attribute.py
```diff
@@ -0,0 +1,87 @@
...
+            for inst in self.block.body:
```
AlexanderKalistratov: I think, you could implement it like this:
```python
for expr in self.getattrs:
    const_assign = self._assign_const(inst)
    insert_before(self.block, const_assign, expr)

    new_expr = Expr.getitem(expr.value, const_assign.target, inst.loc)
    self.calltypes[new_expr] = signature(
        self.typemap[inst.target.name],
        self.typemap[expr.value.name],
        self.typemap[new_expr.index.name]
    )

    expr.value = new_expr

return self.block

```

`insert_before` you could find in ir_utils

sdc/rewrites/dataframe_getitem_attribute.py
```diff
@@ -0,0 +1,87 @@
...
+                        )
+                        inst = Assign(value=new_expr, target=inst.target, loc=inst.loc)
+                new_block.append(inst)
+
+            return new_block
+
+        def _assign_const(self, inst, prefix='$const0'):
+            """Create constant from attribute of specified instruction."""
+            const_node = Const(inst.value.attr, inst.loc)
+            const_var = Var(inst.target.scope, mk_unique_var(prefix), inst.loc)
```
AlexanderKalistratov: You could use `declare_constant` from ir_utils:

```python
declare_constant(inst.value.attr, self.block, self.func_ir, inst.loc)
```

This function also adds this constant to the beginning of the block. So, you don't need to do `new_block.append(const_assign)`

AlexanderKalistratov: When you are declaring variable (including variables assigned to constants) you must update `func_ir._definitions` dictionary. Otherwise some functionality wouldn't work and this may result in errors lately. (btw `declare_constant` updates this dictionary)

AlexanderKalistratov:
```python
func_ir._definitions[const_var.name] = [const_node]
```

sdc/rewrites/dataframe_getitem_attribute.py
densmirn: Do you want to work with original `block`, not copy of it?

sdc/rewrites/dataframe_getitem_attribute.py
```diff
@@ -0,0 +1,97 @@
...
+            self.block = block
+            self.typemap = typemap
+            self.calltypes = calltypes
+            self.getattrs = getattrs = set()
+            for expr in block.find_exprs(op='getattr'):
+                obj = typemap[expr.value.name]
+                if not isinstance(obj, DataFrameType):
+                    continue
+                if expr.attr in obj.columns:
+                    getattrs.add(expr)
```
kozlov-alexey: So we won't rewrite if we didn't find valid column name, right?
This means we won't raise AttributeError as pandas does if we use an invalid attribute name. We should probably state that in limitations, or think if we can look for all_supported_dataframe_methods in this rewrite and if we actually call a method skip rewriting, which aligns to pandas behavior:
```
>>>df
   A  B  values
0  2 -8      -8
1  1  2       2
2  1  3       3
3  1  1       1
4  2  5       5
5  2  6       6
6  1  7       7
>>>df.values
array([[ 2, -8, -8],
       [ 1,  2,  2],
       [ 1,  3,  3],
       [ 1,  1,  1],
       [ 2,  5,  5],
       [ 2,  6,  6],
       [ 1,  7,  7]], dtype=int64)
```

AlexanderKalistratov: Yes, we need to think about it, but in next PR

AlexanderKalistratov: Right now, the main issue us the case when column is named like '_data', '_columns' or '_index'. This will break all uses of dataframe

densmirn:
> @densmirn Cool feature! 🥇 Have you felt yourself a magician during writing this?

At the least I liked working on the :smile:

PR #602, opened 2020-02-14
Add data gen for numpy

Rubtsowa:


sdc/tests/tests_perf/test_perf_numpy.py
```diff
@@ -68,21 +69,16 @@ def _test_case(self, cases, name, total_data_length, data_num=1, input_data=test
         if input_data is None:
             input_data = test_global_input_data_float64
 
-        full_input_data_length = sum(len(i) for i in input_data)
+        if input_data is None:
+            input_data = test_global_input_data_float64
```
densmirn: There is code duplication.

sdc/tests/tests_perf/test_perf_numpy.py
```diff
@@ -68,21 +69,16 @@ def _test_case(self, cases, name, total_data_length, data_num=1, input_data=test
...
-                                           data_length)
-            test_data = np.array(data)
-
-            args = [test_data]
-            for i in range(data_num - 1):
-                np.random.seed(i)
-                extra_data = np.random.ranf(data_length)
-                args.append(np.array(extra_data))
+
+            args = gen_numpy_arr(data_num, data_length, input_data)
```
densmirn: 1. Is it ok that new `args` is not equal to the old one?
2. According to the name `gen_numpy_arr` the function should generate `np.array`, but it generates list of `np.array`'s.

Rubtsowa: 1. Why? lists of arrays are generated, as it was.
2. used to be lists of arrays:

   test_data = np.array(data)

            args = [test_data]
            for i in range(data_num - 1):
                np.random.seed(i)
                extra_data = np.random.ranf(data_length)
                args.append(np.array(extra_data))

AlexanderKalistratov:
@densmirn @Rubtsowa @1e-to what is the status of this?

densmirn:
> @densmirn @Rubtsowa @1e-to what is the status of this?

@Rubtsowa why this feature is needed?

PR #606, opened 2020-02-14
Scale Series.cumsum()

PokhodenkoSA:
image

### DIFF
--- a/sdc/functions/numpy_like.py
+++ b/sdc/functions/numpy_like.py
@@ -31,8 +31,6 @@
 
 """
 
-from collections import namedtuple
-
 import numba
 import numpy
 import sys
--- a/sdc/functions/numpy_like.py
+++ b/sdc/functions/numpy_like.py
@@ -865,61 +865,6 @@ def nancumsum(a):
     pass
 
 
-def get_pool_size():
-    return numba.config.NUMBA_NUM_THREADS
-
-
-@sdc_overload(get_pool_size)
-def get_pool_size_overload():
-    pool_size = get_pool_size()
-
-    def get_pool_size_impl():
-        return pool_size
-
-    return get_pool_size_impl
-
-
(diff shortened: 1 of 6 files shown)


sdc/functions/numpy_like.py
```diff
@@ -472,3 +475,187 @@ def nanprod_impl(a):
...
+        def nancumsum_impl(arr):
+            chunks = get_chunks(len(arr))
+            partial_sum = numpy.zeros(len(chunks), dtype=retty)
+            result = numpy.empty_like(arr)
+
+            for i in prange(len(chunks)):
+                chunk = chunks[i]
+                partial = zero
+                for j in range(chunk.start, chunk.stop):
+                    if ~is_nan(arr[j]):
```
AlexanderKalistratov: in case if `arr[j]` is `nan`, we should put `nan` to `result[j]`.
So, we can rewrite it in this way:
```python
result[j] = partial + arr[j]
if ~isnan(result[j]):
    partial = result[j]
```


And for consistency we could rewrite `cumsum` as:
```python
result[j] = partial + arr[j]
partial = result[j]
```

AlexanderKalistratov:
There is a case which we can't correctly support in case of parallel cumsum.
The case is basically looks something like that:
`[-max_float, max_float, max_float, -max_float]`

As soon as we'll try to sum two `max_float` we'll get `inf`. But in case we are doing it in correct order we'll get `0`. I'm suggesting to ignore such case and assume it as limitation

sdc/functions/numpy_like.py
PokhodenkoSA: With skipna=True, Pandas inserts NaN in place of NaN but NumPy treats NaN as 0 and inserts partial values in place of NaN.

PokhodenkoSA:
image

PR #607, opened 2020-02-16
Unskip 9 tests on df.getitem

densmirn:


sdc/tests/test_dataframe.py
```diff
@@ -332,7 +325,7 @@ def test_impl(n):
         self.assertEqual(count_array_REPs(), 0)
         self.assertEqual(count_parfor_REPs(), 0)
 
-    @skip_numba_jit
+    @skip_numba_jit('iadd unsupported')
```
AlexanderKalistratov: what is `iadd`?

densmirn: Actually `np.sum` of `Series` unsupported as I see. `iadd` was mentioned in internal error.

AlexanderKalistratov: do we support `loc` and `iloc` for dataframes?

densmirn: Also no, but the first error of the test run is about `np.sum`. If to exclude call of `np.sum` we get error on `loc`.

PR #611, opened 2020-02-17
Optimize dseries.rolling.mean()

densmirn:
Previous implementation results:

-- | -- | -- | -- | --
Series.rolling.mean | 1 | Python | 10000000 | 0.6860003471374512
Series.rolling.mean | 1 | SDC | 10000000 | 5.223999977111816

Optimized implementation results:

-- | -- | -- | -- | --
Series.rolling.mean | 1 | Python | 10000000 | 0.6230072975158691
Series.rolling.mean | 1 | SDC | 10000000 | 0.10899996757507324

The optimized implementation executes faster up to ~45 times than previous one and faster up to ~5 times than Python. There is no scalability due to prange isn't used at all because variable nfinite (number of finite values) is common for all threads.


sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -307,6 +305,32 @@ def apply_minp(arr, ddof, minp):
...
+
+@sdc_register_jitable
+def put_mean(value, nfinite, result):
+    """Calculate the window mean with new value."""
+    if numpy.isfinite(value):
+        nfinite += 1
+        if numpy.isnan(result):
+            result = value / nfinite
+        else:
+            result = ((nfinite - 1) * result + value) / nfinite
```
AlexanderKalistratov: Probably we can introduce new function something like `get_result`. Which for `mean` would be:
```python
def get_mean(nfinite, result):
    if nfinite:
        return result/nfinite
    return numpy.nan
```

And then, call it inside `result_or_nan`:
```python
@sdc_register_jitable
def result_or_nan(get, nfinite, minp, result):
    """Get result taking into account min periods."""
    if nfinite < minp:
        return numpy.nan

    return get(nfinite, result)
```

For `mean` we can do recalculations on `put`/`pop` stage, but for more complex functions (std, var) it is not that easy.

Hope, it wouldn't ruin our performance

PR #615, opened 2020-02-17
Fix prange utils

1e-to:


sdc/utilities/prange_utils.py
```diff
@@ -75,15 +78,17 @@ def get_chunks_impl(size, pool_size=0):
...
-            chunks.append(chunk)
+            if i == pool_size - 1:
+                rest = size - size // pool_size
+                stop = min((i + 1) * chunk_size + rest, size)
+            chunks.append(Chunk(start, stop))
 
         return chunks
 
-    return get_chunks_impl
+    return get_chunks_impl
```
PokhodenkoSA:
```suggestion
    return get_chunks_impl

```

densmirn:
> Let's hang `@sdc_register_jitable` on `get_chunks` and call the function inside of the overload implementation to avoid code duplication.

```python
@sdc_overload(get_chunks)
def get_chunks_overload(size, pool_size=0):
    def get_chunks_impl(size, pool_size=0):
        return get_chunks(size, pool_size=pool_size)

    return get_chunks_impl
```

sdc/utilities/prange_utils.py
```diff
@@ -58,12 +58,15 @@ def get_chunks(size, pool_size=0):
...
-    chunk_size = (size - 1) // pool_size + 1
+    chunk_size = size // pool_size
 
     chunks = []
     for i in range(pool_size):
         start = min(i * chunk_size, size)
         stop = min((i + 1) * chunk_size, size)
+        if i == pool_size - 1:
+            rest = size - size // pool_size
+            stop = min((i + 1) * chunk_size + rest, size)
```
densmirn: Maybe to limit the loop `for i in range(pool_size - 1)` and handle the latest step outside of the loop.

sdc/utilities/prange_utils.py
```diff
@@ -54,36 +54,76 @@ def get_pool_size_impl():
...
+
+
+@sdc_overload(get_pool_size)
+def get_pool_size_overload():
+    pool_size = get_pool_size()
+
+    def get_pool_size_impl():
+        return pool_size
+
+    return get_pool_size_impl
```
densmirn: Duplicated code.

sdc/utilities/prange_utils.py
```diff
@@ -30,6 +30,7 @@
 
 from typing import NamedTuple
 from sdc.utilities.utils import sdc_overload
+from sdc.utilities.utils import sdc_register_jitable
```
densmirn: These lines could be combined to a single one.

sdc/utilities/prange_utils.py
```diff
@@ -54,36 +55,19 @@ def get_pool_size_impl():
     return get_pool_size_impl
 
 
+@sdc_register_jitable
 def get_chunks(size, pool_size=0):
     if pool_size == 0:
         pool_size = get_pool_size()
 
-    chunk_size = (size - 1) // pool_size + 1
+    chunk_size = size // pool_size
```
AlexanderKalistratov: What's not valid.

What are you doing?

sdc/utilities/prange_utils.py
```diff
@@ -40,8 +41,8 @@ class Chunk(NamedTuple):
 def get_pool_size():
     if sdc.config.config_use_parallel_overloads:
         return numba.config.NUMBA_NUM_THREADS
-
-    return 1
+    else:
```
AlexanderKalistratov: Why?

sdc/utilities/prange_utils.py
AlexanderKalistratov: After closer look, it is valid, but a bad idea.

Let's assume, you have 56 threads and task size 55. In your implementation it wouldn't be paralleled at all. All 55 would be computed by the last chunk

PokhodenkoSA: It was my idea :) After some thinking I have the same conclusion. We should do the last thread to take lesser work
.

AlexanderKalistratov:
In it's current (not this one) implementation number of chunks is equal to the pool size.

In case task size is less than pool size first chunks would contain valid information, while the last ones would have start outside of task size.
e.g. if task size is 5 and number of threads 16 first five chunks would have valid start and stop ([0,1], [1, 2], [2,3], [3,4], [4,5]). But the rest will have start and stop outside of task size ([5,5])

This is not a problem if we a doing something like this:

```python
for i in prange(len(chunks):
    chunk = chunks[i]
    for j in range(chunk.start, chunk.stop):
        ...
```

In this case `range(chunk.start, chunk.stop)` is zero for extra chunks.
But if we need to do something like this:
```python
for i in prange(len(chunks):
    chunk = chunks[i]
    first_item = items[chunk.start]
    for j in range(chunk.start, chunk.stop):
        ...
```
This could be a problem, since `chunk.start` could be out of range.
So, probably it is better for the case task size is less than pool size return number of chunks equal to task size. I.e. in case of task size 5 and pool size 16 return only first 5 chunks.

What is your thoughts?

PokhodenkoSA:
> What is your thoughts?

I agree.
```python
pool_size = min(pool_size, size)
```

sdc/utilities/prange_utils.py
```diff
@@ -54,36 +54,20 @@ def get_pool_size_impl():
...
 
 
+@sdc_register_jitable
 def get_chunks(size, pool_size=0):
     if pool_size == 0:
         pool_size = get_pool_size()
 
-    chunk_size = (size - 1) // pool_size + 1
+    pool_size = min(pool_size, size)
+    chunk_size = size // pool_size
```
AlexanderKalistratov: Please return the initial formula `(size - 1)//pool_size + 1`

1e-to: This is not working properly.
With an array of 5 elements and 4 threads, chunks are divided into 0-2, 2-4, 4-6, 6-8

AlexanderKalistratov: No, it divided into chunks [0,2), [2,4), [4,5), [5,5). And that's fine.
I think you should add code to discard last chunk in such case

1e-to: In this module or in the function code that use this?

AlexanderKalistratov: In this module. Something like this:
```python
    pool_size = min(pool_size, size)
    chunk_size = (size - 1)//pool_size + 1

    chunks = []
    for i in range(pool_size):
        start = i*chunk_size
        stop = min((i+1)*chunk_size, size)
        if start >= size:
            break
        
        chunks.append(Chunk(start, stop))
```

sdc/utilities/prange_utils.py
```diff
@@ -54,36 +54,21 @@ def get_pool_size_impl():
...
-    chunk_size = (size - 1) // pool_size + 1
+    pool_size = min(pool_size, size)
+    chunk_size = (size - 1)//pool_size + 1
 
     chunks = []
     for i in range(pool_size):
-        start = min(i * chunk_size, size)
-        stop = min((i + 1) * chunk_size, size)
+        start = i*chunk_size
+        stop = min((i+1)*chunk_size, size)
```
densmirn: You could move calculating of `stop` after if-block.

sdc/utilities/prange_utils.py
```diff
@@ -54,36 +54,21 @@ def get_pool_size_impl():
...
 
 
+@sdc_register_jitable
 def get_chunks(size, pool_size=0):
     if pool_size == 0:
         pool_size = get_pool_size()
 
-    chunk_size = (size - 1) // pool_size + 1
+    pool_size = min(pool_size, size)
+    chunk_size = (size - 1)//pool_size + 1
```
densmirn: Why don't you round all the operations by white spaces?

sdc/utilities/prange_utils.py
PokhodenkoSA: I have implemented my vision in #618.

PR #618, opened 2020-02-18
Add tests for get_chunks and change formula for get_chunks

PokhodenkoSA:
See tests. They show how it should work.

### DIFF
--- a/sdc/tests/test_prange_utils.py
+++ b/sdc/tests/test_prange_utils.py
@@ -1,5 +1,5 @@
 # *****************************************************************************
-# Copyright (c) 2019, Intel Corporation All rights reserved.
+# Copyright (c) 2020, Intel Corporation All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
--- a/sdc/tests/test_prange_utils.py
+++ b/sdc/tests/test_prange_utils.py
@@ -79,4 +79,4 @@ class ChunkTest(TestCase):
     def test_get_chunks(self):
         for args, expected_chunks in self._get_chunks_date():
             with self.subTest(args=args):
-                self.assertEqual(get_chunks(*args), expected_chunks)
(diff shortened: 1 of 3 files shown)


sdc/utilities/prange_utils.py
```diff
@@ -54,36 +54,19 @@ def get_pool_size_impl():
     return get_pool_size_impl
 
 
+@sdc_register_jitable
 def get_chunks(size, pool_size=0):
     if pool_size == 0:
```
Hardcode84: Suggestion - remove this check and always pass `pool_size` explicitly.

Pure functions are good

PokhodenkoSA: Done.

sdc/tests/test_prange_utils.py
```diff
@@ -0,0 +1,82 @@
+# *****************************************************************************
+# Copyright (c) 2019, Intel Corporation All rights reserved.
```
densmirn: Please change year in the copyright.

PokhodenkoSA: Done

sdc/utilities/prange_utils.py
densmirn: So now we should explicitly call `get_pool_size()` and pass it to `get_chunks()`. Am I correct?

PokhodenkoSA: Yes

densmirn: Seems it's not so convenient.

PokhodenkoSA: I can introduce function like `parallel_chunks(N)` which will do call `get_chunks(N, get_pool_size())`. @Hardcode84, @densmirn ok?

densmirn: I like the idea.

PokhodenkoSA: Done

PR #619, opened 2020-02-18
Add test on Series.getitem failure

akharche:


AlexanderKalistratov:
What is the issue? getitem with negative slice is failing?

sdc/tests/test_series.py
```diff
@@ -1182,6 +1182,22 @@ def test_impl(A):
...
+        jit_impl = self.jit(test_impl)
+
+        starts = [0, -2]
+        ends = [-2, 0]
+        indices = [[2, 3, 5, 6, 7], ['2', '3', '5', '6', '7'], ['2', '3', '5', '6', '7']]
+        for start, end, index in zip(starts, ends, indices):
+            S = pd.Series([11, 22, 33, 44, 55], index, name='A')
+            ref_result = test_impl(S, start, end)
+            jit_result = jit_impl(S, start, end)
+            pd.testing.assert_series_equal(jit_result, ref_result)
```
kozlov-alexey: Please use self.subTest context to verify each combination in a separate test.
Also I think there's no need to have additional test just for negative slices as they have the same implementation, so it's better to make existing test for int slices, i.e. test_series_getitem_idx_int_slice more generic, e.g:

        ```
slice_args = [0, 2, -2]
        indexes_to_test = {
            'int': [2, 3, 5, 6, 7],
            'float': [2.0, 3, 5, 6, 7],
            'string': ['2', '3', '5', '6', '7']
        }
        for start, end in product(slice_args, slice_args):
            for index in indexes_to_test.values():
                with self.subTest(start=start, stop=end, series_index=index):
                    S = pd.Series([11, 22, 33, 44, 55], index, name='A')
                    result_ref = test_impl(S, start, end)
                    result = jit_impl(S, start, end)
```

File: sdc/tests/test_series.py [hunk 2]
```diff
@@ -1182,6 +1182,22 @@ def test_impl(A):
...
+    def test_series_getitem_idx_int_negative_slice(self):
+        def test_impl(S, start, end):
+            return S[start:end]
+
+        jit_impl = self.jit(test_impl)
+
+        starts = [0, -2]
+        ends = [-2, 0]
+        indices = [[2, 3, 5, 6, 7], ['2', '3', '5', '6', '7'], ['2', '3', '5', '6', '7']]
+        for start, end, index in zip(starts, ends, indices):
```
kozlov-alexey: This would iterate over zipped sequences but till the end of the shortest one, so the last string index in indices  won't be tested at all.

kozlov-alexey:
> What is the issue? getitem with negative slice is failing?
@akharche @AlexanderKalistratov The issue here is in the empty slice for the string series index only (other combinations seem to pass), i.e. original test fails on the second iteration, when start=-2, end=0 and index is ['2', '3', '5', '6', '7']. And the resulting slice should be:
```
a = np.asarray([1, 2, 3, 1, 2, 4, 5])
a[-2:0]
array([], dtype=int32)
```

sdc/tests/test_series.py
AlexanderKalistratov: @kozlov-alexey as far as I get, this is negative tests for negative slices. If it would be combined with positive slices we would need to skip all tests, instead of skiping specific ones.

sdc/tests/test_series.py
```diff
@@ -1182,6 +1182,22 @@ def test_impl(A):
         S = pd.Series(['1', '4', '6', '33', '7'], ['1', '3', '5', '4', '7'], name='A')
         pd.testing.assert_series_equal(hpat_func(S), test_impl(S))
 
+    @unittest.skip('LLVM ERROR: out of memory')
+    def test_series_getitem_idx_int_negative_slice(self):
+        def test_impl(S, start, end):
+            return S[start:end]
+
+        jit_impl = self.jit(test_impl)
```
kozlov-alexey: It would probably be better to use the same name (hpat_func) everywhere.

sdc/tests/test_series.py
kozlov-alexey: @AlexanderKalistratov What do you mean by negative tests? That an empty slice appear as a result? If so, that's true. Otherwise, I think it should not matter whether slice has negatives or positives from the test perspective (assuming it's a correct slice).

AlexanderKalistratov: @kozlov-alexey sounds reasonable. I thought this is some kind of numba limitation (negative slices). But it seems to be our own problem.

PokhodenkoSA (commented):

PR #629, opened 2020-02-20
GroupBy.getitem for single column case aligned with pandas

kozlov-alexey:


sdc/datatypes/hpat_pandas_groupby_functions.py
```diff
@@ -232,6 +305,28 @@ def sdc_pandas_groupby_apply_func(self, func_name, func_args, defaults=None, imp
     return _groupby_method_impl
 
 
+def sdc_pandas_series_groupby_apply_func(self, func_name, func_args, defaults=None, impl_args=None):
+
+    defaults = defaults if defaults else {}
+    impl_args = impl_args if impl_args else {}
```
densmirn: As an option:
```python
defaults = defaults or {}
impl_args = impl_args or {}
```

kozlov-alexey: Fixed in both places.

PR #630, opened 2020-02-21
Simplify series.rolling methods generator + median

densmirn:
Current performance:

-- | -- | -- | -- | --
Series.rolling.median | 1 | Python | 100000 | 0.263
Series.rolling.median | 1 | SDC | 100000 | 0.388
Series.rolling.median | 4 | Python | 100000 | 0.266
Series.rolling.median | 4 | SDC | 100000 | 0.262

`Python 1 / SDC 1 = 0.678`
`SDC 1 / SDC 4 = 1.481`
Scalability is pretty weak, but exists.

Implementation with chunks is much slower than the current one.
```python
chunks = parallel_chunks(length)
for i in prange(len(chunks)):
    chunk = chunks[i]
    for idx in range(chunk.start, chunk.stop):
        finite_arr = win_finite_arr(input_arr, i, win)
        if len(finite_arr) < minp:
            output_arr[i] = numpy.nan
        else:
            output_arr[i] = rolling_func(finite_arr)
```

-- | -- | -- | -- | --
Series.rolling.median | 1 | Python | 100000 | 0.339
Series.rolling.median | 1 | SDC | 100000 | 1.32
Series.rolling.median | 4 | Python | 100000 | 0.3
Series.rolling.median | 4 | SDC | 100000 | 5.2

`Python 1 / SDC 1 = 0.257`
`SDC 1 / SDC 4 = 0.254`
There is no scalability at all.

### DIFF
--- a/sdc/datatypes/hpat_pandas_series_rolling_functions.py
+++ b/sdc/datatypes/hpat_pandas_series_rolling_functions.py
@@ -225,6 +225,15 @@ def arr_var(arr, ddof):
     return numpy.var(arr) * length / (length - ddof)
 
 
+@sdc_register_jitable
+def win_finite_arr(arr, idx, win_size):
+    """Get array of finite values from the window."""
+    start = max(0, idx - win_size + 1)
+    arr_range = arr[start:idx + 1]
+
+    return arr_range[numpy.isfinite(arr_range)]
+
+
 def gen_hpat_pandas_series_rolling_impl(rolling_func):
     """Generate series rolling methods implementations based on input func"""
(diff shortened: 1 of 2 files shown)


AlexanderKalistratov:
So. Does it gives us speedup?
It's absolutely unclear for me why we can't see any scalability.
Let's put this one on hold and continue with rest and return to this if we would have enough time.

samaid:
For rolling method you might want to consider online median when you
collect k smallest values where k is window size. It should work well if k
<< n which is the total size of dataset

PR #635, opened 2020-02-25
Df.head impl

1e-to:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -454,14 +454,117 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
+
+
+def sdc_pandas_dataframe_head_codegen(df, func_name, params, ser_params):
+    all_params = ['df']
+    ser_par = []
+
+    for key, value in params.items():
+        all_params.append('{}={}'.format(key, value))
+    for key, value in ser_params.items():
+        ser_par.append('{}={}'.format(key, value))
```
densmirn: Let's use from `sdc.utilities.sdc_typing_utils import kwsparams2list` here.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -454,14 +454,117 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
 
+    s_par = '{}'.format(', '.join(ser_par[:]))
```
densmirn: Why do you use a slice here?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -454,14 +454,117 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
     loc_vars = {}
     exec(func_text, global_vars, loc_vars)
     _reduce_impl = loc_vars[df_func_name]
 
     return _reduce_impl
 
 
+def df_index_codegen(self):
+    if isinstance(self.index, types.NoneType):
+        func_lines = ''
```
densmirn: You could return a string directly without creating the variable.

sdc/tests/test_dataframe.py
```diff
@@ -945,6 +945,39 @@ def test_impl(n):
...
+                return a.head(n)
+
+            return impl
+
+        for n in [-3, 0, 3, 5, None]:
+            for idx in [[3, 4, 2, 6, 1], None]:
+                ref_impl = get_func(n)
+                sdc_impl = get_func(n)
+                sdc_func = self.jit(sdc_impl)
+                with self.subTest(n=n):
```
densmirn:
```suggestion
                with self.subTest(n=n, index=idx):
```

sdc/tests/test_dataframe.py
```diff
@@ -945,6 +945,39 @@ def test_impl(n):
...
+                                   "STRING": ['a', 'dd', 'c', '12', 'ddf']}, index=idx)
+                with self.subTest(n=n, index=idx):
+                    pd.testing.assert_frame_equal(sdc_func(df, n), test_impl(df, n))
+
+    def test_df_head(self):
+        def get_func(n):
+            def impl(a):
+                return a.head(n)
+
+            return impl
```
densmirn: It''s interesting why do you need implementation generator?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -454,14 +454,117 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
+    else:
+        func_lines = ', index = df._index[:n]'
```
densmirn: White spaces around of `=` are excess.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -454,14 +454,117 @@ def sdc_pandas_dataframe_reduce_columns(df, func_name, params, ser_params):
...
+    def _df_head_impl(df, n=5):
+        series_FLOAT = pandas.Series(get_dataframe_data(df, 0))
+        result_FLOAT = series_FLOAT.head(n=n)
+        series_INT = pandas.Series(get_dataframe_data(df, 1))
+        result_INT = series_INT.head(n=n)
+        series_STRING = pandas.Series(get_dataframe_data(df, 2))
+        result_STRING = series_STRING.head(n=n)
+        return pandas.DataFrame({"FLOAT": result_FLOAT, "INT": result_INT, "STRING": result_STRING},
+                                index = df._index[:n])
+"""
```
densmirn: Don't you want to move the docstring to a function?

sdc/datatypes/hpat_pandas_dataframe_functions.py
densmirn: Also why do you use `format` here?

sdc/tests/test_dataframe.py
1e-to: This is necessary to test n which is getting outside the jit region

PR #636, opened 2020-02-25
Df.isna impl

1e-to:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -935,6 +935,105 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
...
+    func_lines = [f'def _df_{func_name}_impl({joined}):']
+    ind = df_index_codegen_all(df)
+    for i, c in enumerate(columns):
+        result_c = f'result_{c}'
+        func_lines += [f'  series_{c} = pandas.Series(get_dataframe_data({func_params[0]}, {i}))',
+                       f'  {result_c} = series_{c}.{func_name}()']
+        results.append((columns[i], result_c))
+
+    data = ', '.join(f'"{col}": {data}' for col, data in results)
+    func_lines += [f'  return pandas.DataFrame({{{data}}}{ind})']
```
akharche: The index value is not clear. May be it is better to define index code line here?
```suggestion
    index = 'df.index'
    func_lines.append(f"  return pandas.DataFrame({{{data}}}, index={index})")
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -935,6 +935,105 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
...
 
 
+def _dataframe_reduce_columns_codegen_isna(func_name, func_params, columns, df):
+    results = []
+    joined = ', '.join(func_params)
+    func_lines = [f'def _df_{func_name}_impl({joined}):']
+    ind = df_index_codegen_all(df)
+    for i, c in enumerate(columns):
+        result_c = f'result_{c}'
+        func_lines += [f'  series_{c} = pandas.Series(get_dataframe_data({func_params[0]}, {i}))',
```
```suggestion
        func_lines += [f'  series_{c} = pandas.Series(get_dataframe_data({df}, {i}))',
```
where would be defined above as:
df = func_params[0]

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -935,6 +935,105 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
...
+    df_func_name = f'_df_{func_name}_impl'
+    func_text, global_vars = _dataframe_reduce_columns_codegen_isna(func_name, all_params, df.columns, df)
+    loc_vars = {}
+    exec(func_text, global_vars, loc_vars)
+    _reduce_impl = loc_vars[df_func_name]
+
+    return _reduce_impl
+
+
+def df_index_codegen_all(self):
```
kozlov-alexey: Why do we need this function at all? Why can't we always use "index=df._index", i.e. even if it's None it should still be handled by dataframe ctor?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -935,6 +935,105 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
...
+
+    Intel Scalable Dataframe Compiler Developer Guide
+    *************************************************
+    Pandas DataFrame method :meth:`pandas.DataFrame.isna` implementation.
+
+    .. only:: developer
+        Test: python -m sdc.runtests -k sdc.tests.test_dataframe.TestDataFrame.test_df_isna*
+    """
+
+    return sdc_pandas_dataframe_isna_codegen(df, 'isna', {}, {})
```
kozlov-alexey: It would make sense to make params, ser_params have None value by default and just omit them here.
And inside the function redefine them if they are None as follows:
```
params = params if params is not None else {}
ser_params = ser_params if ser_params is not None else {}
```

sdc/tests/test_dataframe.py
```diff
@@ -810,6 +810,19 @@ def test_impl(df):
                            'D': [None, 'dd', '', None]})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_df_isna(self):
+        def test_impl(df):
+            return df.isna()
+        sdc_func = sdc.jit(test_impl)
+        for idx in [[3, 4, 2, 6, 1], ['a', 'b', 'c', 'd', 'e'], None]:
+            df = pd.DataFrame({"FLOAT": [3.2, np.nan, 7.0, 3.3, np.nan],
```
kozlov-alexey: Please avoid using uppercase words/sentences where they are not necessary (they put additional emphasis from text formatting perspective). If you want to highlight that each column has it's own dtype in the name of a column just use lowercase names.

sdc/tests/test_dataframe.py
```diff
@@ -810,6 +810,19 @@ def test_impl(df):
                            'D': [None, 'dd', '', None]})
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_df_isna(self):
+        def test_impl(df):
+            return df.isna()
+        sdc_func = sdc.jit(test_impl)
+        for idx in [[3, 4, 2, 6, 1], ['a', 'b', 'c', 'd', 'e'], None]:
```
kozlov-alexey: That's bad due to inability to comment particular indexes during debugging. I would use:
```
        sdc_func = sdc.jit(test_impl)
        df_indexes = [
               [3, 4, 2, 6, 1],
               ['a', 'b', 'c', 'd', 'e'],
               None
        ]
        for idx in df_indexes:
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -935,45 +935,41 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
     return func_def, global_vars
 
 
-def _dataframe_reduce_columns_codegen_isna(func_name, func_params, columns, df):
+def _dataframe_reduce_columns_codegen_isna(func_name, columns, df):
```
```suggestion
def _dataframe_codegen_isna(func_name, columns, df):
```

sdc/tests/test_dataframe.py
```diff
@@ -813,16 +813,26 @@ def test_impl(df):
...
+
+        for idx in indexes:
+            df = pd.DataFrame({"float": [3.2, np.nan, 7.0, 3.3, np.nan],
+                               "int": [3, 4, 1, 0, 222],
+                               "string": ['a', 'dd', 'c', '12', None]}, index=idx)
             with self.subTest(index=idx):
                 pd.testing.assert_frame_equal(sdc_func(df), test_impl(df))
 
+    @unittest.skip('DF with bool data Segmentation fault')
+    def test_df_bool(self):
```
kozlov-alexey: Unboxing bool column of a DF and calling isna on it works fine, so the bools are not the reason here. It seems the problem is linked with column name "bool", so I would advise just renaming the columns to ordinary names 'A', 'B', 'C', etc and add the bool column into the test test_df_isna. This way it would be complete.
As for the failing test let's rewrite and rename it so that it's clear we have an issue with such column names (you can also dig a bit deeper into the problem and try to locate the root cause and mention it in test skip reason).

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -983,11 +979,8 @@ def sdc_pandas_dataframe_isna_codegen(df, func_name, params, ser_params):
 
 def df_index_codegen_all(self):
     if isinstance(self.index, types.NoneType):
-        func_lines = ''
-    else:
-        func_lines = ', index = df._index'
-
-    return func_lines
+        return ''
+    return 'index=df._index'
```
kozlov-alexey: This hardcodes the name of the df param, not sure if it's worth fixing this.

1e-to: This is a temporary solution until we fix the creation of a dataframe with the None index.

PR #631, opened 2020-02-21
tbb concurrent hash usage example

Hardcode84:

PR #644, opened 2020-02-26
Scale quantile via DAAL

PokhodenkoSA:
This PR is for discussion. It shows how to pass NumPy array to C++ module.
Also it shows that using ctypes functions is more convenient than `ll.add_symbol()` like in #631 b/c function could be run in Python too.
Numba/Numpy:
image

DAAL:
image


Alexander-Makaryev:
@PokhodenkoSA to control number of threads use this
https://github.com/IntelPython/daal4py/blob/master/src/daal4py.cpp#L604

sdc/_daal.cpp
```diff
@@ -0,0 +1,91 @@
...
+    return result;
+}
+
+double quantile(int c, double *p, double q)
+{
+    using namespace daal;
+    using namespace daal::algorithms;
+    using namespace daal::data_management;
+
+    quantiles::Batch<> algorithm;
```
Alexander-Makaryev: You should specify the data type of algorithm as a template parameter and provide data with the same type. Default type is used now but it is not clear what this type is. If you will provide the double data to the float algorithm then time will be spent on converting and copying.

PR #651, opened 2020-02-27
Series str zfill

Rubtsowa:


sdc/tests/test_series.py
```diff
@@ -3148,6 +3148,17 @@ def test_impl(series, width):
                 ref_result = test_impl(series, width)
                 pd.testing.assert_series_equal(jit_result, ref_result)
 
+    @unittest.expectedFailure  # https://github.com/numba/numba/issues/5317
```
densmirn: The test fails due to results of `series.str.zfill` and built-in `zfill` are different in Python. Currently let's have such test with such description. I think the issue link is incorrect here. And let's rename the test to e.g. `test_series_str_zfill_limitation` and add the limitation to the documentation (docstring).

PR #656, opened 2020-03-02
Df reset index

Rubtsowa:


sdc/hiframes/pd_dataframe_ext.py
```diff
@@ -907,47 +907,47 @@ def lower_fillna_dummy(context, builder, sig, args):
...
-            return signature(out_df, *args)
-        return signature(types.none, *args)
-
-
-@lower_builtin(reset_index_dummy, types.VarArg(types.Any))
-def lower_reset_index_dummy(context, builder, sig, args):
-    out_obj = cgutils.create_struct_proxy(
-        sig.return_type)(context, builder)
-    return out_obj._getvalue()
+# @overload_method(DataFrameType, 'reset_index')
```
AlexanderKalistratov: Either remove it or but under `if` condition

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1516,3 +1516,109 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+def sdc_pandas_dataframe_reset_index_codegen(df_type, all_params, columns, result_name, func_line):
+    joined = ', '.join(all_params)
+    func_lines = [f'def _df_reset_index_impl({joined}):']
+    df = all_params[0]
+    func_lines += func_line
+    for i, c in enumerate(columns):
+        result_c = f'result_{c}'
+        func_lines += [
+            f'  series_{c} = pandas.Series(get_dataframe_data({df}, {i}))',
```
AlexanderKalistratov: Why do you create Series if you use only data from it?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1516,3 +1516,109 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+
+def sdc_pandas_dataframe_reset_index_impl(self, level=None, drop=False, inplace=False, col_level=0, col_fill=''):
+    all_params = ['self', 'level=None', 'drop=False', 'inplace=False', 'col_level=0', 'col_fill=""']
+    df_func_name = f'_df_reset_index_impl'
+
+    result_name = []
+    func_lines = []
+    drop_val = str(drop)[14:-1]
+    if drop_val == 'False':
```
AlexanderKalistratov: Looks suspicious. Why do you compare with string?

Rubtsowa: because here drop == Literal[bool](False) or Literal[bool](True)

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+def sdc_pandas_dataframe_reset_index_codegen(df_type, all_params, columns, result_name, func_line):
+    joined = ', '.join(all_params)
+    func_lines = [f'def _df_reset_index_impl({joined}):']
+    df = all_params[0]
+    func_lines += func_line
+    for i, c in enumerate(columns):
+        result_c = f'result_{c}'
+        func_lines += [
+            f'  result_{c} = get_dataframe_data({df}, {i})'
```
akharche: Please, use column id instead it's name to avoid fails when column names are with spaces

sdc/tests/test_dataframe.py
```diff
@@ -1152,6 +1151,15 @@ def test_impl(df):
         hpat_func = self.jit(test_impl)
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_df_reset_index2(self):
+        def test_impl(df):
+            return df.reset_index(drop=False)
+
+        df = pd.DataFrame({'A': [1.0, 2.0, np.nan, 1.0]})
```
akharche: What do you test if you create DataFrame without index?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+    return _apply_impl
+
+
+@sdc_overload_method(DataFrameType, 'reset_index')
+def sdc_pandas_dataframe_reset_index(self, level=None, drop=False, inplace=False, col_level=0, col_fill=''):
+    """
+   Pandas DataFrame method :meth:`pandas.DataFrame.reset_index` implementation.
+
+   .. only:: developer
```
akharche: Is it the old style of docstrings?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
         return init_dataframe_groupby(self, column_id, grouped, sort)
 
     return sdc_pandas_dataframe_groupby_impl
+
+
+def sdc_pandas_dataframe_reset_index_codegen(df_type, all_params, columns, result_name, func_line):
+    joined = ', '.join(all_params)
```
akharche: Please, provide the docstring with resulted function

sdc/datatypes/hpat_pandas_dataframe_functions.py
AlexanderKalistratov: So, why not to use `drop.literal_value`?

examples/dataframe/dataframe_reset_index_drop_False.py
```diff
@@ -0,0 +1,40 @@
+#     this list of conditions and the following disclaimer.
```
kozlov-alexey: In both added examples copyright header is corrupted

sdc/tests/test_dataframe.py
```diff
@@ -1143,15 +1143,46 @@ def test_impl(A):
         hpat_func = self.jit(test_impl)
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df2))
 
-    @skip_numba_jit
-    def test_df_reset_index1(self):
+    def test_df_reset_index_drop_True(self):
         def test_impl(df):
             return df.reset_index(drop=True)
```
kozlov-alexey: These tests are OK, but both of them have drop=True/False hardcoded in the jitted function body, so you only test that your implementation works with literal bool param. To test that it works with variable of type Boolean you need to define a variable in test and then pass this var into the jitted function, e.g:
```
            for skipna_var in [True, False]:
                actual = hpat_func(S, skipna=skipna_var)
                expected = test_impl(S, skipna=skipna_var)
```
And such test is definitely needed.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,149 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+    func_name = 'reset_index'
+
+    ty_checker = TypeChecker('Method {}().'.format(func_name))
+    ty_checker.check(self, DataFrameType)
+
+    if not (isinstance(level, (types.Omitted, types.NoneType)) or level is None):
+        ty_checker.raise_exc(level, 'None', 'level')
+
+    if not isinstance(drop, (types.Omitted, types.Boolean)):
```
```suggestion
    if not (isinstance(drop, (types.Omitted, types.Boolean)) or drop is False):
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,149 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+        ty_checker.raise_exc(drop, 'bool', 'drop')
+
+    if not (isinstance(inplace, (types.Omitted, types.Boolean)) or inplace is False):
+        ty_checker.raise_exc(inplace, 'False', 'inplace')
+
+    if not (isinstance(col_level, (types.Omitted, types.Integer)) or col_level == 0):
+        ty_checker.raise_exc(col_level, '0', 'col_level')
+
+    if not (isinstance(col_fill, (types.Omitted, types.StringLiteral)) or col_fill == ""):
+        ty_checker.raise_exc(col_fill, '""', 'col_fill')
```
kozlov-alexey: There's no point in checking types of unsupported params, instead you should check that these parameters were not provided and raise exception if they were, e.g.:
```
    if not (level is None or isinstance(level, types.Omitted)):
        raise TypingError('{} Unsupported parameters. Given inplace: {}'.format(_func_name, level))
```

sdc/tests/test_dataframe.py
```diff
@@ -1143,15 +1143,46 @@ def test_impl(A):
...
+            'A': ['a', 'b', None, 'a', '', None, 'b'],
+            'B': ['a', 'b', 'd', 'a', '', 'c', 'b'],
+            'C': [np.nan, 1, 2, 1, np.nan, 2, 1],
+            'D': [1, 2, 9, 5, 2, 1, 0]
+        })
+        hpat_func = self.jit(test_impl)
+
+        pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
+
+    def test_df_reset_index_with_index(self):
```
kozlov-alexey:
```suggestion
    def test_df_reset_index_drop_false_index_int(self):
```

sdc/tests/test_dataframe.py
```diff
@@ -1143,15 +1143,46 @@ def test_impl(A):
...
-    def test_df_reset_index1(self):
+    def test_df_reset_index_drop_True(self):
         def test_impl(df):
             return df.reset_index(drop=True)
 
         df = pd.DataFrame({'A': [1.0, 2.0, np.nan, 1.0]})
         hpat_func = self.jit(test_impl)
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
 
+    def test_df_reset_index_drop_False(self):
```
kozlov-alexey:
```suggestion
    def test_df_reset_index_drop_false_index_default(self):
```

sdc/tests/test_dataframe.py
```diff
@@ -1143,15 +1143,46 @@ def test_impl(A):
         hpat_func = self.jit(test_impl)
         pd.testing.assert_frame_equal(hpat_func(df), test_impl(df2))
 
-    @skip_numba_jit
-    def test_df_reset_index1(self):
+    def test_df_reset_index_drop_True(self):
```
kozlov-alexey: Can we extend this test with different indexes (each tested in separate subTest) in the DF?
The indexes can be:
indexes_to_test = {
     'default': None,
     'int': np.arange(n),
     'float': np.arange(n, dtype='float'),
     'str': gen_strlist(n, 2, 'abcd123 ')
}

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,149 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
         return init_dataframe_groupby(self, column_id, grouped, sort)
 
     return sdc_pandas_dataframe_groupby_impl
+
+
+def sdc_pandas_dataframe_reset_index_codegen(df_type, all_params, columns, result_name, func_line):
+    """
+    Example of generated implementation:
+        def _df_reset_index_impl(self, level=None, drop=False, inplace=False, col_level=0, col_fill=""):
+          length = len(get_dataframe_data(self, 0))
```
kozlov-alexey: Will this work with empty DF? Probably it's better to use df_length_codelines function, if length is really needed.

sdc/tests/test_dataframe.py
```diff
@@ -1143,15 +1143,46 @@ def test_impl(A):
...
+    def test_df_reset_index_drop_False(self):
+        def test_impl(df):
+            return df.reset_index(drop=False)
+
+        df = pd.DataFrame({'A': [1.0, 2.0, np.nan, 1.0]})
+        hpat_func = self.jit(test_impl)
+
+        pd.testing.assert_frame_equal(hpat_func(df), test_impl(df))
+
+    def test_df_reset_index_df_different_data(self):
```
kozlov-alexey: What's the point of this test, to test that impl works for multiple columns as it does for single one? If so it's better to update previous tests (test_df_reset_index_drop_False, etc), so that DF in them has more than one column. Testing various data types/runtime values is redundant in my opinion, as method impl does not depend on them at all.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,149 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+    df = all_params[0]
+    func_lines += func_line
+    for i, c in enumerate(columns):
+        result_c = f'result_{i}'
+        func_lines += [
+            f'  result_{i} = get_dataframe_data({df}, {i})'
+        ]
+        result_name.append((result_c, c))
+    data = ', '.join(f'"{column_name}": {column}' for column, column_name in result_name)
+    func_lines += [f'  return pandas.DataFrame({{{data}}}, index=numpy.arange(len(self)))']
```
```suggestion
    func_lines += [f'  return pandas.DataFrame({{{data}}})']
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1627,149 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+def sdc_pandas_dataframe_reset_index_impl(self, level=None, drop=False, inplace=False, col_level=0, col_fill=''):
+    all_params = ['self', 'level=None', 'drop=False', 'inplace=False', 'col_level=0', 'col_fill=""']
+    df_func_name = f'_df_reset_index_impl'
+
+    result_name = []
+    func_lines = []
+    drop_val = str(drop)[14:-1]
+    if drop_val == 'False':
+        result_name, func_lines = sdc_pandas_dataframe_reset_index_drop_False(self, result_name, func_lines)
```
kozlov-alexey: All this is not needed at all - there should be one function that performs codegen and it's generally a bad idea to produce func text with many functions each passing new own lines to subsequent funcs, like here in func_lines argument. Instead the codegen could have drop param and depending on it's value just add or not add some lines in the resulting func code, e.g. in this case you just need "one function to rule them all":

`    func_text, global_vars = sdc_pandas_dataframe_reset_index_codegen(self, all_params, self.columns,                                                                      drop_literal)`
where:
```
def sdc_pandas_dataframe_reset_index_codegen(df_type, all_params, columns, drop):
    """
    Example of generated implementation:
        def _df_reset_index_impl(self, level=None, drop=False, inplace=False, col_level=0, col_fill=""):
          length = len(get_dataframe_data(self, 0))
          res_index = numpy.arange(length)
          result_0 = get_dataframe_data(self, 0)
          return pandas.DataFrame({"index": res_index, "A": result_0}, index=numpy.arange(len(self)))
   """
    joined = ', '.join(all_params)
    df = all_params[0]
    func_lines = [f'def _df_reset_index_impl({joined}):']
    result_columns = []
    if not drop:
        func_lines += [
            f'  old_index = {df}.index'
        ]
        result_columns.append(('old_index', 'index'))
    for i, c in enumerate(columns):
        result_c = f'result_{i}'
        func_lines += [
            f'  result_{i} = get_dataframe_data({df}, {i})'
        ]
        result_columns.append((result_c, c))
    data = ', '.join(f'"{column_name}": {column}' for column, column_name in result_columns)
    func_lines += [f'  return pandas.DataFrame({{{data}}})']

    global_vars = {'pandas': pandas,
                   'numpy': numpy,
                   'get_dataframe_data': get_dataframe_data}

    return func_text, global_vars
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1735,3 +1736,142 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+          result_0 = get_dataframe_data(self, 0)
+          return pandas.DataFrame({"index": res_index, "A": result_0}, index=numpy.arange(len(self)))
+    """
+    result_name = []
+    joined = ', '.join(all_params)
+    func_lines = [f'def _df_reset_index_impl({joined}):']
+    df = all_params[0]
+    if not drop.literal_value:
+        old_index = f'old_index'
+        func_lines += [f'  old_index = {df}.index']
```
```suggestion
        old_index = 'old_index'
        func_lines += [f'  {old_index} = {df}.index']
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1792,3 +1793,138 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+@sdc_overload_method(DataFrameType, 'reset_index')
+def sdc_pandas_dataframe_reset_index(self, level=None, drop=False, inplace=False, col_level=0, col_fill=''):
+    """
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+    Pandas API: pandas.DataFrame.reset_index
+
+    Limitations
+    -----------
+    Reset the index of the DataFrame, and use the default one instead.
```
kozlov-alexey: Why it doesn't say anything about supporting drop as literal values only?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1792,3 +1793,138 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+        raise TypingError('{} Unsupported parameter inplace. Given: {}'.format(func_name, inplace))
+
+    if not (col_level == 0 or isinstance(col_level, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_level. Given: {}'.format(func_name, col_level))
+
+    if not (col_fill == '' or isinstance(col_fill, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_fill. Given: {}'.format(func_name, col_fill))
+
+    if not isinstance(drop, types.Literal):
+        raise Exception('{} Supported only literally drop.'.format(func_name))
```
kozlov-alexey: Why not raise SDCLimitation? or even better why not provide second impl that will call literally?

sdc/datatypes/hpat_pandas_dataframe_functions.py
kozlov-alexey: Eventually, will new implementation work with empty DF?

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1792,3 +1793,138 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+    if not drop.literal_value:
+        old_index = 'old_index'
+        func_lines += [f'  {old_index} = {df}.index']
+        result_name.append((old_index, 'index'))
+    for i, c in enumerate(columns):
+        result_c = 'result_c'
+        func_lines += [
+            f'  {result_c} = get_dataframe_data({df}, {i})'
+        ]
+        result_name.append((result_c, c))
```
kozlov-alexey: So you will add the same (last) column into the DF many times, check with:
df = pd.DataFrame({'A': [1.0, 2.0, np.nan, 1.0], 'B': np.arange(4.0)}, index=[5, 8, 4, 6])

sdc/tests/test_dataframe.py
```diff
@@ -1236,13 +1237,27 @@ def test_impl(A):
...
+                with self.assertRaises(Exception) as raises:
+                    hpat_func(df, drop)
+                msg = 'reset_index Supported only literally drop.'
+                self.assertIn(msg.format(types.bool_), str(raises.exception))
+
+    def test_df_reset_index_drop_false_index_int(self):
+        def test_impl(df):
+            return df.reset_index(drop=False)
+
+        df = pd.DataFrame({'A': [1.0, 2.0, np.nan, 1.0]}, index=[5, 8, 4, 6])
```
kozlov-alexey: Testing single columns is obviously not enough to verify implementation is correct.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1792,3 +1793,138 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+    if not (col_level == 0 or isinstance(col_level, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_level. Given: {}'.format(func_name, col_level))
+
+    if not (col_fill == '' or isinstance(col_fill, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_fill. Given: {}'.format(func_name, col_fill))
+
+    if not isinstance(drop, types.Literal):
+        raise Exception('{} Supported only literally drop.'.format(func_name))
+
+    return sdc_pandas_dataframe_reset_index_impl(self, level=level, drop=drop, inplace=inplace,
```
kozlov-alexey: Why do you need to pass level, inplace and other unsupported params into this function? The aren't used there at all.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1792,3 +1793,138 @@ def _df_set_column_unicode_key_impl(self, key, value):
...
+    if not (inplace is False or isinstance(inplace, types.Omitted)):
+        raise TypingError('{} Unsupported parameter inplace. Given: {}'.format(func_name, inplace))
+
+    if not (col_level == 0 or isinstance(col_level, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_level. Given: {}'.format(func_name, col_level))
+
+    if not (col_fill == '' or isinstance(col_fill, types.Omitted)):
+        raise TypingError('{} Unsupported parameter col_fill. Given: {}'.format(func_name, col_fill))
+
+    if not isinstance(drop, types.Literal):
```
kozlov-alexey: Below implementation will only work with Boolean literals, so the check must be aligned.

sdc/datatypes/hpat_pandas_dataframe_functions.py
kozlov-alexey: @Rubtsowa I checked with:
df = pd.DataFrame({})
df = pd.DataFrame({}, index=[5, 8, 4, 6])
new implementation is called, but results don't match to reference results. Moreover 2nd case segfaults. For reset_index we need two more tests:
1. test_df_reset_index_drop_true_index_int, that will test drop=True
2. test_df_reset_index_empty_df (will be skipped)

PR #659, opened 2020-03-03
Df len

Rubtsowa:
Method worked is not correct. Always returned result lambda was 0.


sdc/tests/test_dataframe.py
```diff
@@ -1879,6 +1879,15 @@ def test_impl():
...
         self.assertTrue(isinstance(three, np.ndarray))
 
+    def test_df_len(self):
+        def test_impl(df):
+            return len(df)
+
+        hpat_func = self.jit(test_impl)
+        n = 11
+        df = pd.DataFrame({'A': np.arange(n), 'B': np.arange(n) ** 2})
+        assert (hpat_func(df), test_impl(df))
```
AlexanderKalistratov: This assert is always true.
Why do you use plain assert at all?

Rubtsowa: because result len() - number.

AlexanderKalistratov: https://docs.python.org/3/library/unittest.html#assert-methods

AlexanderKalistratov:
too many files and changes

PR #663, opened 2020-03-03
Removing lazy unboxing for DataFrames

kozlov-alexey:


AlexanderKalistratov:
Fixed in https://github.com/IntelPython/sdc/pull/667

PR #667, opened 2020-03-04
Optimize get_dataframe_data function

Hardcode84:
* Remove lazy unboxing
* Simplify and force inline `get_dataframe_data` function to remove excessive increfs/decrefs

Based on #663

### DIFF
--- a/sdc/hiframes/boxing.py
+++ b/sdc/hiframes/boxing.py
@@ -113,8 +113,6 @@ def unbox_dataframe(typ, val, c):
 
         dataframe.data = c.builder.insert_value(
             dataframe.data, native_val.value, col_ind)
-        dataframe.unboxed = c.builder.insert_value(
-            dataframe.unboxed, c.context.get_constant(types.int8, 1), col_ind)
 
     # TODO: support unboxing index
     if typ.index == types.none:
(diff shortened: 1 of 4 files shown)


sdc/hiframes/boxing.py
```diff
@@ -100,9 +100,19 @@ def unbox_dataframe(typ, val, c):
...
+    for col_ind in range(n_cols):
+        series_obj = c.pyapi.object_getattr_string(val, typ.columns[col_ind])
+        arr_obj = c.pyapi.object_getattr_string(series_obj, "values")
+        ty_series = typ.data[col_ind]
+        if isinstance(ty_series, types.Array):
+            native_val = unbox_array(typ.data[col_ind], arr_obj, c)
+        elif ty_series == string_array_type:
+            native_val = unbox_str_series(string_array_type, series_obj, c)
+
+        dataframe.data = c.builder.insert_value(
```
AlexanderKalistratov: I believe we can set data once, not in every iteration

Hardcode84: This code works like this:
```
data1 = {undef,undef,undef,undef}
data2 = insert(data1, val1, 0) # {val1, undef, undef, undef}
data3 = insert(data2, val2, 1) # {val1, val2, undef, undef}
data4 = insert(data3, val3, 2) # {val1, val2, val3, undef}
...etc...
```

AlexanderKalistratov: Can't it be something like:
```python
    data_tup = c.context.make_tuple(
        c.builder, types.Tuple(types), values)
```
?

Not sure how to get types though

Hardcode84: meh, not worth it

AlexanderKalistratov:
Please remove `unboxed` from DataFrame model:
https://github.com/IntelPython/sdc/blob/master/sdc/hiframes/pd_dataframe_type.py#L96

PR #668, opened 2020-03-04
Update SDC wheels build

Vyacheslav-Smirnov:
- Update wheel requirements in setup.py and requirements.txt
 - Comment import parquet lib due to it is not used (will remove it in next PR)
 - Change name to sdc in setup.py (this name is used during wheels build and should be as conda)


Vyacheslav-Smirnov:
@AlexanderKalistratov , could you please look at this PR?
If you are ok - then I will merge it.

requirements.txt
```diff
@@ -1,3 +1,5 @@
 scipy
-pandas>=0.23
-numba>=0.44
+numpy>=1.16
+pandas==0.25.3
+pyarrow==0.15.1
+numba==0.48
```
AlexanderKalistratov: I'm not sure how it works. We need numba from specific channel. Can we somehow check it?

sdc/io/parquet_pio.py
```diff
@@ -336,15 +336,15 @@ def generic(self, args, kws):
...
-if _has_pyarrow:
-    from .. import parquet_cpp
-    ll.add_symbol('get_arrow_readers', parquet_cpp.get_arrow_readers)
-    ll.add_symbol('del_arrow_readers', parquet_cpp.del_arrow_readers)
-    ll.add_symbol('pq_read', parquet_cpp.read)
-    ll.add_symbol('pq_read_parallel', parquet_cpp.read_parallel)
-    ll.add_symbol('pq_get_size', parquet_cpp.get_size)
-    ll.add_symbol('pq_read_string', parquet_cpp.read_string)
-    ll.add_symbol('pq_read_string_parallel', parquet_cpp.read_string_parallel)
+# if _has_pyarrow:
```
AlexanderKalistratov: It is better to delete, than comment

setup.py
```diff
@@ -343,7 +344,13 @@ def run(self):
       author='Intel Corporation',
       packages=find_packages(),
       package_data={'sdc.tests': ['*.bz2'], },
-      install_requires=['numba'],
+      install_requires=[
+          'scipy',
+          'numpy>=1.16',
+          'pandas==0.25.3',
+          'pyarrow==0.15.1',
+          'numba==0.48'
```
AlexanderKalistratov: Can't we get it from requirements.txt somehow?

requirements.txt
Vyacheslav-Smirnov: The numba should be taken from intel channel.
I expect that we can deal with it using the right channel order in install command

sdc/io/parquet_pio.py
Vyacheslav-Smirnov: I am planning another PR to remove this opencv deps as well

setup.py
Vyacheslav-Smirnov: Requirements is just a file which can be used for packages installation
Main requirements are set upped here

Vyacheslav-Smirnov:
Thank you, so I am going to merge the PR

PR #672, opened 2020-03-06
Implement private df method _set_column which currently sets non-existing column to the df

densmirn:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1628,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+    else:
+        func_lines += ['  length = len(value)']
+    func_lines += df_index_codelines(self)
+
+    results = []
+    for i, col in enumerate(self.columns):
+        res_data = f'res_data_{i}'
+        func_lines += [
+            f'  data_{i} = get_dataframe_data(self, {i})',
+            f'  {res_data} = pandas.Series(data_{i}, index=res_index, name="{col}")',
```
akharche: I'm lost, where you declare "res_index"?

densmirn: `res_index` is defined inside of `df_index_codelines()`. Maybe it make sense to use expression `self.index` instead of `res_index`. @akharche what do you think about that?

akharche: Got it! But imho it is not clear. May be, df_index_codelines() should return the appropriate index codeline that is set to res_index

densmirn: You are right, but again, maybe it make sense to use expression `self.index` instead of `res_index`?

densmirn: I can't use `self.index`, because in case of empty input DataFrame I should define length of index based on length of input value used for setting column. I left comment in the code about what is returned by `df_index_codelines()`.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1628,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+
+@sdc_overload_method(DataFrameType, '_set_column')
+def df_set_column_overload(self, key, value):
+    """
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+
+    Limitations
+    -----------
+    - Supported setting item only for non-empty DataFrame.
```
```suggestion
    - Supports setting a columns in a non-empty DataFrame as a 1D array.
```

sdc/tests/test_dataframe.py
```diff
@@ -572,6 +573,51 @@ def test_impl(df, arr):
...
+        all_data = [{'A': [0, 1, 2], 'C': [0., np.nan, np.inf]}, {}]
+        key, value = 'B', np.array([1., -1., 0.])
+
+        for data in all_data:
+            with self.subTest(data=data):
+                df_ref = pd.DataFrame(data)
+                df_jit = df_ref.copy(deep=True)
+                df_jit = sdc_func(df_jit, key, value)
+                test_impl(df_ref, key, value)
+                pd.testing.assert_frame_equal(df_jit, df_ref)
```
kozlov-alexey: Maybe use two different names for the input and two different names for results, not mixing them, e.g?
```
                df1 = pd.DataFrame(data)
                df2 = df1 .copy(deep=True)
                test_impl(df1, key, value)
                result = df1  # in pandas setitem modifies original DF
                result_ref = sdc_func(df2, key, value)  # in SDC new DF is returned and original is unchanged
                pd.testing.assert_frame_equal(result, result_ref)
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1627,3 +1628,108 @@ def sdc_pandas_dataframe_groupby_impl(self, by=None, axis=0, level=None, as_inde
...
+        func_lines += df_length_codelines(self)
+        func_lines += [
+            '  if length == 0:',
+            '    raise SDCLimitation("Could not set item for DataFrame with empty columns")',
+            '  elif length != len(value):',
+            '    raise ValueError("Length of values does not match length of index")',
+        ]
+    else:
+        func_lines += ['  length = len(value)']
+    func_lines += df_index_codelines(self)  # provide res_index = ...
```
kozlov-alexey: So the lines df_index_codelines produces actually use length defined above, that's not very good. 
Do you think it would be better if it's impl is changed to take str variable length as a param and use where needed it if it's not None?

PR #677, opened 2020-03-08
Add simple script for benchmark execution

Vyacheslav-Smirnov:
- Separate script
 - Use files from Intel Anaconda channel
 - Add modules list as parameter


buildscripts/run_benchmarks.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def run_benchmarks(sdc_utils, module_list, num_threads_list):
+    os.chdir(str(sdc_utils.src_path.parent))
+
+    for module in module_list:
+        for num_threads in num_threads_list:
+            os.environ['NUMBA_NUM_THREADS'] = num_threads
+            sdc_utils.log_info(f'Run Intel SDC benchmarks on {num_threads} threads', separate=True)
+            run_command(f'python -W ignore -m sdc.runtests {module}')
```
densmirn: Where the function is imported from?

buildscripts/run_benchmarks.py
```diff
@@ -0,0 +1,62 @@
...
+
+from pathlib import Path
+from utilities import SDC_Build_Utilities
+
+
+def run_benchmarks(sdc_utils, module_list, num_threads_list):
+    os.chdir(str(sdc_utils.src_path.parent))
+
+    for module in module_list:
+        for num_threads in num_threads_list:
```
densmirn: You could combine the loops using `itertools.product`.

buildscripts/run_benchmarks.py
Vyacheslav-Smirnov: It is a bug, going to update it

buildscripts/run_benchmarks.py
Vyacheslav-Smirnov: To my mind 2 loops here are more obvious than using itertools.product

buildscripts/run_benchmarks.py
```diff
@@ -0,0 +1,63 @@
...
+            os.environ['NUMBA_NUM_THREADS'] = num_threads
+            sdc_utils.log_info(f'Run Intel SDC benchmarks on {num_threads} threads', separate=True)
+            sdc_utils.run_command(f'python -W ignore -m sdc.runtests -v {module}')
```
PokhodenkoSA: Actually `module` could be any parameters for Python unittest module. I usually filter my tests like: `sdc.tests.test_series -k series_max`

Vyacheslav-Smirnov: Then I will add a separate parameter for "args"

PR #679, opened 2020-03-10
Scale getitem by Boolean mask (array or Series) for Series and DFs

kozlov-alexey:
New implementation not slower than python on 1 thread, scales on 4 cores, but doesn't scale any further:

  |   |   |   | median | min | max | compile | boxing
-- | -- | -- | -- | -- | -- | -- | -- | --
name | nthreads | type | size |   |   |   |   |  
Series.getitem_idx_bool_array | 1 | Python | 1E+07 | 0.282 | 0.2362 | 0.2978 | NaN | NaN
  |   | SDC | 1E+07 | 0.2732 | 0.2648 | 0.2805 | 0.4704 | 0.0012
  | 4 | SDC | 1E+07 | 0.1273 | 0.1205 | 0.1304 | 0.4575 | 0.0007
  | 8 | SDC | 1E+07 | 0.1396 | 0.1332 | 0.1614 | 0.483 | 0.0012
  | 16 | SDC | 1E+07 | 0.1292 | 0.123 | 0.1514 | 0.4871 | 0.0011
Series.getitem_idx_bool_series | 1 | Python | 1E+07 | 0.306 | 0.2906 | 0.3822 | NaN | NaN
  |   | SDC | 1E+07 | 0.3292 | 0.3225 | 0.3915 | 0.4856 | 0.0013
  | 4 | SDC | 1E+07 | 0.1028 | 0.0966 | 0.1424 | 0.4662 | 0.0008
  | 8 | SDC | 1E+07 | 0.1096 | 0.0993 | 0.1418 | 0.4797 | 0.0014
  | 16 | SDC | 1E+07 | 0.1362 | 0.1185 | 0.1392 | 0.5224 | 0.0021

  |   |   |   | median | min | max | compile | boxing
-- | -- | -- | -- | -- | -- | -- | -- | --
name | nthreads | type | size |   |   |   |   |  
DataFrame.getitem_idx_bool_array | 1 | Python | 1E+07 | 0.8207 | 0.8069 | 0.8369 | NaN | NaN
  |   | SDC | 1E+07 | 0.8649 | 0.8586 | 0.8872 | 0.6992 | 1.2021
  | 4 | SDC | 1E+07 | 0.4043 | 0.3912 | 0.422 | 0.535 | 0.5604
  | 8 | SDC | 1E+07 | 0.3538 | 0.3458 | 0.3788 | 0.7094 | 1.1428
DataFrame.getitem_idx_bool_series | 1 | Python | 1E+07 | 0.8197 | 0.8019 | 0.832 | NaN | NaN
  |   | SDC | 1E+07 | 0.6764 | 0.6649 | 0.6969 | 0.4725 | 1.2326
  | 4 | SDC | 1E+07 | 0.1272 | 0.1256 | 0.1766 | 0.461 | 0.5541
  | 8 | SDC | 1E+07 | 0.3883 | 0.2524 | 0.4399 | 0.5251 | 1.1307

### DIFF
--- a/sdc/datatypes/common_functions.py
+++ b/sdc/datatypes/common_functions.py
@@ -663,12 +663,11 @@ def sdc_reindex_series_overload(arr, index, name, by_index):
     data_is_str_arr = isinstance(arr.dtype, types.UnicodeType)
 
     def sdc_reindex_series_impl(arr, index, name, by_index):
-        res_index = by_index.copy()
         if data_is_str_arr == True:  # noqa
-            res_data_as_list = [''] * len(by_index)
+            _res_data = [''] * len(by_index)
             res_data_nan_mask = numpy.zeros(len(by_index), dtype=types.bool_)
         else:
-            res_data = numpy.empty(len(by_index), dtype=data_dtype)
+            _res_data = numpy.empty(len(by_index), dtype=data_dtype)
 
         # build a dict of self.index values to their positions:
         map_index_to_position = Dict.empty(
@@ -686,7 +685,7 @@ def sdc_reindex_series_overload(arr, index, name, by_index):
         for i in numba.prange(len(by_index)):
             if by_index[i] in map_index_to_position:
                 pos_in_self = map_index_to_position[by_index[i]]
-                res_data[i] = arr[pos_in_self]
+                _res_data[i] = arr[pos_in_self]
                 if data_is_str_arr == True:  # noqa
                     res_data_nan_mask[i] = isna(arr, i)
             else:
@@ -697,10 +696,12 @@ def sdc_reindex_series_overload(arr, index, name, by_index):
             raise IndexingError(msg)
 
         if data_is_str_arr == True:  # noqa
-            res_data = create_str_arr_from_list(res_data_as_list)
+            res_data = create_str_arr_from_list(_res_data)
             str_arr_set_na_by_mask(res_data, res_data_nan_mask)
+        else:
+            res_data = _res_data
 
-        return pandas.Series(data=res_data, index=res_index, name=name)
+        return pandas.Series(data=res_data, index=by_index, name=name)
 
     return sdc_reindex_series_impl
 
--- a/sdc/datatypes/common_functions.py
+++ b/sdc/datatypes/common_functions.py
@@ -32,6 +32,7 @@
 
 import numpy
 import pandas
+from pandas.core.indexing import IndexingError
 
 import numba
 from numba.targets import quicksort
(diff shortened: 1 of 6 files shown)


sdc/datatypes/common_functions.py
```diff
@@ -641,3 +644,59 @@ def _almost_equal_impl(x, y):
...
+            else:
+                map_index_to_position[value] = i
+
+        index_mismatch = 0
+        for i in numba.prange(len(by_index)):
+            if by_index[i] in map_index_to_position:
+                pos_in_self = map_index_to_position[by_index[i]]
+                res_data[i] = arr[pos_in_self]
+                if data_is_str_arr == True:  # noqa
+                    res_data_nan_mask[i] = isna(arr, i)
```
densmirn: Looks `res_data` isn't defined in case when `data_is_str_arr == True`. So you could add else-block where you could fill `res_data`.

sdc/datatypes/common_functions.py
```diff
@@ -641,3 +644,59 @@ def _almost_equal_impl(x, y):
...
+            value_type=types.int32
+        )
+
+        for i, value in enumerate(index):
+            if value in map_index_to_position:
+                raise ValueError("cannot reindex from a duplicate axis")
+            else:
+                map_index_to_position[value] = i
+
+        index_mismatch = 0
```
densmirn: Looks `index_mismatch` could be of a boolean type.

sdc/datatypes/common_functions.py
kozlov-alexey: @densmirn Thank you! It seems that this case (string series reindexing) is not used in SDC functions implementation yet (getitem uses only bool series), hence it's slipped through functional tests. And that's a case that shows how much we need unit tests for internal functions.

sdc/datatypes/common_functions.py
kozlov-alexey: @densmirn Not sure if it's really possible, because it's used inside prange and simply assigning  True value would be a race between threads. I prefer to leave it as is, so that it's reduced to a number of faults occurred during parallel iteration.

sdc/datatypes/common_functions.py
```diff
@@ -646,3 +649,60 @@ def _almost_equal_impl(x, y):
         return abs(x - y) <= numpy.finfo(common_dtype).eps
 
     return _almost_equal_impl
+
+
+def sdc_reindex_series(arr, index, name, by_index):
+    pass
+
+
+@sdc_overload(sdc_reindex_series, jit_options={'parallel': True})
```
AlexanderKalistratov: you don't need `jit_options={'parallel': True}`.
It overrides default behavior - based on environment variable

kozlov-alexey: Yes, that's valid point, I forgot to remove it. Let's fix that in following PR.

PR #681, opened 2020-03-12
Reduce ir for some functions

Hardcode84:
* Manually inline `get_dataframe_data` function calls to reduce excessive increfs/decrefs for large dataframes

`get_dataframe_data`  function itself is still here to avoid breaking any ongoing PR

### DIFF
--- a/sdc/datatypes/hpat_pandas_dataframe_functions.py
+++ b/sdc/datatypes/hpat_pandas_dataframe_functions.py
@@ -1109,8 +1109,8 @@ def sdc_pandas_dataframe_drop_codegen(func_name, func_args, df, drop_cols):
             break
 
     for column_id, column_name in enumerate(saved_df_columns):
-        func_text.append(f'new_col_{column_id}_data_{"df"} = get_dataframe_data({"df"}, '
-                         f'{df_columns_indx[column_name]})')
+        func_text.append(f'new_col_{column_id}_data_{"df"} = {"df"}._data['
+                         f'{df_columns_indx[column_name]}]')
         column_list.append((f'new_col_{column_id}_data_df', column_name))
 
     data = ', '.join(f'"{column_name}": {column}' for column, column_name in column_list)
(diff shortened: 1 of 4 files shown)


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1279,8 +1279,7 @@ def df_getitem_slice_idx_main_codelines(self, idx):
     for i, col in enumerate(self.columns):
         res_data = f'res_data_{i}'
         func_lines += [
-            f'  data_{i} = get_dataframe_data(self, {i})',
-            f'  {res_data} = pandas.Series(data_{i}[idx], index=res_index[idx], name="{col}")'
+            f'  {res_data} = pandas.Series((self._data[{i}])[idx], index=res_index[idx], name="{col}")'
```
densmirn: Maybe skip constructing of Series and use array directly?

Hardcode84: I've tried and it causes some tests to fail. And also it reduces ir only for about 10Mb (175->165Mb)

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1015,8 +1015,8 @@ def sdc_pandas_dataframe_drop_impl(df, labels=None, axis=0, index=None, columns=
             break
 
     for column_id, column_name in enumerate(saved_df_columns):
-        func_text.append(f'new_col_{column_id}_data_{"df"} = get_dataframe_data({"df"}, '
-                         f'{df_columns_indx[column_name]})')
+        func_text.append(f'new_col_{column_id}_data_{"df"} = {"df"}._data['
```
kozlov-alexey: Curved brackets aren't necessary here, since we use quoted str in them anyway. It's the same as:
### 
    for column_id, column_name in enumerate(saved_df_columns):
        func_text.append(f'new_col_{column_id}_data = df._data[{df_columns_indx[column_name]}]')
        column_list.append((f'new_col_{column_id}_data', column_name))

Hardcode84:
ping @AlexanderKalistratov @kozlov-alexey

PR #698, opened 2020-03-16
correct doc for Series.idxmin

Rubtsowa:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3878,6 +3878,10 @@ def hpat_pandas_series_idxmin(self, axis=None, skipna=None):
 
     Pandas API: pandas.Series.idxmin
 
+    Limitations
+    -----------
+    Parameters axis, skipna are currently unsupported by Intel Scalable Dataframe Compiler
```
densmirn: `skipna` is supported restrictly. So reflect it here.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3878,6 +3878,11 @@ def hpat_pandas_series_idxmin(self, axis=None, skipna=None):
 
     Pandas API: pandas.Series.idxmin
 
+    Limitations
+    -----------
+    Parameter axis is currently unsupported by Intel Scalable Dataframe Compiler.
+    Parameter skipna is unsupported with str data.
```
densmirn:
```suggestion
    Parameter ``axis`` is supported only with default value ``None``.
    Parameter ``skipna`` cannot be ``False`` with data of string type.
```

PR #703, opened 2020-03-16
Read csv

PokhodenkoSA:
This PR is for assessing the proposed solution.
This PR extends read_csv with:
1. names
2. dtype
3. usecols


sdc/io/csv_ext.py
```diff
@@ -470,43 +593,59 @@ def _gen_csv_reader_py_pyarrow(col_names, col_typs, usecols, sep, typingctx, tar
...
+        signature = "filepath_or_buffer"
     func_text = "def csv_reader_py({}):\n".format(signature)
-    func_text += "  with objmode({}):\n".format(typ_strs)
-    func_text += "    df = pandas_read_csv(fname, names={},\n".format(col_names)
-    func_text += "       parse_dates=[{}],\n".format(date_inds)
-    func_text += "       dtype={{{}}},\n".format(pd_dtype_strs)
-    func_text += "       skiprows={},\n".format(skiprows)
-    func_text += "       usecols={}, sep='{}')\n".format(usecols, sep)
-    for cname in col_names:
+    func_text += "  with objmode({}):\n".format(nb_objmode_vars)
```
Hardcode84: nit: You can wrap these lines into into list of strings to avoid `func_text +=` copypaste

Hardcode84:
As I see you didn't add `simplify` call after rewrite yet and there can be a potential problem with it, you are working on block level in rewrites, but simplify should be applied on function level IIRC.

Hardcode84:
> 
> 
> As I see you didn't add `simplify` call after rewrite yet and there can be a potential problem with it, you are working on block level in rewrites, but simplify should be applied on function level IIRC.

nevermind, just saw `remove_unused_recursively`

Hardcode84:
I am ok with these changes but what with the failing CI?

PokhodenkoSA:
> I am ok with these changes but what with the failing CI?

I have found bug in corner case. When column name is empty in the file then Pandas and PyArrow read it differently. I have fixed it. Waiting for results.

sdc/rewrites/read_csv_consts.py
```diff
@@ -0,0 +1,161 @@
...
+        # TODO: 1. save instructions of build_map, build_list for read_csv params
+        # 2. check that vars are used only in read_csv
+        # 3. replace vars with build_tuple inplace
+
+        self.func_ir = func_ir
+        self.block = block
+        self.consts = consts = {}
+
+        # Find all assignments with a right-hand read_csv() call
+        for inst in block.find_insts(ir.Assign):
```
AlexanderKalistratov: you could use [find_operations(block, 'call')](https://github.com/IntelPython/sdc/blob/master/sdc/rewrites/ir_utils.py#L48) instead.

sdc/rewrites/read_csv_consts.py
```diff
@@ -0,0 +1,161 @@
...
+
+    require(isinstance(var, ir.Var))
+    var_def = get_definition(func_ir, var)
+    require(isinstance(var_def, ir.Expr))
+    build_ops = ['build_tuple', 'build_list', 'build_set', 'build_map']
+    require(var_def.op in build_ops)
+    return var_def.items, var_def.op
+
+
+def _new_definition(func_ir, var, value, loc):
```
AlexanderKalistratov: you could use [make_assign](https://github.com/IntelPython/sdc/blob/master/sdc/rewrites/ir_utils.py#L100)

sdc/rewrites/read_csv_consts.py
```diff
@@ -0,0 +1,161 @@
...
+            if expr.op != 'call':
+                continue
+            call = guard(find_callname, func_ir, expr)
+            if call not in self._pandas_read_csv_calls:
+                continue
+            # protect from repeat rewriting
+            if hasattr(inst, 'consts'):
+                continue
+            # collect constant parameters
+            for key, var in expr.kws:
```
AlexanderKalistratov: Only named parameters are stored in .kws. Positional are stored in .args. You need  to check both.
I used [get_call_parameters](https://github.com/IntelPython/sdc/blob/master/sdc/rewrites/ir_utils.py#L331) to extract both into dict:
https://github.com/IntelPython/sdc/blob/master/sdc/rewrites/dataframe_constructor.py#L74

But it requires to define parameters list:
https://github.com/IntelPython/sdc/blob/master/sdc/rewrites/dataframe_constructor.py#L56

Probably it could be extracted somehow from function definition?

sdc/rewrites/read_csv_consts.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def apply(self):
+        new_block = self.block.copy()
+        new_block.clear()
+        vars_to_remove = []
+
+        for inst in self.block.body:
+            if inst in self.consts:
+                # protect from repeat rewriting
+                inst.consts = consts = self.consts[inst]
```
AlexanderKalistratov: alternatively you could in `match` check that specific parameter is actually a dict, and if it is not - do not trigger `match`

sdc/datatypes/hpat_pandas_functions.py
```diff
@@ -38,60 +38,261 @@
...
+        any([
+            isinstance(names, types.Tuple) and isinstance(usecols, types.Tuple),
+            isinstance(names, types.Tuple) and isinstance(usecols, (types.Omitted, type(None))),
+            isinstance(names, (types.Omitted, type(None))) and isinstance(usecols, types.Tuple),
+        ]),
+        isinstance(header, types.Omitted) or header == 'infer',
+    ])
+
+    # cannot create function if parameters provide not enough info
+    assert infer_from_file or infer_from_params
```
AlexanderKalistratov: Could you please replace assert with exception?

sdc/hiframes/hiframes_untyped.py
```diff
@@ -759,7 +759,8 @@ def get_dtypes(df):
     @staticmethod
     def infer_column_names_and_types_from_constant_filename(fname_const, skiprows, col_names, sep=','):
```
AlexanderKalistratov: Does it used somewhere?

PokhodenkoSA: It is part of the old code. I think it is possible to clean it in new PR.

sdc/datatypes/hpat_pandas_functions.py
PokhodenkoSA: Fixed

sdc/hiframes/hiframes_untyped.py
PokhodenkoSA: I will remove it from PR

sdc/rewrites/read_csv_consts.py
PokhodenkoSA: I have found make_assign not very obvious. But I have tryed to replace it with make_assign and it works. I am ok to use make_assign.

sdc/rewrites/read_csv_consts.py
PokhodenkoSA: Fixed. Use checking.

sdc/rewrites/read_csv_consts.py
PokhodenkoSA: FIxed. Use find_operations.

sdc/io/csv_ext.py
PokhodenkoSA: I have tried it. Code is little bit more clear in this implementation even with repetition.

sdc/rewrites/read_csv_consts.py
PokhodenkoSA: Read_csv ususally call via named parametrs. So it is not very critical now. I can do it in separate PR>

PR #712, opened 2020-03-18
Improve docstring for Series.add()

akharche:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2516,7 +2516,16 @@ def hpat_pandas_series_add(self, other, level=None, fill_value=None, axis=0):
...
     Limitations
     -----------
-    - Parameters level, fill_value are currently unsupported by Intel Scalable Dataframe Compiler
+    Parameter ``level`` is supported only with default value ``None``.
+
+    Parameter ``fill_value`` is supported only with default value ``None``.
+
+    Parameter ``axis`` is supported only with default value ``0``.
+
+    Parameter ``other`` can be one of the following:
```
densmirn: Description of the parameters `level` and `fill_value` could be combined to `Parameter ``level`` and ``fill_value`` are supported only with default value ``None``.`. Maybe remove excess empty lines?

PokhodenkoSA: If this parameters should be always in default values why not to say that we do not support this parameters? It means that we recommend to omit them:
```suggestion
    - Parameter ``level``, ``fill_value`` and ``axis`` are currently unsupported by Intel Scalable Dataframe Compiler.
```

densmirn: Actually we cannot say that a parameter is unsupported at all, at the least the default value is valid.

PokhodenkoSA: @samaid could you propose please the best way to describe this kind of limitations?

PokhodenkoSA: As we discussed with @samaid the best way is list and short form:
- Parameters ``level``, ``fill_value`` and ``axis`` are unsupported.

PR #714, opened 2020-03-18
Doc for Series.argsort

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4558,7 +4558,7 @@ def hpat_pandas_series_argsort(self, axis=0, kind='quicksort', order=None):
 
     Limitations
     -----------
-    - Parameters kind, order are currently unsupported by Intel Scalable Dataframe Compiler
+    Parameters ``kind``, ``order`` and ``axis`` are supported only with default value ``None``.
```
densmirn: Only parameter `order` has default value `None`.

PR #715, opened 2020-03-18
Fix docstring for Series.count()

akharche:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4448,7 +4448,7 @@ def hpat_pandas_series_count(self, level=None):
 
     Limitations
     -----------
-    - Parameter level is currently unsupported by Intel Scalable Dataframe Compiler
+    Parameter ``level`` is supported only with default value ``None``.
```
PokhodenkoSA: Is it new correct way to define unsupported parameters?

densmirn: I'm following the point `shouldn't be phrase unsupported in SDC` in the checklist.

densmirn: Moreover we cannot say that a parameter is unsupported at all, at the least the default value is valid

PokhodenkoSA: This point in checklist was about ``Unsupported by Intel SDC.`` which appears when there is no corresponding documentation in SDC for method from Pandas.

akharche: Fixed

PR #717, opened 2020-03-18
Doc for Series.min

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3486,7 +3486,7 @@ def hpat_pandas_series_min(self, axis=None, skipna=None, level=None, numeric_onl
 
     Limitations
     -----------
-    - Parameters level, numeric_only are currently unsupported by Intel Scalable Dataframe Compiler
+    Parameters ``level``, ``numeric_only``, ``axis`` are supported only with default value ``None``.
```
PokhodenkoSA: I think create Limitation**s** as a list is a good practice.
```suggestion
    - Parameters ``level``, ``numeric_only``, ``axis`` are supported only with default value ``None``.
```

PokhodenkoSA:
```suggestion
    Parameters ``level``, ``numeric_only`` and ``axis`` are supported only with default value ``None``.
```

PokhodenkoSA: We should say that this parameters should not be used (should be omitted) or should be in default values. 
```suggestion
    Parameters ``level``, ``numeric_only``, ``axis`` are currently unsupported by Intel Scalable Dataframe Compiler.
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3486,7 +3486,7 @@ def hpat_pandas_series_min(self, axis=None, skipna=None, level=None, numeric_onl
 
     Limitations
     -----------
-    Parameters ``level``, ``numeric_only``, ``axis`` are supported only with default value ``None``.
+    Parameters ``level``, ``numeric_only`` and ``axis`` are currently unsupported by Intel Scalable Dataframe Compiler.
```
PokhodenkoSA: Sort in order of function parameters
```suggestion
    Parameters ``axis``, ``level`` and ``numeric_only`` are currently unsupported by Intel Scalable Dataframe Compiler.
```

PR #718, opened 2020-03-18
Doc for Series.sort_values

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4708,8 +4708,8 @@ def hpat_pandas_series_sort_values(self, axis=0, ascending=True, inplace=False,
 
     Limitations
     -----------
-    - Parameter inplace is currently unsupported by Intel Scalable Dataframe Compiler
-    - Parameter kind currently only 'mergesort' and 'quicksort' are supported as literal values.
+    - Parameter ``inplace`` is supported only with default value ``None``.
+    - Parameter ``kind`` currently only ``'mergesort'`` and ``'quicksort'`` are supported as literal values.
```
densmirn: It looks hard, maybe replace it with something like `Parameter ``kind`` is supported only with values ``'mergesort'`` and ``'quicksort'``.`?

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4708,8 +4708,8 @@ def hpat_pandas_series_sort_values(self, axis=0, ascending=True, inplace=False,
 
     Limitations
     -----------
-    - Parameter inplace is currently unsupported by Intel Scalable Dataframe Compiler
-    - Parameter kind currently only 'mergesort' and 'quicksort' are supported as literal values.
+    - Parameter ``inplace`` is supported only with default value ``None``.
```
densmirn: Parameter `inplace` has another value as default.

PR #721, opened 2020-03-18
Fix docstring for Series.div()

akharche:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2988,7 +2988,8 @@ def hpat_pandas_series_div(self, other, level=None, fill_value=None, axis=0):
 
     Limitations
     -----------
-    - Parameters level, fill_value are currently unsupported by Intel Scalable Dataframe Compiler
+    Parameters ``level``, ``fill_value`` are supported only with default value ``None``.
+    Parameter ``axis`` is supported only with default value ``0``.
```
densmirn: In this case parameters `level`, `fill_value` and `axis` could be omitted only according to the code in typing stage. So you need to change either the docstring or the code in typing stage.

PR #723, opened 2020-03-18
Docs for pandas.Series()

PokhodenkoSA:
image

Note:
1. Removed filter for sdc.hiframes because overload(pd.Series) in it
2. I know only unsupported parameters as limitations.
3. Examples for series are in other functions.
4. See also - no in Pandas docs, so no in SDC docs.


PokhodenkoSA:
image

AlexanderKalistratov:
I don't think we are supporting iterable and dicts as `data` parameter

PokhodenkoSA:
> I don't think we are supporting iterable and dicts as `data` parameter

Could you give an advice how to find what is the border of

PR #730, opened 2020-03-19
Docs for Series.mean()

PokhodenkoSA:
image


PokhodenkoSA:
> Wouldn't you like to add block `seealso`?

There is not `seealso` in Pandas docs.

PR #731, opened 2020-03-19
Docs for Series.prod()

PokhodenkoSA:
image


PokhodenkoSA:
> Wouldn't you like to add block `seealso`?

There is no seealso in Pandas docs.

PR #732, opened 2020-03-19
Docs for Series.sum()

PokhodenkoSA:
image


PokhodenkoSA:
> Is't ok that block `seealso` contains link to itself?

The same way as Pandas docs do. `seealso` is taken from Pandas docs.

PokhodenkoSA:
Added more seealso from Pandas docs.

PR #734, opened 2020-03-19
Fix docstring for Series.max()

akharche:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3572,7 +3572,7 @@ def hpat_pandas_series_max(self, axis=None, skipna=None, level=None, numeric_onl
 
     Limitations
     -----------
-    - Parameters level, numeric_only are currently unsupported by Intel Scalable Dataframe Compiler
+    Parameters axis, skipna, level, numeric_only are currently unsupported.
```
densmirn: Let's use double quotes ```` around of parameters names. Looks `skipna` is supported.

PR #742, opened 2020-03-20
correct doc for Series.setitem

Rubtsowa:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -462,14 +462,16 @@ def sdc_pandas_series_setitem(self, idx, value):
     """
     Intel Scalable Dataframe Compiler User Guide
     ********************************************
-    Pandas API: pandas.Series.__setitem__
+    Pandas API: pandas.Series.setitem
```
kozlov-alexey: Why not pandas.Series.__setitem__? From the python user perspective this will look like a method called setitem, not __setitem__ operator.

densmirn: Likely it works like a anchor for documentation generator when the generator finds the name in rst file `_api_ref.pandas.series_templ.rst`.

PR #744, opened 2020-03-20
Initial attempt to parallelize groupby

kozlov-alexey:
With this PR total exec time of groping operation of groupby is reduced x3 times, also threaded parallelism is added via separation of work of filling groupby dict among threads (scales poorly though):


sdc/datatypes/hpat_pandas_groupby_functions.py
```diff
@@ -45,12 +45,28 @@
...
+        return left
+
+    for key, right_group_list in right.items():
+        left_group_list = left.get(key)
+        if left_group_list is None:
+            left[key] = right_group_list
+        else:
+            left_group_list.extend(right_group_list)
+
+    return left
```
densmirn: Is `left` returned changed?

kozlov-alexey: @densmirn Yes, left holds the result of merging. I will add a docstring and change the name to merge_groupby_dicts_inplace.

PR #751, opened 2020-03-23
Updating docstrings and adding groupby examples

kozlov-alexey:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -5280,6 +5277,12 @@ def hpat_pandas_series_describe(self, percentiles=None, include=None, exclude=No
             or percentiles is None):
         ty_checker.raise_exc(percentiles, 'list-like', 'percentiles')
 
+    if not (isinstance(include, (types.Omitted, types.NoneType)) or include is None):
+        raise TypingError('{} Unsupported parameters. Given include: {}'.format(_func_name, include))
```
densmirn: Why not use `ty_checker.raise_exc()` instead of explicit raise here and below?

AlexanderKalistratov:
So, what are we doing with this one?

sdc/datatypes/hpat_pandas_series_functions.py
kozlov-alexey: @densmirn It will say that it expects certain types for the parameter, but not that it's unsupported.

kozlov-alexey: > So, what are we doing with this one?

I think this can be merged once CI passes. We have agreed with @densmirn that we'll address that issue with TypingError later and change that to SDCLimitation raised by special ty_checker method.

PR #761, opened 2020-03-24
Add example to Basic Usage in doc

densmirn:
image

### DIFF
--- a/buildscripts/run_examples.py
+++ b/buildscripts/run_examples.py
@@ -34,24 +34,15 @@ from pathlib import Path
 from utilities import SDC_Build_Utilities
 
 
-examples_to_skip = {'basic_usage_nyse_predict.py'}
-
-
 def run_examples(sdc_utils):
     total = 0
     passed = 0
     failed = 0
-    skipped = 0
     failed_examples = []
 
     os.chdir(str(sdc_utils.examples_path))
     for sdc_example in Path('.').glob('**/*.py'):
         total += 1
-
-        if sdc_example.name in examples_to_skip:
-            skipped += 1
-            continue
-
         sdc_example = str(sdc_example)
         try:
             sdc_utils.log_info(sdc_utils.line_double)
@@ -65,8 +56,7 @@ def run_examples(sdc_utils):
             passed += 1
             sdc_utils.log_info(f'{sdc_example} PASSED')
 
-    summary_msg = f'SDC examples summary: {total} RUN, {passed} PASSED, {failed} FAILED, {skipped} SKIPPED'
-    sdc_utils.log_info(summary_msg, separate=True)
+    sdc_utils.log_info(f'SDC examples summary: {total} RUN, {passed} PASSED, {failed} FAILED', separate=True)
     for failed_example in failed_examples:
         sdc_utils.log_info(f'FAILED: {failed_example}')
 
@@ -89,7 +79,7 @@ if __name__ == '__main__':
     sdc_utils = SDC_Build_Utilities(args.python, args.sdc_channel)
     sdc_utils.log_info('Run Intel(R) SDC examples', separate=True)
     sdc_utils.log_info(sdc_utils.line_double)
-    sdc_utils.create_environment(['scipy'])
+    sdc_utils.create_environment(['scipy', 'sklearn'])
     sdc_utils.install_conda_package(['sdc'])
 
     run_examples(sdc_utils)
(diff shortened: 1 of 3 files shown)


akharche:
@densmirn Maybe change description somehow to avoid words repetition?

The code below illustrates a typical ML workflow that consists of data pre-processing and predicting stages. Intel SDC is intended to compile data pre-processing stages that includes reading dataset from a csv file, data filtering and performing Pearson Correlation operation. The prediction based on   gradient boosting regression module is made using scikit-learn module.

akharche:
@samaid Can we use your script based on "prices" dataset here as an example of SDC usage?

Hardcode84:
Tests failed.

When I tried to run it locally on pure pandas (0.25.3) it is also failed. Had to change `columns=('symbol', 'volume')` -> `columns=['symbol', 'volume']`

densmirn:
> Tests failed.
> 
> When I tried to run it locally on pure pandas (0.25.3) it is also failed. Had to change `columns=('symbol', 'volume')` -> `columns=['symbol', 'volume']`

The example shouldn't be ran on pure Pandas. We need `sklearn` to run the example. So these are 2 options. First one is to skip the example somehow. Second one is to install `sklearn` to CI environment. @Vyacheslav-Smirnov what option is better to select and how to reach selected option?

densmirn:
> > Tests failed.
> > When I tried to run it locally on pure pandas (0.25.3) it is also failed. Had to change `columns=('symbol', 'volume')` -> `columns=['symbol', 'volume']`
> 
> The example shouldn't be ran on pure Pandas. We need `sklearn` to run the example. So these are 2 options. First one is to skip the example somehow. Second one is to install `sklearn` to CI environment. @Vyacheslav-Smirnov what option is better to select and how to reach selected option?

Exclude the example from common pipeline by skipping it.

Vyacheslav-Smirnov:
@densmirn ,
I would prefer not to skip any example - they all should be executed and all should be passed.
To add sklearn (is it scikit-learn??) to CI for example build just add required package to https://github.com/IntelPython/sdc/pull/761/files#diff-12db1f3244c35f97542b6b38613320cfL82
sdc_utils.create_environment(['scipy'])

Vyacheslav-Smirnov:
Also if this example requires additional package this package should be mentioned in the README.rst in sections related to examples.

buildscripts/run_examples.py
```diff
@@ -56,7 +65,8 @@ def run_examples(sdc_utils):
             passed += 1
             sdc_utils.log_info(f'{sdc_example} PASSED')
 
-    sdc_utils.log_info(f'SDC examples summary: {total} RUN, {passed} PASSED, {failed} FAILED', separate=True)
+    summary_msg = f'SDC examples summary: {total} RUN, {passed} PASSED, {failed} FAILED, {skipped} SKIPPED'
+    sdc_utils.log_info(summary_msg, separate=True)
```
Vyacheslav-Smirnov: @densmirn ,
I would prefer not to skip any example - they all should be executed and all should be passed.
To add sklearn (is it scikit-learn??) to CI for example build just add required package to https://github.com/IntelPython/sdc/pull/761/files#diff-12db1f3244c35f97542b6b38613320cfL82
sdc_utils.create_environment(['scipy'])

Vyacheslav-Smirnov: Like sdc_utils.create_environment(['scipy', 'sklearn'])

densmirn: Done!

densmirn: I hope it won't slow down CI.

densmirn: I skipped the example again, because the example uses big external dataset as input. So we are not able to run it in CI. This example is just for documentation and successfully works with downloaded dataset on the local system.

densmirn: @Vyacheslav-Smirnov are you ok with that? If yes, please re-review the PR, if not, let's think up about another solution, but I don't think we have enough time before release.

Vyacheslav-Smirnov: I just do not like the idea of skipping examples in CI) They all should work.
My thought is - can we just include the example in doc without adding it in examples?
Or maybe there is some less restrictive example?
If none of above suits - then lets merge this PR for now because the documentation need to be completed.

densmirn: I would like to let user run the example directly from repo, so the file should be located there. Maybe we can think up about less restrictive example in the future. Now I prefer to merge the PR. Thanks!

PR #700, opened 2020-03-16
correct doc for Series.str.lower

Rubtsowa:

PR #639, opened 2020-02-26
Series str ljust

Rubtsowa:
median       min       max   compile    boxing
name                nthreads type   size   width
Series.str.ljust(1) 1        Python 10000  16     0.000000  0.000000  0.000000       NaN       NaN
                                           64     0.000000  0.000000  0.015686       NaN       NaN
                                           512    0.002062  0.000000  0.010079       NaN       NaN
                                           1024   0.000000  0.000000  0.000000       NaN       NaN
                                    100000 16     0.028421  0.020251  0.040757       NaN       NaN
                                           64     0.028304  0.020301  0.030622       NaN       NaN
                                           512    0.030380  0.020239  0.030559       NaN       NaN
                                           1024   0.031252  0.022146  0.046903       NaN       NaN
                             SDC    10000  16     0.000000  0.000000  0.007000  2.491599  0.000000
                                           64     0.000000  0.000000  0.016000  2.491970  0.000000
                                           512    0.002000  0.000000  0.010000  2.528332  0.000035
                                           1024   0.000000  0.000000  0.016000  2.471732  0.000000
                                    100000 16     0.030000  0.020000  0.031000  2.516526  0.000404
                                           64     0.029000  0.023000  0.031000  2.490221  0.001428
                                           512    0.022000  0.020000  0.033000  2.522679  0.008508
                                           1024   0.016000  0.015000  0.022000  2.561616  0.016234

Series.str.ljust(1) 4        Python 10000  16     0.008054  0.000000  0.010115       NaN       NaN
                                           64     0.008068  0.000000  0.010132       NaN       NaN
                                           512    0.008059  0.000000  0.010123       NaN       NaN
                                           1024   0.010068  0.000000  0.010143       NaN       NaN
                                    100000 16     0.053427  0.046877  0.062499       NaN       NaN
                                           64     0.040517  0.030533  0.048801       NaN       NaN
                                           512    0.029037  0.025412  0.031586       NaN       NaN
                                           1024   0.030822  0.024564  0.041181       NaN       NaN
                             SDC    10000  16     0.000000  0.000000  0.008000  2.499237  0.000108
                                           64     0.000000  0.000000  0.008000  2.544949  0.000045
                                           512    0.000000  0.000000  0.008000  2.557466  0.000000
                                           1024   0.002000  0.000000  0.008000  2.537159  0.000051
                                    100000 16     0.033000  0.031000  0.041000  2.517652  0.015579
                                           64     0.031000  0.022000  0.035000  2.929404  0.012506
[...]

PR #764, opened 2020-03-24
Add perf limitation

1e-to:
I made restrictions for all methods except string methods ljust and lower, @densmirn  where can I provide documentation for them?

### DIFF
--- a/sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
+++ b/sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
@@ -468,6 +468,8 @@ sdc_pandas_dataframe_rolling_apply.__doc__ = sdc_pandas_dataframe_rolling_docstr
     """
     Limitations
     -----------
+    This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
+    between staying in JIT-region with that function or going back to interpreter mode.
     Supported ``raw`` only can be `None` or `True`. Parameters ``args``, ``kwargs`` unsupported.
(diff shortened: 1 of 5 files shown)


densmirn:
> I made restrictions for all methods except string methods ljust and lower, @densmirn where can I provide documentation for them?

https://github.com/IntelPython/sdc/pull/639
https://github.com/IntelPython/sdc/pull/700

sdc/datatypes/hpat_pandas_dataframe_rolling_functions.py
```diff
@@ -575,6 +583,8 @@ def sdc_pandas_dataframe_rolling_var(self, ddof=1):
     """
     Limitations
     -----------
+    This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
```
PokhodenkoSA: I think it is better to convert it to list.

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -1154,6 +1154,8 @@ def hpat_pandas_series_rolling_var(self, ddof=1):
     """
     Limitations
     -----------
+    This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
```
PokhodenkoSA: I think it should be converted to list.

PokhodenkoSA: And in other places where Limitations consist of several statements they should be organized in list.

PR #767, opened 2020-03-24
change check parameter to_strip

Rubtsowa:


sdc/datatypes/hpat_pandas_stringmethods_functions.py
```diff
@@ -943,7 +943,7 @@ def hpat_pandas_stringmethods_lstrip(self, to_strip=None):
     ty_checker = TypeChecker('Method strip().')
     ty_checker.check(self, StringMethodsType)
 
-    if not isinstance(to_strip, (NoneType, StringLiteral, UnicodeType)):
+    if not isinstance(to_strip, (NoneType, StringLiteral, UnicodeType, Omitted)) and to_strip is not None:
```
PokhodenkoSA: I think it was enough to add Omitted, because `to_strip is not None` is equal to check `not isinstance(to_strip, NoneType)`.
```suggestion
    if not isinstance(to_strip, (NoneType, StringLiteral, UnicodeType, Omitted)):
```

Rubtsowa: TypingError: Method strip(). The object to_strip
 given: None
 expected: str

PokhodenkoSA: Yes, you are right. `NoneType` is numba type here. The correct check for my example is `isinstance(to_strip, type(None))`. The better way is to use your check `to_strip is not None`.

PokhodenkoSA:
This PR fixed perf tests.

PR #769, opened 2020-03-25
correct list of examples

Rubtsowa:


densmirn:
The same is generated in https://github.com/IntelPython/sdc/pull/768.

PR #771, opened 2020-03-25
Expand TypeChecker to raise SDCLimitation('...Unsupported...')

densmirn:


sdc/utilities/sdc_typing_utils.py
```diff
@@ -70,9 +84,26 @@ def raise_exc(self, data, expected_types, name=''):
             expected types inserting directly to the exception
         name: :obj:`str`
             name of the parameter
+        exc_cls: :obj:`Exception`
+            class of the exception to be raised
+        """
+        self._raise_exc(self, exc_cls, self.default_tmpl, self.func_name,
+                        name, data, expected_types)
+
+    def raise_unsupported_exc(self, data, name='', exc_cls=TypingError):
```
kozlov-alexey: Maybe use default exception as SDCLimitation, as it seems to be more appropriate in this case?

sdc/utilities/sdc_typing_utils.py
```diff
@@ -70,9 +89,26 @@ def raise_exc(self, data, expected_types, name=''):
             expected types inserting directly to the exception
         name: :obj:`str`
             name of the parameter
+        exc_cls: :obj:`Exception`
+            class of the exception to be raised
+        """
+        self._raise_exc(exc_cls, self.default_tmpl, self.func_name,
+                        name, data, expected_types)
+
+    def raise_unsupported_exc(self, data, name='', exc_cls=SDCLimitation):
```
kozlov-alexey: I was thinking of replacing all raises related to SDC limitations with this, but it seems this only covers case of unsupported parameters, so for instance in series.fillna we have:
```
            raise TypingError('{} Not implemented when Series dtype is {} and\
                 inplace={}'.format(_func_name, self.dtype, inplace))
```
Maybe add in raise_exc and internal func additional parameter e.g. msg, which if exists will be used instead of default msg format?

densmirn: Do you mean to replace call
```python
msg = '{} Not implemented when Series dtype is {} and inplace={}'.format(_func_name, self.dtype, inplace)
raise SDCLimitation(msg)
```
with something like this
```python
msg = '{} Not implemented when Series dtype is {} and inplace={}'.format(_func_name, self.dtype, inplace)
ty_checker.raise_exc(None, None, exc_cls=SDCLimitation, msg=msg)
```
?

I think `TypeChecker` shouldn't be used for such complicated cases, `TypeChecker` doesn't make sense, when it increases the amount of code and makes it harder.
Maybe I didn't catch you correctly, please correct me in this case.

PR #774, opened 2020-03-25
Fix Series docs

1e-to:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -1843,19 +1843,19 @@ def hpat_pandas_series_astype(self, dtype, copy=True, errors='raise'):
```
akharche: dtype two times in caption
:caption: Cast a pandas object to a specified dtype dtype.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3465,7 +3465,7 @@ def hpat_pandas_series_rename(self, index=None, copy=True, inplace=False, level=
 
     Limitations
     -----------
-    - Parameter level is currently unsupported by Intel Scalable Dataframe Compiler
+    - Parameter ``level`` is currently unsupported by Intel Scalable Dataframe Compiler
```
akharche: May be a bit redundant? It's ok in short variant I think
```suggestion
    - Parameter ``level`` is currently unsupported.
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -3972,8 +3972,8 @@ def hpat_pandas_series_idxmin(self, axis=None, skipna=None):
 
     Limitations
     -----------
-    Parameter ``axis`` is supported only with default value ``None``.
-    Parameter ``skipna`` cannot be ``False`` with data of string type.
+    - Parameter ``axis`` is supported only with default value ``None``.
```
akharche: Let's be more consistent and write either "is supported only with default value" or 'not supported'

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4958,8 +4955,9 @@ def hpat_pandas_series_fillna(self, value=None, method=None, axis=None, inplace=
 
     Limitations
     -----------
-    - Parameters method, limit, downcast are currently unsupported by Intel Scalable Dataframe Compiler
-    - Parameter inplace supported as literal value only
+    - Parameters ``method``, ``limit`` and ``downcast`` are currently unsupported
+    by Intel Scalable Dataframe Compiler.
+    - Parameter ``inplace`` supported as literal value only.
```
akharche:
```suggestion
    - Parameter ``inplace`` is supported with literal value only.
```

Hardcode84:
conflict

AlexanderKalistratov:
Conflict

PR #777, opened 2020-03-25
Excluding tests for non supported functionality from CI

kozlov-alexey:


kozlov-alexey:
Decided to close this and make code-cleanup in a separated PR later.

PR #778, opened 2020-03-25
correct doc for Series.median

Rubtsowa:


sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4581,7 +4581,8 @@ def hpat_pandas_series_median(self, axis=None, skipna=None, level=None, numeric_
 
     Limitations
     -----------
-    - Parameters ``axis``, ``level`` and ``numeric_only`` are currently unsupported by Intel Scalable Dataframe Compiler
+
+    * Parameters ``axis``, ``level`` and ``numeric_only`` are supported supported only with default value ``None``
```
densmirn:
```suggestion
    * Parameters ``axis``, ``level`` and ``numeric_only`` are supported only with default value ``None``.
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4581,7 +4581,8 @@ def hpat_pandas_series_median(self, axis=None, skipna=None, level=None, numeric_
 
     Limitations
     -----------
-    - Parameters ``axis``, ``level`` and ``numeric_only`` are currently unsupported by Intel Scalable Dataframe Compiler
+
+    * Parameters ``axis``, ``level`` and ``numeric_only`` are supported only with default value ``None``.
```
akharche: What's the reason of using "*" instead of "-" as separator?

PR #782, opened 2020-03-25
Groupby limitations

1e-to:


sdc/datatypes/hpat_pandas_groupby_functions.py
```diff
@@ -625,7 +625,13 @@ def sdc_pandas_series_groupby_var(self, ddof=1, *args):
...
     'method_name': 'count',
     'example_caption': 'Compute count of group, excluding missing values.',
-    'limitations_block': '',
+    'limitations_block':
+    """
+    Limitations
+    -----------
+    - This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
+    between staying in JIT-region with that function or going back to interpreter mode.
+    """,
```
Rubtsowa: Put this in a separate variable to avoid code duplication

1e-to: I do not think it is necessary here

Rubtsowa: @kozlov-alexey  what do you think?

1e-to: If we need to add or change limitations, we will have to rewrite the text from the variables back to the lines. To the fact that so clearly

densmirn: I vote for inserting the constant text `This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
    between staying in JIT-region with that function or going back to interpreter mode.` where it requires through string formatting:

```python
performance_limitation = `This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
    between staying in JIT-region with that function or going back to interpreter mode.`
...
'limitations_block':
    """
    Limitations
    -----------
    - {}.
    """.format(performance_limitation),
...
```

densmirn: or
```
performance_limitation = `This function may reveal slower performance than Pandas* on user system. Users should exercise a tradeoff
    between staying in JIT-region with that function or going back to interpreter mode.`
...
'limitations_block':
    f"""
    Limitations
    -----------
    - {performance_limitation}
    """,
...
```

1e-to:
> documentation based on these

image

PR #785, opened 2020-03-26
Update compilation section in docs

AlexanderKalistratov:


docs/source/compilation.rst
```diff
@@ -80,35 +158,34 @@ should also be deterministic. The below example is not supported since the funct
...
+.. code-block::
+
+    names = ['A', 'B']
+    usecols = ['A']
+    dtypes={'A': np.float64}
+    pd.read_csv(file_name, names=names, usecols=usecols, dtype=dtypes)
+
 Alternatively you can take file reading out of the compiled region.
+
+Note: if data file contains integer data with empy positions (Nans) it is highly recomended to manually specify column type to float.
```
PokhodenkoSA:
```suggestion
Note: if data file contains integer data with empty positions (Nans) it is highly recomended to manually specify column type to float.
```

AlexanderKalistratov:
> You mixed reading `DataFrame` from a file and performing `cummax` in `Unsupported APIs` block. Maybe it makes sense to provide unified examples? I mean only reading or only performing a operation.

@densmirn That's a really valid point. I wasn't able to pick up a single function on which a could use both approaches - replace by similar function and use objmode. It is really hard to return DataFrame or Series from objmode

PR #788, opened 2020-03-27
Df.loc impl

1e-to:


sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1944,6 +1994,52 @@ def sdc_pandas_dataframe_iat_impl(self):
...
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+
+    Pandas API: pandas.DataFrame.loc
+
+    Examples
+    --------
+    .. literalinclude:: ../../../examples/dataframe/dataframe_loc.py
+       :language: python
+       :lines: 34-
```
```suggestion
       :lines: 36-
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,63 @@ def _df_getitem_unicode_idx_impl(self, idx):
...
         return None
 
     accessor = self.accessor.literal_value
 
+    if accessor == 'loc':
+        if isinstance(idx, types.Integer):
+            return gen_df_getitem_loc_single_label_impl(self.dataframe, idx)
+
+        raise TypingError('Operator getitem(). The index must be a single label, a  list or array of labels,\
+                          a slice object with labels, a boolean array or a callable. Given: {}'.format(idx))
```
densmirn: Is `getitem()` correct? Shouldn't we use this message info in `limitations` block in docstring?

sdc/datatypes/hpat_pandas_dataframe_functions.py
Rubtsowa: It is better ':lines: 36-'

sdc/tests/test_dataframe.py
```diff
@@ -1100,6 +1100,17 @@ def test_impl(df):
...
+        def test_impl(df):
+            return df.loc[4]
+
+        sdc_func = sdc.jit(test_impl)
+        idx = [3, 4, 1, 2, 0]
+        df = pd.DataFrame({"A": [3.2, 4.4, 7.0, 3.3, 1.0],
+                           "B": [3, 4, 1, 0, 222],
+                           "C": [3.1, 8.4, 7.1, 3.2, 1]}, index=idx)
+        pd.testing.assert_series_equal(sdc_func(df), test_impl(df), check_names=False)
+
```
Rubtsowa: Add test with index not contained in indices DF.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,63 @@ def _df_getitem_unicode_idx_impl(self, idx):
...
+    """
+    Example of generated implementation:
+        def _df_getitem_single_label_loc_impl(self, idx):
+            for i in numba.prange(len(self._dataframe.index)):
+                if self._dataframe._index[i] == idx:
+                data_0 = pandas.Series(self._dataframe._data[0], index=self._dataframe.index)
+                result_0 = data_0.at[idx]
+                data_1 = pandas.Series(self._dataframe._data[1], index=self._dataframe.index)
+                result_1 = data_1.at[idx]
+                return pandas.Series(data=[result_0[0], result_1[0]], index=['A', 'B'], name=str(idx))
```
densmirn: Please fix indentations.

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,63 @@ def _df_getitem_unicode_idx_impl(self, idx):
...
+            raise IndexingError('Index is out of bounds for axis')
+    """
+    func_lines = ['def _df_getitem_single_label_loc_impl(self, idx):',
+                  '  for i in numba.prange(len(self._dataframe.index)):',
+                  '    if self._dataframe._index[i] == idx:']
+    if isinstance(self.index, types.NoneType):
+        func_lines = ['def _df_getitem_single_label_loc_impl(self, idx):',
+                      '  if -1 < idx < len(self._dataframe._data):']
```
densmirn: You will have incorrect indentation if `index` is `None`:
```python
  if -1 < idx < len(self._dataframe._data):  # 2 white spaces
      data_0 =...  # 6 white spaces
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
densmirn: What's about this recommendation?

1e-to: This must be so that the user understands that he entered something incorrect
The same is done in the series.

densmirn: I meant:
1. Actually the operator is not `getitem`, it's `loc`.
2. `The index must be a single label, a  list or array of labels, a slice object with labels, a boolean array or a callable` looks like a limitation, doesn't?

1e-to: How to fix it then?

densmirn:
```suggestion
        ty_checker = TypeChecker('Operator loc().')
        ty_checker.raise_exc(idx, 'int', 'idx')
```

I meant to insert limitations block to docstring as it was done in https://github.com/IntelPython/sdc/commit/0e1ce3a124d428737f7e2e0e98cd739d574dcf0e#diff-37d3d013a811f054d85ea0713b88b1eeR1723-R1731. However `idx` can be only of integer according to the code.

1e-to: I already do that
    - Loc works with basic case only: single label
in limitations

densmirn: You still raise exception with incorrect message.

1e-to: Fixed

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,65 @@ def _df_getitem_unicode_idx_impl(self, idx):
...
+    results = []
+    result_index = []
+    for i, c in enumerate(self.columns):
+        result_c = f"result_{i}"
+        func_lines += [f"{space}    data_{i} = pandas.Series(self._dataframe._data[{i}], index=self._dataframe.index)",
+                       f"{space}    {result_c} = data_{i}.at[idx]"]
+        results.append(result_c)
+        result_index.append(c)
+    data = '[0], '.join(col for col in results) + '[0]'
+    func_lines += [f"{space}    return pandas.Series(data=[{data}], index={result_index}, name=str(idx))",
```
densmirn, line 1900: Better to rename `space` to `indent` something like that:
```suggestion
    indent = ' ' * 6
    if isinstance(self.index, types.NoneType):
        func_lines = ['def _df_getitem_single_label_loc_impl(self, idx):',
                      '  if -1 < idx < len(self._dataframe._data):']
        indent =  ' ' * 4
    results = []
    result_index = []
    for i, c in enumerate(self.columns):
        result_c = f"result_{i}"
        func_lines += [f"{indent}data_{i} = pandas.Series(self._dataframe._data[{i}], index=self._dataframe.index)",
                       f"{indent}{result_c} = data_{i}.at[idx]"]
        results.append(result_c)
        result_index.append(c)
    data = '[0], '.join(col for col in results) + '[0]'
    func_lines += [f"{space}return pandas.Series(data=[{data}], index={result_index}, name=str(idx))",
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1944,6 +1996,57 @@ def sdc_pandas_dataframe_iat_impl(self):
...
+    """
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+
+    Pandas API: pandas.DataFrame.loc
+
+    Limitations
+    -----------
+    - Parameter ``'name'`` in new DataFrame can be String only
+    - Loc works with basic case only: single label
```
densmirn, line 2010:
```suggestion
    - Parameter ``idx`` is supported only to be a single value, e.g. :obj:`df.loc['A']`.
```

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1944,6 +1996,57 @@ def sdc_pandas_dataframe_iat_impl(self):
...
+def sdc_pandas_dataframe_loc(self):
+    """
+    Intel Scalable Dataframe Compiler User Guide
+    ********************************************
+
+    Pandas API: pandas.DataFrame.loc
+
+    Limitations
+    -----------
+    - Parameter ``'name'`` in new DataFrame can be String only
```
densmirn, line 2009: What does the parameter `name` mean?

1e-to: In this case it means that result series (if it series) has name string
Maybe need change limitation to more understanding

densmirn: What is difference between Pandas and SDC in case of `name` of the Series?

densmirn: Doesn't we support numeric name for the Series?

1e-to: Yes, we support only string name series
And if base series containe numeric name, we change it into string

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,93 @@ def _df_getitem_unicode_idx_impl(self, idx):
...
+    Example of generated implementation:
+        def _df_getitem_single_label_loc_impl(self, idx):
+            idx_list = []
+            for i in range(len(self._dataframe.index)):
+                if self._dataframe._index[i] == idx:
+                idx_list.append(i)
+            data_0 = []
+            for i in numba.prange(len(idx_list)):
+                index_in_list_0 = idx_list[i]
+                data_0.append(self._dataframe._data[0][index_in_list_0])
```
AlexanderKalistratov, line 1881: you can't do `append` in `prange` loop. Also you could use `sdc_take`. @kozlov-alexey

sdc/datatypes/hpat_pandas_dataframe_functions.py
```diff
@@ -1867,13 +1867,93 @@ def _df_getitem_unicode_idx_impl(self, idx):
     ty_checker.raise_exc(idx, expected_types, 'idx')
 
 
+def df_getitem_single_label_loc_codegen(self, idx):
+    """
+    Example of generated implementation:
+        def _df_getitem_single_label_loc_impl(self, idx):
+            idx_list = []
+            for i in range(len(self._dataframe.index)):
+                if self._dataframe._index[i] == idx:
```
AlexanderKalistratov, line 1876: What would happen if `_index` is `None`?

Also it is better to do it in parallel. Split it into chunks, create list per chunk and then merge them

AlexanderKalistratov:
@1e-to conflict

sdc/datatypes/hpat_pandas_dataframe_functions.py
1e-to: > What would happen if _index is None?
It is okay because I dont use dataframe._index in case of index = None

sdc/functions/numpy_like.py
```diff
@@ -710,6 +710,45 @@ def dropna_impl(arr, idx, name):
...
+        length = 0
+
+        for i in prange(len(chunks)):
+            chunk = chunks[i]
+            res = 0
+            for j in range(chunk.start, chunk.stop):
+                if arr[j] == idx:
+                    res += 1
+            length += res
+            arr_len[i] = res
```
AlexanderKalistratov, line 733: You could allocate list of list. The length of the first list is equal to number of chunks. In this case you could safely use `append` for list related to chunk. So in this case single loop would be enough. (and another loop (probably, not parallel) to merge all lists into one)

PR #795, opened 2020-04-08
delete some files in pd_series_ext.py

Rubtsowa:


AlexanderKalistratov:
@densmirn so are split tests actually skipped?

densmirn:
> @densmirn so are split tests actually skipped?

Yes, they are. The test I talked about before is on `str.split` (not `Series.str.split`).

AlexanderKalistratov:
Just thought that we need to delete tests too

PR #797, opened 2020-04-10
Disable build of sdc.parquet_cpp

PokhodenkoSA:
This PR disables building `parquet_cpp` module. Tests are already disabled.
Parquet support is not work because HPAT pipeline is not work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -162,8 +162,9 @@ ext_parquet = Extension(name="sdc.parquet_cpp",
 
 _ext_mods = [ext_hdist, ext_chiframes, ext_set, ext_str, ext_dt, ext_io, ext_transport_seq]
 
-if _has_pyarrow:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -162,8 +162,9 @@ def readme():
 
 _ext_mods = [ext_hdist, ext_chiframes, ext_set, ext_str, ext_dt, ext_io, ext_transport_seq]
 
-if _has_pyarrow:
-    _ext_mods.append(ext_parquet)
+# Support of Parquet is disabled because HPAT pipeline is not work now
```
PokhodenkoSA, line 165:
```suggestion
# Support of Parquet is disabled because HPAT pipeline does not work now
```

Vyacheslav-Smirnov:
Suppose we need to completely remove this extension but due to lack of time let's just disable its build.

PR #805, opened 2020-04-15
Prange

1e-to:


kozlov-alexey:
@1e-to Shall we close this one? I think this was resolved by migrating to newer Numba.

PR #814, opened 2020-04-21
VTune integration

1e-to:


Hardcode84:
Native part is not included?

vtune_integration.py
```diff
@@ -0,0 +1,40 @@
...
+ctypes_domain = functype_domain(itt.__itt_domain_create)
+domain = ctypes_domain(b"VTune.Profiling.SDC\0")
+
+functype_string_handle_create = ctypes.CFUNCTYPE(ctypes.c_void_p, ctypes.c_char_p)
+ctypes_string_handle_create = functype_string_handle_create(itt.__itt_string_handle_create)
+
+ll.add_symbol('__itt_task_begin_new', itt.__itt_task_begin)
+ll.add_symbol('__itt_task_end_new', itt.__itt_task_end)
+
+def vtune_profiling_boxing(name_handle):
```
AlexanderKalistratov, line 18: Why do we need this?

vtune_integration.py
```diff
@@ -0,0 +1,40 @@
+import ctypes
```
AlexanderKalistratov, line 1: Let's rename file somehow. Probably, simlpy `vtune.py`

examples/vtune_integration/basic_usage_vtune_profiler.py
```diff
@@ -0,0 +1,27 @@
...
+from numba import njit
+import sdc
+import sdc.vtune_integration as vt
+import ctypes
+import itt
+
+handle = vt.ctypes_string_handle_create(b"Head\0")
+
+functype_task_begin = ctypes.CFUNCTYPE(None, ctypes.c_void_p, ctypes.c_void_p)
+ctypes_task_begin = functype_task_begin(itt.__itt_task_begin)
```
AlexanderKalistratov, line 11: I believe it should be done in `vtune.py`, not here. And here just import these functions

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -2268,6 +2269,7 @@ def hpat_pandas_series_corr_impl(self, other, method='pearson', min_periods=None
 
 
 @sdc_overload_method(SeriesType, 'head')
+@vtune_profiling_overload(name_handle="Series_head")
```
AlexanderKalistratov, line 2272: It should be part of `sdc_overload_method` decorator. We can extract type name and method name from where.

AlexanderKalistratov: You actually could call this decorator inside `sdc_overload_method`

AlexanderKalistratov:
Great progress actually!

PokhodenkoSA (commented):

PR #820, opened 2020-04-28
Adding extension type for pd.RangeIndex

kozlov-alexey:
This commit adds Numba extension types for pandas.RangeIndex class,
allowing creation of pd.RangeIndex objects inside and passing and returning them
to/from nopython functions.


sdc/datatypes/range_index_type.py
```diff
@@ -0,0 +1,68 @@
...
+
+class RangeIndexType(types.IterableType):
+    def __init__(self, is_named=False):
+        self.is_named = is_named
+        super(RangeIndexType, self).__init__(
+            name='RangeIndexType({})'.format(is_named))
+
+    @property
+    def dtype(self):
+        return types.int64
```
densmirn, line 47: Is the property going to be expanded. If isn't maybe replace the property with the attribute `self.dtype = types.int64` to simplify a bit the code.

sdc/tests/test_indexes.py
```diff
@@ -0,0 +1,386 @@
...
+class TestRangeIndex(TestCase):
+
+    def test_range_index_create_and_box(self):
+        def test_impl(start, stop, step, name):
+            return pd.RangeIndex(start, stop, step, name=name)
+        sdc_func = self.jit(test_impl)
+
+        for params in _generate_valid_range_params():
+            for name in test_global_index_names:
+                start, stop, step = params
```
densmirn, line 66: This line could be moved from internal loop to external one. The action could be applied in several places.

sdc/tests/test_indexes.py
```diff
@@ -0,0 +1,386 @@
...
+            '  return pd.RangeIndex({})'
+        ]
+        test_impl_text = '\n'.join(func_lines)
+
+        # use non default values for all parameters except one (tested)
+        non_default_params = {'start': 2, 'stop': 7, 'step': 2, 'name': "'index'"}
+        for arg in non_default_params.keys():
+            with self.subTest(omitted=arg):
+                kwargs = non_default_params.copy()
+                del kwargs[arg]
```
densmirn, line 152: Alternative option:
```python
kwargs = {key: val for key, val in non_default_params.items() if key != arg}
```

sdc/extensions/indexes/range_index_ext.py
```diff
@@ -0,0 +1,366 @@
...
+
+    if not (isinstance(name, (types.NoneType, types.Omitted, types.StringLiteral, types.UnicodeType)) or name is None):
+        ty_checker.raise_exc(name, 'string or none', 'name')
+
+    if ((isinstance(start, (types.NoneType, types.Omitted)) or start is None)
+            and (isinstance(stop, (types.NoneType, types.Omitted)) or stop is None)
+            and (isinstance(step, (types.NoneType, types.Omitted)) or step is None)):
+        def pd_range_index_ctor_dummy_impl(
+                start=None, stop=None, step=None, dtype=None, copy=False, name=None, fastpath=None):
+            raise TypeError("RangeIndex(...) must be called with integers")
```
AlexanderKalistratov, line 117: Why not raise error at compile time?

sdc/extensions/indexes/range_index_ext.py
```diff
@@ -0,0 +1,366 @@
...
+            _start, _stop = types.int64(0), types.int64(start)
+        else:
+            _stop = types.int64(stop)
+
+        _step = types.int64(step) if step is not None else types.int64(1)
+        if _step == 0:
+            raise ValueError("Step must not be zero")
+
+        if _start == 0 and _stop == 0 and _step == 0:
+            raise TypeError("RangeIndex(...) must be called with integers")
```
AlexanderKalistratov, line 140: Is this correct error message? We are checking `_start`, `_stop` and `_step` to be non-zero. But reports what it must be integers

sdc/extensions/indexes/range_index_ext.py
```diff
@@ -0,0 +1,366 @@
...
+    )
+
+    if typ.is_named:
+        name = c.pyapi.from_native_value(types.unicode_type, range_index.name)
+    else:
+        name = c.pyapi.make_none()
+
+    res = c.pyapi.call_method(pd_class_obj, "RangeIndex", (start, stop, step, dtype, copy, name))
+
+    c.pyapi.decref(pd_class_obj)
```
AlexanderKalistratov, line 181: Shouldn't we also decrease reference for `start`, `stop` and `step` objects?
And for `dtype`, `copy` and `name`?

AlexanderKalistratov: Or, creating objects from native doesn't increment reference for it?

sdc/datatypes/range_index_type.py
kozlov-alexey: No, probably it won't be. I changed to use class attribute instead, that should work too.

sdc/tests/test_indexes.py
kozlov-alexey: Yes, thank you, fixed it.

sdc/extensions/indexes/range_index_ext.py
kozlov-alexey: Thanks! I incorrectly merged two different errors into one.

sdc/extensions/indexes/range_index_ext.py
kozlov-alexey: @AlexanderKalistratov  We would have different runtime behavior with pandas then. I'm not sure which way we should take - either comply with pandas as much as possible, or intercept at compile time errors like this that pandas could only catch at runtime, but have compilation failure instead of runtime error. What do you think?

sdc/extensions/indexes/range_index_ext.py
kozlov-alexey: Nice catch! Most likely yes, we need to decref for all python object we created using Python C API functions that return new references. It appears that we don't do this in box/unbox of Series and DFs which I used as a template too, but it's better to address that in a separate PR/task.

PR #821, opened 2020-04-28
Series combine

Rubtsowa:


examples/series/series_combine.py
```diff
@@ -0,0 +1,40 @@
...
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+import numpy as np
+import pandas as pd
+from numba import njit
+
+
+@njit
+def series_copy():
```
densmirn, line 33: Looks the function name is invalid.

examples/series/series_combine.py
```diff
@@ -0,0 +1,40 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+# OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+# OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+import numpy as np
```
densmirn, line 27: The import isn't used at all.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4882,3 +4882,65 @@ def sdc_pandas_series_skew_impl(self, axis=None, skipna=None, level=None, numeri
...
+        ty_checker.raise_exc(fill_value, 'number', 'fill_value')
+
+    def sdc_pandas_series_combine_impl(self, other, func, fill_value=None):
+
+        if fill_value is None:
+            fill_value = numpy.nan
+
+        len_val = max(len(self), len(other))
+        result = numpy.empty(len_val, self._data.dtype)
+        for ind in range(len_val):
```
densmirn, line 4935: Can we parallel the method based on chunks?

1e-to, line 4935: It is case for non-indexes series. Also, it should rewrite with prange

examples/series/series_combine.py
1e-to: Wrong name function

sdc/datatypes/hpat_pandas_series_functions.py
densmirn: + usage of chunks to predict scalability

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4882,3 +4882,65 @@ def sdc_pandas_series_skew_impl(self, axis=None, skipna=None, level=None, numeri
...
+
+    ty_checker.check(other, SeriesType)
+
+    if not isinstance(fill_value, (types.Omitted, types.NoneType, types.Number)) and fill_value is not None:
+        ty_checker.raise_exc(fill_value, 'number', 'fill_value')
+
+    def sdc_pandas_series_combine_impl(self, other, func, fill_value=None):
+
+        if fill_value is None:
+            fill_value = numpy.nan
```
kozlov-alexey, line 4931: This will make fill_value type undefined at compile time. You can probably use the same approach as in operators: https://github.com/IntelPython/sdc/blob/e87095a53f4810e26f944288a51dbe7dbdc2c5c0/sdc/sdc_function_templates.py#L144

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4882,3 +4882,65 @@ def sdc_pandas_series_skew_impl(self, axis=None, skipna=None, level=None, numeri
...
+
+        len_val = max(len(self), len(other))
+        result = numpy.empty(len_val, self._data.dtype)
```
kozlov-alexey, line 4934: This is actually wrong, result dtype should be common dtype for result dtype of func(a, b) where a,b are series values and dtype of _fill_value. Provided tests do not cover this, but e.g. this won't pass:
```
    def test_series_combine_integer_new(self):
        def test_impl(S1, S2):
            return S1.combine(S2, lambda a, b: 2 * a + b, 16.2)

        S1 = pd.Series([1, 2, 3, 4, 5])
        S2 = pd.Series([6, 21, 3, 5])
        result = hpat_func(S1, S2)
        result_ref = test_impl(S1, S2)
        print(f"DEBUG: result:\n{result},\nresult_ref:\n{result_ref}")
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4887,3 +4889,86 @@ def sdc_pandas_series_skew_impl(self, axis=None, skipna=None, level=None, numeri
...
+                    val_self = _fill_value
+                else:
+                    ind_self = self_indexes[j]
+                    val_self = self[ind_self]._data[0]
+
+                if other_indexes[j] == -1:
+                    val_other = _fill_value
+                else:
+                    ind_other = other_indexes[j]
+                    val_other = other[ind_other]._data[0]
```
densmirn, line 4969:
```suggestion
                self_idx = self_indexes[j]
                if self_idx == -1:
                    val_self = _fill_value
                else:
                    val_self = self[self_idx]._data[0]

                other_idx = other_indexes[j]
                if other_idx == -1:
                    val_other = _fill_value
                else:
                    val_other = other[other_idx]._data[0]
```

densmirn: or
```suggestion
                self_idx = self_indexes[j]
                val_self = _fill_value if self_idx == -1 else self[self_idx]._data[0]

                other_idx = other_indexes[j]
                val_other = _fill_value if other_idx == -1 else other[other_idx]._data[0]
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4930,22 +4932,43 @@ def sdc_pandas_series_combine(self, other, func, fill_value=None):
...
+    ret_type = sig.return_type
+
+    fill_dtype = types.float64 if fill_is_default else fill_value
+    res_dtype = find_common_dtype_from_numpy_dtypes([], [ret_type, fill_dtype])
+
     def sdc_pandas_series_combine_impl(self, other, func, fill_value=None):
 
-        if fill_value is None:
-            fill_value = numpy.nan
+        if fill_value is not None:
```
kozlov-alexey, line 4945:
```suggestion
        _fill_value = numpy.nan if fill_value is None else fill_value:
```

sdc/tests/test_series.py
```diff
@@ -2825,7 +2817,7 @@ def test_impl(S1, S2):
         S2 = pd.Series([1, 2, 3, 4, 5])
         pd.testing.assert_series_equal(hpat_func(S1, S2), test_impl(S1, S2))
 
-    @skip_numba_jit
+    @unittest.expectedFailure
```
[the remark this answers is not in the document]
kozlov-alexey: @Rubtsowa No need to skip the test if that's how impl is intended to work. Use check_dtype=False in assert_series_equal and add a comment just before this check to refer to SDC Limitation.

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4930,22 +4932,43 @@ def sdc_pandas_series_combine(self, other, func, fill_value=None):
...
+
+        chunks = parallel_chunks(len_val)
+        for i in prange(len(chunks)):
+            chunk = chunks[i]
+            for j in range(chunk.start, chunk.stop):
+                self_idx = self_indexes[j]
+                if self_idx == -1:
+                    val_self = _fill_value
+                else:
+                    val_self = self[self_idx]._data[0]
```
kozlov-alexey, line 4963: self_idx (and other_idx) is position in the Series, not the index, so instead of using getitem on a Series, that performs index lookup and returns a Series, so that you have to take _data[0] from it, you can just write:
```suggestion
                    val_self = self._data[self_idx]
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4887,3 +4889,77 @@ def sdc_pandas_series_skew_impl(self, axis=None, skipna=None, level=None, numeri
...
+        for i in prange(len(chunks)):
+            chunk = chunks[i]
+            for j in range(chunk.start, chunk.stop):
+                self_idx = self_indexes[j]
+                val_self = _fill_value if self_idx == -1 else self._data[self_idx]
+
+                other_idx = other_indexes[j]
+                val_other = _fill_value if other_idx == -1 else other._data[other_idx]
+
+                result[j] = func(val_self, val_other)
```
kozlov-alexey, line 4962: Why not?
```suggestion
        result = numpy.empty(len_val, res_dtype)
        for i in prange(len(result)):
            self_idx, other_idx = self_indexes[i], other_indexes[i]
            val_self = _fill_value if self_idx == -1 else self._data[self_idx]
            val_other = _fill_value if other_idx == -1 else other._data[other_idx]
            result[i] = func(val_self, val_other)
```

sdc/tests/test_series.py
```diff
@@ -2783,29 +2781,24 @@ def test_impl(S1, S2):
...
     def test_series_combine_assert1(self):
         def test_impl(S1, S2):
             return S1.combine(S2, lambda a, b: 2 * a + b)
         hpat_func = self.jit(test_impl)
 
         S1 = pd.Series([1, 2, 3])
         S2 = pd.Series([6., 21., 3., 5.])
-        with self.assertRaises(AssertionError):
-            hpat_func(S1, S2)
+        pd.testing.assert_series_equal(hpat_func(S1, S2), test_impl(S1, S2))
```
kozlov-alexey, line 2791: General comment for tests: not all combinations of input series dtypes and fill_value are tested e.g. the one I mentioned before - where float fill_value is assigned to otherwise int series. There are no tests with series with non-default indexes (we refer to samelen, but it's not fully correct - series may have same len, but not same indexes), and no tests for checking func impact on result dtype, so it's hard to see from such tests what's really tested and what is not. So the suggestion is to organize tests in a different manner:

1. product of diff series dtypes (default int, int64, float64),
   same series indexes (but not same series sizes),
   fill_value is specified and of different dtypes (None, np.nan, 4, 4.2)
   Covers: test_series_combine_value_samelen
2. 
   product of diff series dtypes (default int, int64, float64),
   same series indexes (but not same series sizes),
   with fill_value is omitted
   Covers: test_series_combine_float3264, test_series_combine_integer_samelen, test_series_combine_samelen, test_series_combine_different_types
3. 
   product of diff series dtypes (default int, int64, float64),
   series indexes that align with and without -1 in indexers
   fill_value is specified and of different dtypes (None, np.nan, 4, 4.2)
   Covers: test_series_combine_integer, test_series_combine_value
4. 
   product of diff series dtypes (default int, int64, float64),
   series indexes that align with and without -1 in indexers 
   fill_value is omitted
   Covers: test_series_combine, test_series_combine_assert1, test_series_combine_assert2, test_series_combine_different_types

New test:
5. (for testing func changes dtype properly)
    product of diff series dtypes (default int, int64, float64),
    same series indexes (but not same series sizes),
    fill_value = 0
    with diff functions (chaning and not chaning res dtype e.g. preserving int domain, e.g. ** and + and not, e.g. /)

kozlov-alexey: For example, test 1 can look like this (it can also be split into two: one when we use check_dtype=False and one when we don't):
```
    def test_series_combine_same_index_fill_value(self):
        def test_impl(S1, S2):
            return S1.combine(S2, lambda a, b: 2 * a + b)

        n = 11
        np.random.seed(0)
        A = np.random.randint(-100, 100, n)
        B = np.arange(n) * 2 + 1
        series_index = 1 + np.arange(n)

        series_dtypes = [None, np.int64, np.float64]
        fill_values = [None, np.nan, 4, 4.2]
        for dtype1, dtype2, fill_value in product(series_dtypes, series_dtypes, fill_values):
            S1 = pd.Series(A, index=series_index, dtype=dtype1)
            S2 = pd.Series(B, index=series_index, dtype=dtype2)
            with self.subTest(S1_dtype=dtype1, S2_dtype=dtype2, fill_value=fill_value):
                result = hpat_func(S1, S2)
                result_ref = test_impl(S1, S2)
                # check_dtype=False due to difference to pandas in some cases
                pd.testing.assert_series_equal(result, result_ref, check_dtype=False)
```

sdc/datatypes/hpat_pandas_series_functions.py
```diff
@@ -4923,8 +4923,12 @@ def sdc_pandas_series_combine(self, other, func, fill_value=None):
...
     Limitations
     -----------
-    - Only supports the case when data in series of the same type.
-    - With the default fill_value parameter value, the type of the resulting series will be float.
+    - Resulting series dtype may be wider than in pandas due to
+      type-stability requirements and depends on fill_value dtype
+      and result of series indexes alignment.
+    - Indixes should be strictly ascending, as inside the function
+      they are sorted in ascending order and the answer becomes
+      different from the result of the pandas.
```
kozlov-alexey, line 4931: @Rubtsowa What? This sounds like a bug...

Rubtsowa: This 'bug' in function **sdc_join_series_indexes**

kozlov-alexey: @Rubtsowa Then it should be fixed (please create a JIRA case with a reproducer) before this is merged. Adapting to a bug is no way. @AlexanderKalistratov correct?

PokhodenkoSA (commented):

PR #826, opened 2020-05-01
PyTest

PokhodenkoSA:


PokhodenkoSA:
Failed experiment. Using PyTest with xdist with 2 threads (HT?) on Azure machines is not fast. I will try to split tests to 10 free jobs in Azure.

PR #833, opened 2020-05-12
Run tests in separate processes

PokhodenkoSA:
I did not find parameter in unittest for excluding tests from run. I know that pytest has exclude parameter. Moving to PyTest is separate task.
So the solution now is rough because it requires to support the list of tests in 2 files. It is just for checking the solution. I will try to find exclude flag for unittest or will try to move to pytest.


Vyacheslav-Smirnov:
If these modules contains all tests that should be executed, then I am ok with this solution for now.
But guess we need to change it in future.

PokhodenkoSA:
It runs all tests as previous just runs it in separate processes.
I will merge it. More sophisticated solutions are welcome.

PR #838, opened 2020-05-14
Update to Numba 0.49.1

PokhodenkoSA:
Also removed mentions of 0.48.0. I hope I have found all places.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -316,7 +316,7 @@ setup(name=SDC_NAME_STR,
           'numpy>=1.16',
           'pandas==0.25.3',
           'pyarrow==0.15.1',
-          'numba>=0.48'
+          'numba >=0.49.0,<0.50.0'
           ],
       cmdclass=sdc_build_commands,
       ext_modules=_ext_mods,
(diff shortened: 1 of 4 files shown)


Vyacheslav-Smirnov:
@PokhodenkoSA , also need to set 0.49.0 in requirements.txt

setup.py
```diff
@@ -316,7 +316,7 @@ def run(self):
           'numpy>=1.16',
           'pandas==0.25.3',
           'pyarrow==0.15.1',
-          'numba>=0.48'
+          'numba >=0.49.0,<0.50.0'
```
AlexanderKalistratov, line 319: Why not 0.49.1?

PokhodenkoSA: For not be binded to 0.49.1 when 0.49.2 will be released.

AlexanderKalistratov: So then probably it would be better `> 0.49.0` ?

PokhodenkoSA: B/c 0.50.0 could break backward compatibility.

AlexanderKalistratov: `< 0.49.0, <0.50.0`

PokhodenkoSA: But 0.49.0 is also workable.

PokhodenkoSA: Numba 0.49.0 has significant regressions, it is better not to use it and start from 0.49.1.

PR #844, opened 2020-05-18
Parallel sort via tbb

AlexanderKalistratov:


AlexanderKalistratov:
Still need to update setup, requirements, conda recipe, whatever to add dependency for tbb-devel

sdc/native/sort.cpp
```diff
@@ -0,0 +1,87 @@
...
+        break;
+    case 4:
+        run_sort(exact_void_range<4>);
+        break;
+    case 8:
+        run_sort(exact_void_range<8>);
+        break;
+    default:
+        // fallback to c qsort?
+        if      (size < 4)    run_sort(_void_range<4>)
```
Hardcode84, line 67: <=

sdc/native/stable_sort.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    inline int copy_size() { return tail - head; }
+
+    inline quant* as_quant(v_type* data) { return reinterpret_cast<quant*>(data); }
+};
+
+template<class T>
+inline void merge_sorted_main_loop(buffer_queue<T>& left, buffer_queue<T>& right, buffer_queue<T>& out, void* compare = nullptr)
+{
+    if (compare)
```
Hardcode84, line 78: You can write this shorter, like
```auto less = compare ? reinterpret_cast<compare_func>(compare) : [](T* r, T* l){ return *r < *l; }```

Comment (AlexanderKalistratov) [reply sdc/native/stable_sort.cpp:78]: calling `comapre` instead of using simple `<` causes 20-30% of performance degradation

File: sdc/native/utils.cpp [hunk 3]
```diff
@@ -26,4 +26,19 @@ void* upper_bound(void* first, void* last, void* value, int size, int item_size,
...
+
+    if (!arena.is_active())
+        arena.initialize();
+
+    return arena;
+}
+
+void set_threads_num(uint64_t threads)
+{
+    get_arena().initialize(threads);
```
Hardcode84, line 41: I'm not sure it is going to work as intended. IMO Better to just save value to global variable here and than pass this var into arena ctor.
`static tbb::task_arena arena(num_threads);`

PR #870, opened 2020-05-30
Merge refactored DataFrame to master

densmirn:


sdc/tests/test_dataframe.py
```diff
@@ -2108,6 +2466,7 @@ def test_impl(n):
         hpat_func = self.jit(test_impl)
         pd.testing.assert_series_equal(hpat_func(n), test_impl(n))
 
+    @unittest.skip('New DataFrame structure: implement unboxing')
```
AlexanderKalistratov, line 2469: !!!

densmirn: Thanks for this catch. Undecorated in https://github.com/IntelPython/sdc/pull/871.

PR #872, opened 2020-06-03
Update README with tbb information

AlexanderKalistratov:
Add information about tbb to README

Should it be merged to master or integration branch?


PokhodenkoSA:
I would like to propose Upstream First Principle. In short: merge to master and then cherry-pick to release branch.
According to GitLab flow: https://docs.gitlab.com/ee/topics/gitlab_flow.html#release-branches-with-gitlab-flow
> Merging into master and then cherry-picking into release is called an “upstream first” policy, which is also practiced by Google and Red Hat.

PR #875, opened 2020-06-03
Add default constructor to void_range::iterator

AlexanderKalistratov:


AlexanderKalistratov:
> Why is this needed?

For msvc compiler. For some reason it is required to use std::stable_sort.

PR #884, opened 2020-06-10
SAT-2966 - SDC: test_series_operator_lt_str fails on Linux 3.6

kozlov-alexey:


kozlov-alexey:
Tested full series suite on python 3.6. and with this fix it passed:
```
Ran 488 tests in 2085.784s
OK (skipped=131, expected failures=14)
```

sdc/tests/test_sdc_numpy.py
```diff
@@ -359,6 +360,104 @@ def sdc_impl(a):
             with self.subTest(data=case):
                 np.testing.assert_array_equal(ref_impl(array0), sdc_func(array1))
 
+    def test_fillna_numeric_inplace_false(self):
```
densmirn, line 363: You could combine tests `test_fillna_numeric_inplace_false` and `test_fillna_numeric_inplace_true` by moving generic code to separate method something like that:
```python
def _test_fillna_numeric_inplace(self, pyfunc, cfunc):
    # generic code
```
Then call the method in the tail of both tests.

kozlov-alexey: @densmirn The assertion code block of these tests are not identical (inplace=True methods modify original arrays), so this new **_test_fillna_numeric** func has to accept inplace arg and alter behavior based on it's value (made it this way).

PR #892, opened 2020-07-08
remove skip_sdc_jit

Rubtsowa:


sdc/tests/test_utils.py
```diff
@@ -175,16 +165,7 @@ def msg_and_func(msg_or_func=None):
...
-    if sdc.config.test_expected_failure:
-        wrapper = unittest.expectedFailure
-    # wrapper = lambda f: f  # disable skipping
-    return wrapper(func) if func else wrapper
-
-
-def skip_sdc_jit(msg_or_func=None):
-    msg, func = msg_and_func(msg_or_func)
-    wrapper = unittest.skipIf(sdc.config.config_pipeline_hpat_default, msg or "sdc pipeline not supported")
+    wrapper = unittest.skipUnless(False, msg or "numba pipeline not supported")
```
kozlov-alexey, line 168:
```suggestion
    wrapper = unittest.skip(msg or "numba pipeline not supported")
```

sdc/tests/test_utils.py
```diff
@@ -94,21 +94,11 @@ def gen_df_int_cols(input_data, with_index=False):
...
-    if sdc.config.config_pipeline_hpat_default:
-        from sdc.distributed import Distribution
-        vals = sdc.distributed.dist_analysis.array_dists.values()
-        return sum([v == Distribution.REP for v in vals])
-    else:
-        return 0
+    return 0
 
 
 def count_parfor_REPs():
```
kozlov-alexey, line 100: These functions (count_array_REPs and count_parfor_REPs) were used in old pipeline only anyway. So if we keep them, it's better remove the other branch (with return 0) and keep impl from sdc.config.config_pipeline_hpat_default == True branch (for reference only).

sdc/config.py
```diff
@@ -71,10 +66,8 @@ def strtobool(val):
 Default value used to select whether sdc functions would inline
 '''
 
-if not config_pipeline_hpat_default:
-    # avoid using MPI transport if no SDC compiler pipeline used
-    config_transport_mpi_default = False
-    config_transport_mpi = config_transport_mpi_default
+config_transport_mpi_default = False
```
kozlov-alexey, line 69: I think these two are not needed now as well and can be removed.

sdc/config.py
```diff
@@ -71,10 +66,8 @@ def strtobool(val):
...
+config_transport_mpi = config_transport_mpi_default
```
kozlov-alexey, line 71: Most likely numba_compiler_define_nopython_pipeline_orig can also be removed.

PR #894, opened 2020-07-16
arguments_annotation_default_func_class

KsanaKozlova:


numba_typing/arguments_annotation_default_func_class.py
```diff
@@ -0,0 +1,24 @@
+from inspect import signature, getfullargspec
+from typing import get_type_hints
+
+
+def get_arguments_annotation_default_func(method):
+    sig = signature(method)
+    argspec = getfullargspec(method)
+    arg_index = 0
```
densmirn, line 8: This variable is not used at all and could be dropped.

numba_typing/arguments_annotation_default_func_class.py
```diff
@@ -0,0 +1,24 @@
+from inspect import signature, getfullargspec
+from typing import get_type_hints
+
+
+def get_arguments_annotation_default_func(method):
```
densmirn, line 5: What do you think about renaming the function to `get_func_annotations` and the argument name to `func`?

numba_typing/arguments_annotation_default_func_class.py
```diff
@@ -0,0 +1,24 @@
...
+    sig = signature(method)
+    argspec = getfullargspec(method)
+    arg_index = 0
+    annotations = {}
+    defaults = {}
+
+    for arg in argspec.args:
```
densmirn, line 12: Looks like you could use signature to iterate through parameters:
`for arg, in sig.parameters.values():`
Also let's fuse both for-loops to a single one.

numba_typing/arguments_annotation_default_func_class.py
```diff
@@ -0,0 +1,24 @@
...
+            raise SyntaxError
+        else:
+            annotations[arg] = sig.parameters[arg].annotation
+    for arg in argspec.args:
+        if (sig.parameters[arg].default != sig.empty):
+            defaults[arg] = sig.parameters[arg].default
+    return annotations, defaults
+
+
+def get_arguments_annotation_class(cls):
```
densmirn, line 23: Let's rename it to something like that:
`get_cls_annotations`

densmirn:
One more think, could you please add [docstrings](https://www.python.org/dev/peps/pep-0257/) for all the functions?

numba_typing/type_annotations.py
```diff
@@ -21,11 +24,20 @@ def get_func_annotations(func):
...
-
-    print(get_func_annotations(foo))
+    annotations = get_type_hints(cls)
+    for x in annotations:
+        if get_args_union(annotations[x]):
+            annotations[x] = get_args_union(annotations[x])
+    return annotations, {}
+
+
+def get_args_union(annot):
```
densmirn, line 34: Maybe make the function to handle any annotation something like that:
```python
def get_annotation_types(annotation):
   """Get types of passed annotation."""
    try:
        if annotation.__origin__ is Union:
            return annotation.__args__
    except AttributeError:
        pass

    return (annotation,)
```

numba_typing/tests/test_type_annotations.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_get_cls_annotations(self):
+        class MyClass(object):
+            x: int = 3
+            y: str = "string"
+
+            def __init__(self, x: str, y: int):
+                self.x = x
+                self.y = y
+
+        self.assertEqual(type_annotations.get_cls_annotations(MyClass), ({'x': int, 'y': str}, {}))
```
densmirn, line 23: Let's avoid long lines everywhere something like that:
```python
result = type_annotations.get_cls_annotations(TestClass)
expected = ({'x': int, 'y': str}, {})
self.assertEqual(result, expected)
```

numba_typing/tests/test_type_annotations.py
```diff
@@ -0,0 +1,46 @@
...
+            pass
+        with self.subTest("annotations and all default values"):
+            self.assertEqual(type_annotations.get_func_annotations(func_two),
+                             ({'a': int, 'b': str, 'c': List[int]}, {'a': 2, 'b': 'string', 'c': [1, 2, 3]}))
+
+        def func_three(a: Dict[int, str], b: str = "string", c: int = 1):
+            pass
+        with self.subTest("annotations and not all default values"):
+            self.assertEqual(type_annotations.get_func_annotations(func_three),
+                             ({'a': Dict[int, str], 'b': str, 'c': int}, {'b': 'string', 'c': 1}))
```
densmirn, line 42: Expected something like that:
```suggestion
    def test_get_func_annotations(self):
        def func_one(a: int, b: Union[int, float], c: str):
            pass

        def func_two(a: int = 2, b: str = "string", c: List[int] = [1, 2, 3]):
            pass

        def func_three(a: Dict[int, str], b: str = "string", c: int = 1):
            pass

        expected_results = {
            func_one: ({'a': int, 'b': [int, float], 'c': str}, {}),
            func_two: ({'a': int, 'b': str, 'c': List[int]}, {'a': 2, 'b': 'string', 'c': [1, 2, 3]}),
            func_three: ({'a': Dict[int, str], 'b': str, 'c': int}, {'b': 'string', 'c': 1}),
        }
        for f, expected in expected_results.items():
            with self.subTest(func=f.__name__):
                self.assertEqual(type_annotations.get_func_annotations(f), expected)
```

numba_typing/type_annotations.py
densmirn: Then it could be expanded if necessary.

numba_typing/tests/test_type_annotations.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Union, Dict, List
+
+
+class TestTypeAnnotations(unittest.TestCase):
+
+    def test_get_func_annotations_exception(self):
+
+        def foo(a: int, b, c: str = "string"):
+            pass
+        self.assertRaises(SyntaxError,  type_annotations.get_func_annotations, foo)
```
densmirn, line 12: Let's use context manager and get exception message to assert:
```suggestion
        with self.assertRaises(SyntaxError) as raises:
            type_annotations.get_func_annotations(foo)
        self.assertIn('Not found annotation for parameter b', str(raises.exception))
```

numba_typing/type_annotations.py
AlexanderKalistratov: The only comment is I'd prefer this function to always return list. Even if it would contain single item

numba_typing/tests/test_type_annotations.py
```diff
@@ -0,0 +1,46 @@
...
+                self.y = y
+
+        self.assertEqual(type_annotations.get_cls_annotations(MyClass), ({'x': int, 'y': str}, {}))
+
+    def test_get_func_annotations(self):
+        def func_one(a: int, b: Union[int, float], c: str):
+            pass
+        with self.subTest("annotations"):
+            self.assertEqual(type_annotations.get_func_annotations(func_one),
+                             ({'a': int, 'b': [int, float], 'c': str}, {}))
```
AlexanderKalistratov, line 30: It would be better if annotation would always contain list as value:
```python
{'a': [int], 'b': [int, float], 'c': [str]}
```

numba_typing/tests/test_type_annotations.py
```diff
@@ -0,0 +1,46 @@
...
+        def foo(a: int, b, c: str = "string"):
+            pass
+        self.assertRaises(SyntaxError,  type_annotations.get_func_annotations, foo)
+
+    def test_get_cls_annotations(self):
+        class MyClass(object):
+            x: int = 3
+            y: str = "string"
+
+            def __init__(self, x: str, y: int):
```
AlexanderKalistratov, line 19: That's a very strange test. Does it makes intentionally what annotation of class members differs from annotation in __init__?

That's an interesting case, but in this case it causes only confusion

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_func_annotations(func):
+    """Get annotations and default values of the fuction parameters."""
+    sig = signature(func)
+    annotations = {}
+    defaults = {}
+
+    for name, param in sig.parameters.items():
+        if param.annotation == sig.empty:
+            raise SyntaxError(f'Not found annotation for parameter {name}')
```
AlexanderKalistratov, line 13:
```python
f'No annotation for parameter {name}'
```

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,89 @@
...
+    try:
+        if annotation.__origin__ is Union:
+            return list(annotation.__args__)
+    except AttributeError:
+        pass
+
+    return [annotation, ]
+
+
+def expend_annotations(info):
```
densmirn, line 43: Let's rename the function to for example `product_annotations` and `info` -> `annotations`. Then you could apply the similar changes in the test.

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,89 @@
...
+        pass
+
+    return [annotation, ]
+
+
+def product_annotations(annotations):
+    """Get all variants of annotations."""
+    annot = annotations[0]
+    vals = annotations[1]
+    list_of_annot = list(product(*annot.values()))
```
densmirn, line 47:
```suggestion
    types_product = list(product(*annot.values()))
```
?

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,89 @@
...
+    tvs = {}
```
densmirn, line 48: You could consider defaultdict for this case:
```suggestion
    typevars = defaultdict(list)
```
Maybe this variable is not needed at all.

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,89 @@
...
+    tvs_unique = {}
+    count = 1
+    for x in annot:
+        for y in annot[x]:
+            if isinstance(y, TypeVar) and y.__constraints__ != ():
+                if x in tvs:
+                    tvs[x].append(y)
+                else:
+                    tvs[x] = [y, ]
+                if y not in tvs_unique.keys():
```
densmirn, line 58:
```suggestion
                if typ not in typevars_unique:
```

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,89 @@
...
+                    tvs_unique[y] = y.__constraints__
+                    count *= len(y.__constraints__)
+
+    prod = list(product(*tvs_unique.values()))
+    temp_res = []
+
+    for i in range(len(list_of_annot)):
```
densmirn, line 65:
```suggestion
    for typs in types_product:
         ...
         typs[num]
         ...
```

numba_typing/type_annotations.py
```diff
@@ -112,7 +122,7 @@ def update_sig(temp_sig, typevar):
     for constr_type in typevar.__constraints__:
         sig = {}
         for name, typ in temp_sig.items():
-            if True in exsist_typevar(typ, typevar):
+            if True in exist_typevar(typ, typevar):
```
AlexanderKalistratov, line 125: Can't it be:
```python
if typevar in get_typevars(typ):
```
?

numba_typing/type_annotations.py
```diff
@@ -40,17 +40,27 @@ def get_annotation_types(annotation):
...
     #unique_typevars = get_internal_typevars(list_of_sig)
 
     for sig in list_of_sig:
-        result_product.extend(get_internal_typevars(sig))
+        signature.extend(get_internal_typevars(sig))
+    
+    return add_vals_to_signature(signature, vals)
 
-    return result_product
+def add_vals_to_signature(signature, vals):
```
AlexanderKalistratov, line 55: you could use list comprehension instead:
```python
[[sig, vals] for sig in signature]
```

numba_typing/type_annotations.py
```diff
@@ -40,17 +40,27 @@ def get_annotation_types(annotation):
...
+    '''Add default values ​​to all signatures'''
+    result = []
```
AlexanderKalistratov, line 57: I think it would be better to return namedtuple (parameters, defaults) instead

AlexanderKalistratov:
@densmirn could you please take a look?

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,150 @@
...
+
+    for sig in list_of_sig:
+        signature.extend(get_internal_typevars(sig))
+
+    return add_vals_to_signature(signature, vals)
+
+
+def add_vals_to_signature(signature, vals):
+    '''Add default values ​​to all signatures'''
+    Annotations = namedtuple('Annotations', ['parameters', 'defaults'])
```
AlexanderKalistratov, line 58: let's name it ```Signature``` and declare it outside of function. And then import it in tests too

numba_typing/type_annotations.py
```diff
@@ -0,0 +1,149 @@
...
+    if typ == typevar:
+        return (final_typ)
+    elif isinstance(typ, _GenericAlias):
+        result = list()
+        for arg in typ.__args__:
+            result.append(replace_typevar(arg, typevar, final_typ))
+        result_type = typ.copy_with(tuple(result))
+        return (result_type)
+
+    return (typ)
```
densmirn, line 149:
```suggestion
    if typ == typevar:
        return final_typ
    elif isinstance(typ, _GenericAlias):
        result = list()
        for arg in typ.__args__:
            result.append(replace_typevar(arg, typevar, final_typ))
        result_type = typ.copy_with(tuple(result))
        return result_type

    return typ
```

AlexanderKalistratov:
@KsanaKozlova please fix pep issues

PR #895, opened 2020-07-16
Getting local variable annotations of function and generics

vlad-perevezentsev:
Getting local variable annotations of function and generics


numba_typing/ast_example.py
```diff
@@ -0,0 +1,45 @@
...
+        tree = ast.parse(f.read())
+    analyzer = Analyzer()
+    analyzer.visit(tree)
+    print(analyzer.types)
+
+
+class Analyzer(ast.NodeVisitor):
+    def __init__(self):
+        self.types = {}
+        self.flag = False
```
Hardcode84, line 17: more descriptive name instead of just 'flag'?

numba_typing/ast_example.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def visit_AnnAssign(self, node):
+        if self.function_flag:
+            # need to check for child functions
+            if isinstance(node.annotation, ast.Subscript):  # containers and generics
+                container_name = node.annotation.value.id
+                if isinstance(node.annotation.slice.value, ast.Tuple):
+                    type_list = []
+                    for t in node.annotation.slice.value.elts:
+                        type_list.append(t.id)
```
densmirn, line 40: It could be done in other way `types  = [t.id for t in node.annotation.slice.value.elts]`

numba_typing/ast_example.py
```diff
@@ -0,0 +1,54 @@
...
+    def visit_FunctionDef(self, obj):
+        self.function_flag = True
+        for i in obj.body:
+            self.visit(i)
+
+    def visit_AnnAssign(self, node):
+        if self.function_flag:
+            # need to check for child functions
+            if isinstance(node.annotation, ast.Subscript):  # containers and generics
+                container_name = node.annotation.value.id
```
densmirn, line 36: Maybe to create separate short variables for `node.annotation` and `node.target` because they're used many times. E.g. `annotation, target = node.annotation, node.target`.

numba_typing/ast_example.py
```diff
@@ -0,0 +1,54 @@
...
+                if isinstance(node.annotation.slice.value, ast.Tuple):
+                    type_list = []
+                    for t in node.annotation.slice.value.elts:
+                        type_list.append(t.id)
+                    list_to_str = ','.join(type_list)
+                    self.types[node.target.id] = [eval(f'{self.module}.{container_name}[{list_to_str}]')]
```
densmirn, line 42: `f'{self.module}.{container_name}` could be moved to a separate variable, e.g. `full_container_name`.

numba_typing/ast_example.py
```diff
@@ -0,0 +1,54 @@
...
+    def visit_AnnAssign(self, node):
+        if self.function_flag:
+            # need to check for child functions
+            if isinstance(node.annotation, ast.Subscript):  # containers and generics
+                container_name = node.annotation.value.id
+                if isinstance(node.annotation.slice.value, ast.Tuple):
+                    type_list = []
+                    for t in node.annotation.slice.value.elts:
+                        type_list.append(t.id)
+                    list_to_str = ','.join(type_list)
```
densmirn, line 41: Perhaps it doesn't make sense to create separate variable for that and use the expression directly in f-string.

numba_typing/test_generics.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def qwe():
+    qwe_int: int = 10
+    qwe_int_op_n: Optional[int]
+    qwe_any: Any
+    qwe_l: List[int] = [1, 2, 3]
+    qwe_list: Optional[int] = [1, 2, 3]
+    qwe_list_u: Union[int, float]
+    qwe_tuple: Union[str, int] = ("qqq", 1)
```
AlexanderKalistratov, line 24: It's not Union. It's tuple.

numba_typing/test_generics.py
```diff
@@ -0,0 +1,25 @@
...
+    qwe_dict: Union[str, str] = {'hello': 'world'}
```
AlexanderKalistratov, line 25: It's not Union, it's dictionary

numba_typing/ast_example.py
```diff
@@ -0,0 +1,54 @@
...
+                    self.types[node.target.id] = [eval(f'{self.module}.{container_name}[{list_to_str}]')]
+                else:
+                    self.types[node.target.id] = [
+                        eval(f'{self.module}.{container_name}[{node.annotation.slice.value.id}]')]
+            else:
+                try:  # not containers
+                    self.types[node.target.id] = [eval(f'{node.annotation.id}')]
+                except NameError:
+                    self.types[node.target.id] = [eval(f'{self.module}.{node.annotation.id}')]  # if Any type
```
densmirn, line 50: You could assign `self.types[node.target.id]` once. To do that, you should assign all `[eval(...)]` to some variable, e.g. `res_types`.

densmirn:
One more think, could you please add docstrings for all the functions/methods?

numba_typing/ast_example.py
```diff
@@ -0,0 +1,55 @@
...
+    def visit_AnnAssign(self, node):
+        if not isinstance(node.parent, ast.FunctionDef):
+            return
+        target, annotation = node.target, node.annotation
+        # need to check for child functions
+        if isinstance(annotation, ast.Subscript):  # containers and generics
+            container_name = annotation.value.id
+            if isinstance(annotation.slice.value, ast.Tuple):
+                type_list_in_str = ','.join([elt.id for elt in annotation.slice.value.elts])
+                exec(f'{target.id} = {container_name}[{type_list_in_str}]', None, self.locals_parameter)
```
densmirn, line 46: Maybe we will prepare command to execute in each if-else branch, then call `exec` after all the branches with the prepared command:
```python
if ...:
    if ...:
        cmd = ...
    else ...:
        cmd = ...
else:
    cmd = ...
exec(cmd, None, self.locals_parameter)
```

numba_typing/ast_example.py
```diff
@@ -0,0 +1,55 @@
...
+class Analyzer(ast.NodeVisitor):
+    def __init__(self):
+        self.locals_parameter = {}
+
+    def visit_ImportFrom(self, node):
+        for alias in node.names:
+            exec(f'from {node.module} import {alias.name}', None, self.locals_parameter)
+
+    def visit_Import(self, node):
+        exec(f'import {node.names[0].name}', None, self.locals_parameter)
```
densmirn, line 35: What does mean `node.names[0]`? Why not for example node.names[1]?
Perhaps better to create some variable with the understandable name for that.

numba_typing/ast_example.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def visit_AnnAssign(self, node):
+        if not isinstance(node.parent, ast.FunctionDef):
+            return
+        target, annotation = node.target, node.annotation
+        # need to check for child functions
+        if isinstance(annotation, ast.Subscript):  # containers and generics
+            container_name = annotation.value.id
+            if isinstance(annotation.slice.value, ast.Tuple):
+                type_list_in_str = ','.join([elt.id for elt in annotation.slice.value.elts])
```
densmirn, line 45: 2 remarks:
```suggestion
                types_as_str = ','.join(elt.id for elt in annotation.slice.value.elts)
```
?

numba_typing/ast_example.py
```diff
@@ -0,0 +1,55 @@
...
+        module_tree = ast.parse(f.read())
+    assign_parent(module_tree)
+    assign_parent(func_tree)
+    analyzer = Analyzer()
+    analyzer.visit(module_tree)
+    analyzer.visit(func_tree)
+    return analyzer.locals_parameter
+
+
+def assign_parent(tree):
```
densmirn, line 20:
```suggestion
def assign_parents(tree):
```

densmirn:
> One more think, could you please add docstrings for all the functions/methods?

Again

numba_typing/local_variable_type_annotations.py
```diff
@@ -29,27 +30,33 @@ def visit_AnnAssign(self, node):
...
-                exec(f'{target.id} = {annotation.id}', self.global_parameter, self.locals_parameter)
-            except NameError:  # if Any type
-                exec_variables = f'{target.id} = {self.module_name}.{annotation.id}'
+                exec(f'{target.id} = [{annotation.id}]', self.global_parameter, self.locals_parameter)
+            except NameError:  # if TypeVar
+                exec_variables = f'{target.id} = [{self.module_name}.{annotation.id}]'
                 exec(exec_variables, self.global_parameter, self.locals_parameter)
 
 
 if __name__ == '__main__':
```
AlexanderKalistratov, line 61: Do you still need this?

numba_typing/local_variable_type_annotations.py
```diff
@@ -29,27 +30,33 @@ def visit_AnnAssign(self, node):
...
                 container_name = annotation.value.id
-                if isinstance(annotation.slice.value, ast.Tuple):
-                    types_as_str = ','.join(elt.id for elt in annotation.slice.value.elts)
-                    exec_variables = f'{target.id} = {self.module_name}.{container_name}[{types_as_str}]'
-                else:
-                    exec_variables = f'{target.id} = {self.module_name}.{container_name}[{annotation.slice.value.id}]'
+                module_import_name = ''
             except AttributeError:  # typing.
-                module_import_name = annotation.value.value.id
+                module_import_name = annotation.value.value.id + '.'
```
AlexanderKalistratov, line 35: We could and should avoid this approach. We can extract all global and non-local variables from function object itself instead of extracting objects from module.

Function object contains list of globals it captured in `__globals__` member.
Also you should use `inspect.getclosurevars` to get captured non-locals for the following case:
```python
def foo():
    T = TypeVar('T')
    def bar():
        a: T
    ...

    return bar


my_func = foo()
```

So, instead of trying to get objects from module, we should use function global and closure vars, unite it in single dictionary and pass it to `exec` as globals.

numba_typing/local_variable_type_annotations.py
```diff
@@ -1,5 +1,6 @@
 import ast
 import inspect
+import test_generics
```
AlexanderKalistratov, line 3: Why?

numba_typing/test.py
```diff
@@ -0,0 +1,64 @@
...
+            t_set: set
+        result = ast_example.get_variable_annotations(test_func)
+        expected_result = {'t_int': int, 't_str': str, 't_float': float, 't_bool': bool, 't_bytes': bytes,
+                           't_list': list, 't_dict': dict, 't_tuple': tuple, 't_set': set}
+        self.assertEqual(result, expected_result)
+
+    def test_get_variable_annotations_generic_types(self):
+
+        def test_func():
+            # t_any: Any
```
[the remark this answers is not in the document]
vlad-perevezentsev: I forgot to delete files. Fixed

numba_typing/test_variable_annotations.py
```diff
@@ -0,0 +1,90 @@
...
+        expected_result = {'t_class_a': [A], 't_class_b': [B], 't_class_c': [C]}
+        self.assertEqual(result, expected_result)
+
+    def test_get_variable_annotations_non_locals(self):
+        def foo():
+            Q = TypeVar('Q')
+
+            def bar():
+                t_typevar: Q
+            ...
```
AlexanderKalistratov, line 81: why do you need it here?

AlexanderKalistratov:
@densmirn could you please take a look?

PR #897, opened 2020-07-22
Added e2e benchmark

akharche:
This PR:
- adds folder for e2e SDC examples and benchmarks
- adds exchange_benchmark.py example under Appache licence


benchmarks/exchange_benchmark.py
```diff
@@ -0,0 +1,94 @@
...
+
+def main():
+    print("Run Pandas...")
+    t_start = time.time()
+    process_data()
+    print("TOTAL Pandas time: ", time.time() - t_start)
+
+    f = open(os.devnull, 'w')
+    t_start = time.time()
+    with redirect_stdout(f):
```
densmirn, line 82: Maybe it makes sense to measure only body of the context manager.
```python
with redirect_stdout(f):
    t_start = time.time()
    sdc_process_data()  # Warming up
    t_end = time.time()
```

benchmarks/exchange_benchmark.py
```diff
@@ -0,0 +1,94 @@
...
+        sdc_process_data()  # Warming up
+    print("SDC WARM_UP time: ", time.time() - t_start)
+
+    print("Run SDC...")
+    t_start = time.time()
+    df1, res, end_time = sdc_process_data()
+    print("NO boxing SDC time: ", end_time)
+    print("TOTAL SDC time: ", time.time() - t_start)
```
densmirn, line 90: We get end time after the first printing, let's do it before the printing.

PR #900, opened 2020-07-28
fix rolling count

Rubtsowa:


sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -730,7 +730,62 @@ def impl(self, ddof=1):
...
+
+            for idx in range(interlude_stop, chunk.stop):
+                put_value = input_arr[idx]
+                pop_value = input_arr[idx - win]
+                nfinite, result = put(put_value, nfinite, result)
+                nfinite, result = pop(pop_value, nfinite, result)
+                output_arr[idx] = get_result(nfinite, minp, result)
+
+            if minp > 1:
+                output_arr[:minp-1] = numpy.nan
```
densmirn, line 780: Could we somehow apply the change in put/pop/get_result of `count` to not copy-paste the most of the existing code?

sdc/datatypes/hpat_pandas_series_rolling_functions.py
```diff
@@ -625,6 +625,9 @@ def impl(self):
                 nfinite, result = pop(pop_value, nfinite, result)
                 output_arr[idx] = get_result(nfinite, minp, result)
 
+        if minp > 1:
+            output_arr[:minp-1] = numpy.nan
```
densmirn, line 629: Doesn't it break other tests on rolling?

Rubtsowa: @densmirn  I run all tests, all OK.

PR #904, opened 2020-07-30
Fix RangeIndex tests

akharche:
Pandas 1.0 supports int32 types in RangeIndex but SDC not:
- Remove int32 from invalid types
- Change Exception type and message


akharche:
@PokhodenkoSA It's ok after restarting

PR #911, opened 2020-08-17
automatic generation of type checks for overload

vlad-perevezentsev:


numba_typing/overload_list.py
```diff
@@ -0,0 +1,183 @@
...
+    def match_generic(self, p_type, n_type):
+        res = True
+        for arg in p_type.__args__:
+            res = res and self.match(arg, n_type)
+        return res
+
+
+def choose_func_by_sig(sig_list, values_dict, defaults_dict={}):
+    checker = TypeChecker()
+    for sig in sig_list:  # sig = (Signature,func)
```
Hardcode84, line 158: Instead of accessing `sig[0]` and `sig[1]` each time you can write `for sig, func in sig_list:`

numba_typing/overload_list.py
```diff
@@ -0,0 +1,183 @@
...
+from numba import types
+from numba import typeof
+from numba.extending import overload
+from type_annotations import product_annotations, get_func_annotations
+from numba import njit
+import typing
+from numba import NumbaDeprecationWarning, NumbaPendingDeprecationWarning
+import warnings
+from numba.typed import List, Dict
+from inspect import getfullargspec
```
Hardcode84, line 12: There are some unused imports here, like `numpy`, `typeof` and `njit`, please clean them up.

numba_typing/overload_list.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+warnings.simplefilter('ignore', category=NumbaDeprecationWarning)
+warnings.simplefilter('ignore', category=NumbaPendingDeprecationWarning)
```
Hardcode84, line 16: Why is this needed?

numba_typing/overload_list.py
```diff
@@ -0,0 +1,180 @@
...
+    res = isinstance(n_type, types.List) or isinstance(n_type, types.ListType)
+    if isinstance(p_type, type):
+        return res
+    else:
+        return res and self.match(p_type.__args__[0], n_type.dtype)
+
+
+def check_tuple_type(self, p_type, n_type):
+    res = False
+    if isinstance(n_type, types.Tuple):
```
AlexanderKalistratov, line 64: Instead of using `n_type.key` you could use `n_type.types` which is defined for both `Tuple` and `UniTuple`, so you can unite both branches:

Something like:
```python
if isinstance(n_type, types.Tuple, types.UniTuple):
    res = True
    for p_val, n_val in zip(p_type.__args__, n_type.types):
        res = res and self.match(p_val, n_val)

return False
```

And btw you need to check that size of `p_type.__args__` and `n_type.types` are the same.
And I believe you don't have tests for the case when they are different.

numba_typing/overload_list.py
```diff
@@ -0,0 +1,180 @@
...
+            elif self._is_generic(p_type):
+                origin_type = self._get_origin(p_type)
+                if origin_type == typing.Generic:
+                    return self.match_generic(p_type, n_type)
+                else:
+                    return self._types_dict[origin_type](self, p_type, n_type)
+            elif isinstance(p_type, typing.TypeVar):
+                return self.match_typevar(p_type, n_type)
+            else:
+                if p_type in (list, tuple):
```
AlexanderKalistratov, line 133: Shouldn't dict be here too?

numba_typing/overload_list.py
```diff
@@ -0,0 +1,180 @@
...
+        if isinstance(p_type, type):
+            return res
+        for p_val, n_val in zip(p_type.__args__, n_type.keyvalue_type):
+            res = res and self.match(p_val, n_val)
+    return res
+
+
+class TypeChecker:
+
+    _types_dict = {int: check_int_type, float: check_float_type, bool: check_bool_type,
```
AlexanderKalistratov, line 92: I'd prefer this checks to be added using `add_type_check` function

AlexanderKalistratov:
https://github.com/IntelPython/sdc/blob/numba_typing/numba_typing/local_variable_type_annotations.py#L11

This code actually modifies original function globals. This shouldn't happen, please fix.

numba_typing/overload_list.py
```diff
@@ -0,0 +1,180 @@
...
+TypeChecker.add_type_check(int, check_int_type)
+TypeChecker.add_type_check(float, check_float_type)
+TypeChecker.add_type_check(str, check_str_type)
+TypeChecker.add_type_check(bool, check_bool_type)
+TypeChecker.add_type_check(list, check_list_type)
+TypeChecker.add_type_check(tuple, check_tuple_type)
+TypeChecker.add_type_check(dict, check_dict_type)
+
+
+def choose_func_by_sig(sig_list, values_dict, defaults_dict):
```
AlexanderKalistratov, line 158: Let's do it this way:

```python
def choose_func_by_sig(sig_list, values_dict, defaults_dict):
    def check_signature(sig_params, types_dict):
        checker = TypeChecker()
        for name, typ in types_dict.items():  # name,type = 'a',int64
            if isinstance(typ, types.Literal):
                typ = typ.literal_type

            if not checker.match(sig_params[name], typ):
                return False
            
            return True

    for sig, func in sig_list:  # sig = (Signature,func)
        for param in sig.parameters:  # param = {'a':int,'b':int}
            if check_signature(param, values_dict):
                return func

    return None
```

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+D['qwe'] = 1
+D['qaz'] = 2
+D_1[1] = True
+D_1[0] = False
+list_type = types.ListType(types.int64)
+D_list = Dict.empty(key_type=types.unicode_type, value_type=list_type)
+D_list['qwe'] = List([3, 4, 5])
+str_1 = 'qwe'
+str_2 = 'qaz'
+test_cases = [('int', [1, 2], {'a': int, 'b': int}), ('float', [1.0, 2.0], {'a': float, 'b': float}),
```
AlexanderKalistratov, line 110: `{'a': int, 'b': int}` why do you need two parameters of the same type?
What are you actually testing?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('bool', [True, True], {'a': bool, 'b': bool}), ('str', ['str_1', 'str_2'], {'a': str, 'b': str}),
+              ('list', [[1, 2], [3, 4]], {'a': typing.List[int], 'b':list}),
```
AlexanderKalistratov, line 112: Why do you need second list parameter?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('List_typed', [L, [3, 4]], {'a': typing.List[int], 'b':list}),
+              ('tuple', [(1, 2.0), ('3', False)], {'a': typing.Tuple[int, float], 'b':tuple}),
+              ('dict', ['D', 'D_1'], {'a': typing.Dict[str, int], 'b': typing.Dict[int, bool]}),
+              ('union_1', [1, False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
+              ('union_2', ['str_1', False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
```
AlexanderKalistratov, line 117: And why you are not testing second Union parameter?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('nested_list', ['L_int', 'L_float'], {'a': typing.List[typing.List[int]],
+                                                     'b': typing.List[typing.List[typing.List[float]]]}),
+              ('TypeVar_TT', ['L_f', [3.0, 4.0]], {'a': 'T', 'b': 'T'}),
+              ('TypeVar_TK', [1.0, 2], {'a': 'T', 'b': 'K'}),
+              ('TypeVar_ListT_T', ['L', 5], {'a': 'typing.List[T]', 'b': 'T'}),
+              ('TypeVar_ListT_DictKT', ['L', 'D'], {'a': 'typing.List[T]', 'b': 'typing.Dict[K, T]'}),
+              ('TypeVar_ListT_DictK_ListT', ['L', 'D_list'], {'a': 'typing.List[T]',
+                                                              'b': 'typing.Dict[K, typing.List[T]]'})]
+
+test_cases_default = [('int_defaults', [1], {'a': int}, {'b': 0}), ('float_defaults', [1.0], {'a': float}, {'b': 0.0}),
```
AlexanderKalistratov, line 127: And what about both - annotation and default value?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('union_1', [1, False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
+              ('union_2', ['str_1', False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
+              ('nested_list', ['L_int', 'L_float'], {'a': typing.List[typing.List[int]],
+                                                     'b': typing.List[typing.List[typing.List[float]]]}),
+              ('TypeVar_TT', ['L_f', [3.0, 4.0]], {'a': 'T', 'b': 'T'}),
+              ('TypeVar_TK', [1.0, 2], {'a': 'T', 'b': 'K'}),
+              ('TypeVar_ListT_T', ['L', 5], {'a': 'typing.List[T]', 'b': 'T'}),
+              ('TypeVar_ListT_DictKT', ['L', 'D'], {'a': 'typing.List[T]', 'b': 'typing.Dict[K, T]'}),
+              ('TypeVar_ListT_DictK_ListT', ['L', 'D_list'], {'a': 'typing.List[T]',
+                                                              'b': 'typing.Dict[K, typing.List[T]]'})]
```
AlexanderKalistratov, line 125: Are where any tests for TypeVar with specified types restriction?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('bool', [True, True], {'a': bool, 'b': bool}), ('str', ['str_1', 'str_2'], {'a': str, 'b': str}),
+              ('list', [[1, 2], [3, 4]], {'a': typing.List[int], 'b':list}),
+              ('List_typed', [L, [3, 4]], {'a': typing.List[int], 'b':list}),
+              ('tuple', [(1, 2.0), ('3', False)], {'a': typing.Tuple[int, float], 'b':tuple}),
+              ('dict', ['D', 'D_1'], {'a': typing.Dict[str, int], 'b': typing.Dict[int, bool]}),
+              ('union_1', [1, False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
+              ('union_2', ['str_1', False], {'a': typing.Union[int, str], 'b': typing.Union[float, bool]}),
+              ('nested_list', ['L_int', 'L_float'], {'a': typing.List[typing.List[int]],
+                                                     'b': typing.List[typing.List[typing.List[float]]]}),
+              ('TypeVar_TT', ['L_f', [3.0, 4.0]], {'a': 'T', 'b': 'T'}),
```
AlexanderKalistratov, line 120: Any negative tests for such case?

numba_typing/test_overload_list.py
```diff
@@ -0,0 +1,146 @@
...
+              ('TypeVar_ListT_DictK_ListT', ['L', 'D_list'], {'a': 'typing.List[T]',
+                                                              'b': 'typing.Dict[K, typing.List[T]]'})]
+
+test_cases_default = [('int_defaults', [1], {'a': int}, {'b': 0}), ('float_defaults', [1.0], {'a': float}, {'b': 0.0}),
+                      ('bool_defaults', [True], {'a': bool}, {'b': False}),
+                      ('str_defaults', ['str_1'], {'a': str}, {'b': '0'}),
+                      ('tuple_defaults', [(1, 2)], {'a': tuple}, {'b': (0, 0)})]
+
+
+for name, val, annotation in test_cases:
```
AlexanderKalistratov, line 133: I'd prefer you to group all cases into 4-5 tests with subtests. Something like this:

```python
def test_common_types():
    test_cases = [({'a': 0}, {'a': int}),
                 ({'a': 0.}, {'a': float}),
                ...]
    
    for case in test_cases:
        with self.Subtest(case=case):
            run_test(case)
```

numba_typing/test_overload_list.py
vlad-perevezentsev: The second parameter is used to pass arguments to the function

PR #919, opened 2020-08-31
Add addrspace to categorical

AlexanderKalistratov:


kozlov-alexey (approved):

tomashek:
Is this change needed for OsX, and warrants RC2? We already dropped RC1.

PR #920, opened 2020-09-02
add new python typing

KsanaKozlova:


AlexanderKalistratov:
Lets move it to folder 'numba_typing/types' and split into several files: numbers, array, pandas(series) and special(LiteralType)

AlexanderKalistratov:
Then import everything in __init__.py.

And please take a look how to use __all__

PR #923, opened 2020-09-07
Fix for supporting DF categoricals unboxing

kozlov-alexey:


densmirn (approved):

sdc/hiframes/api.py
```diff
@@ -162,7 +163,7 @@ def fix_df_array_list_str_impl(column):  # pragma: no cover
     if isinstance(column, RangeIndexType):
         return lambda column: np.array(column)
 
-    if isinstance(column, (types.Array, StringArrayType)):
+    if isinstance(column, (types.Array, StringArrayType, Categorical)):
```
AlexanderKalistratov, line 166: Are there other methods which need the same fix?

sdc/tests/categorical/test_df_category.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_boxing(self):
+        @nb.njit
+        def func(c):
+            return c
+
+        pd_value = self._pd_value()
+        boxed = func(pd_value)
+        assert(boxed.equals(pd_value))
+
+    @skip_numba_jit("capturing DFs (not only categoricals) as freevar not working")
```
kozlov-alexey, line 78: OK, here's what's going on: this doesn't test lowering, but test possibility of capture as freevar/const and it currently works only for categorical series since impl for them was added via @lower_const in https://github.com/IntelPython/sdc/blob/8c0914ac0ec853c5507d8c3f75d1250ef37e265c/sdc/datatypes/series/boxing.py#L33.

sdc/hiframes/api.py
kozlov-alexey: I think half of all DF and Series methods should have some special handling of Categoricals (see e.g. df.values). If we talk only about boxing/unboxing/fix_df_array/fix_df_index function, then probably no, since we don't support CategoricalIndexes yet. At least it looks so.

PR #925, opened 2020-09-22
Adding debug decorator to print Numba compile stats

kozlov-alexey:
This adds a decorator to print Numba compilation stats for a decorated function
and all nested functions (literally all compiled overloads). Filtering
can be made using decorator arguments.

NOTE: this may show a bit lesser times than compile time calculated using
common approach via first_run_exec_time - second_run_exec_time.

### DIFF
--- a/sdc/decorators.py
+++ b/sdc/decorators.py
@@ -31,6 +31,10 @@ This is a function decorator definition
 import numba
 import sdc
 
+from functools import wraps
+from numba.core.registry import CPUDispatcher
+from sdc.utilities.utils import print_compile_times
+
 
 def jit(signature_or_function=None, **options):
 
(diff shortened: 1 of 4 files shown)


kozlov-alexey:
Example output (if level=0 arg is used):

> Function: csv_reader_py$4
>     Pipeline: nopython
>     Time: 0.7950319000000032
> 
> Function: sdc_pandas_dataframe_drop_impl$34
>     Pipeline: nopython
>     Time: 2.902112299999988
> ...

kozlov-alexey:
@Hardcode84 Do you think we can merge this? Some builds are failed due to new pyarrow version, which needs investigation, but this PR is not changing main functionality only adds debug possibilities and is not related to these issues.

Hardcode84:
Is this critical for us at  the moment? I would prefer it to have at least some tests (e.g. run some test in separate process and match output format with regex)

kozlov-alexey:
> run some test in separate process and match output format with regex

@Hardcode84 I added some tests as you have suggested (haven't done that with separate process though - it seems easier to do this in the same test process).

Hardcode84 (approved):

PR #926, opened 2020-09-23
Migrating to Numba 0.51.2

kozlov-alexey:
Adds prefer_literal flag to sdc overload decorators with default value `True` to satisfy previous Numba behavior.

Following problems were solved during transition:
- Added type-checks to functions decorated with intrinsic that require literal arguments (e.g. getitem accessors) to handle new Numba behavior to prefer_literal=False default typing (introduced in https://github.com/numba/numba/pull/6058/)
- Fixed incorrect type-checks in `fillna` and `astype` overloads to avoid returning wrong function implementation when compiling with non-literal args first.


Hardcode84 (approved):
Write in description why these changes were needed, LGTM otherwise

kozlov-alexey:
> Write in description why these changes were needed, LGTM otherwise

Done.

PR #918, opened 2020-08-30
Changing csv_reader_py impl to return df from objmode

kozlov-alexey:
Motivation: returning Tuple of columns read from csv file with
pyarrow csv reader from objmode and further calling init_dataframe
ctor to create native DF turned out to be inefficient in sense of
LLVM IR size and compilation time. With this PR we now rely on DF
unboxing and return py DF from objmode.

**Compile time of read_csv + df.count():**
solutions\columns | 4 | 8 | 16 | 32 | 64 | 128 | 256
-- | -- | -- | -- | -- | -- | -- | --
Numba master + both SDC fixes   (2b8b0034d74) | 8.897234 | 9.306839 | 10.54691 | 12.52175 | 17.41399 | 30.47878 | 65.63396
Numba master + SDC fix #1 (964e498a9) | 9.283413 | 9.83861 | 13.30219 | 21.7165 | 53.07618 | 187.4615 | 1026.31
Numba 0.50.1  + SDC master | 9.212505 | 10.238 | 14.08183 | 25.16768 | 72.9872 | 290.3359 | 2141.832
Ratio (both fixes to master) | 1.035435 | 1.100051 | 1.335162 | 2.009917 | 4.191296 | 9.525835 | 32.63299

PR #932, opened 2020-09-30
Revert "Changing csv_reader_py impl to return df from objmode"

Hardcode84:
Reverts IntelPython/sdc#918

Due to unstable tests on Windows


PokhodenkoSA (approved):

kozlov-alexey (commented):
@PokhodenkoSA Why revert is needed and why not just skip affected tests and create a JIRA issue?
Has the problem with tests been seen locally or just in particular env?

PR #949, opened 2020-11-23
Refactor df.geitem by slice idx to improve compile time

kozlov-alexey:
Motivation: df[idx] when idx is a slice produces DataFrame with the same
internal structure as original one (only index type may change). Hence
it can be copied and filled with column[idx] slices, which improves
compilation time.


kozlov-alexey:
Perf tests have to be executed (or added if needed) to ensure no degradation in execution time was introduced.

AlexanderKalistratov (approved):

PR #951, opened 2020-12-31
Add PositionalIndex and EmptyIndex types and align indexes API

kozlov-alexey:
Motivation: removing types.None as representation of default pd.RangeIndex
created when index=None is used in DF and Series ctors, that requires
most of Series and DF implementations to branch basing on index types.
Instead all Series and DF functions should use common indexes API
aligned to pandas one.

Changed in this PR:
- types.None index is removed;
- Added EmptyIndexType (instead types.None index) to represent empty
pandas index with dtype='Object' i.e. pd.Index([]);
- Added PositionalIndexType as a replacement for types.None index for
non-empty DFs and Series;
- Changed unboxing of RangeIndex objects to either RangeIndexType of
PositionalIndexType depending on whether it's default range or not;
- Changed fix_df_index and dataframe_constructor.py to create
EmptyIndexType or other index depending on whether tuple of columns is empty;
- Moved implementations for index types from common and
numpy_like functions to specific index files;
- Updated operators and other Series methods implementations to avoid
branching on index types and used index objects API instead;
- Reorganized index tests and added tests verifying specific function
(e.g. series reindexing) for all types of indexes.


kozlov-alexey:
@AlexanderKalistratov Can you please approve so we can finally merge this?

AlexanderKalistratov (approved):

PR #954, opened 2021-01-20
Update wheels build + Update linux build tag for wheels

Vyacheslav-Smirnov:
- Update wheels build procedure - align with other packages; current build via conda output does not support required additional options
 - Use manylinux1 for Python 3.6
 - Use manylinux2014 for Python > 3.6


Vyacheslav-Smirnov:
Mac build fails on gold/2021 branch also. I do not think that it is caused by current changes.
So I propose to merge this PR.

conda-recipe/meta.yaml
```diff
@@ -13,15 +13,12 @@ build:
...
     - SDC_CONFIG_PIPELINE_SDC
+    - WHEELS_OUTPUT_FOLDER
 
 requirements:
   build:
     - {{ compiler('c') }}      # [not osx]
     - {{ compiler('cxx') }}    # [not osx]
-    - wheel
-    - python
-    - numba   {{ NUMBA_VERSION }}
```
kozlov-alexey, line 23: Numba recently became a build dependency for sdc, so it's better to keep it.

conda-recipe/meta.yaml
```diff
@@ -45,24 +42,6 @@ test:
   imports:
     - sdc
 
-outputs:
```
kozlov-alexey, line 48: Why to remove 'outputs' section at all?

conda-recipe/meta.yaml
Vyacheslav-Smirnov: Ok, let's keep numba here

conda-recipe/meta.yaml
Vyacheslav-Smirnov: One of the purpose of this PR is to align wheels build procedure with others packages.
Here was used conda output - it does not suits for us due to we cannot set additional arguments for wheels build.
That is why output section is removed and wheel build lines are added in build.sh and bld.bat

PR #961, opened 2021-02-17
Fixing bug in stability of mergesort impl for StringArray

kozlov-alexey:
Motivation: for StringArray type legacy implementation of stable sort
computed result when sorting with ascending=False by reversing the
result of argsorting with ascending=True, which produces wrong order in
groups of elements with the same value. Implemented solution adds
new function argument 'ascening' and uses it when calling native function
impl via serial stable_sort.


AlexanderKalistratov (approved):

kozlov-alexey:
As an exceptional case agreed to merge and include in the next release despite internal CI fails with python3.6.

PR #970, opened 2021-04-02
Update compiler dialect flags to use C++17

kozlov-alexey:
Motivation: need C++17 features in native extensions.


setup.py
```diff
@@ -111,8 +111,13 @@ def check_file_at_path(path2file):
...
 ind = [PREFIX_DIR + '/include', ]
 lid = [PREFIX_DIR + '/lib', ]
-eca = ['-std=c++11', "-O3", "-DTBB_PREVIEW_WAITING_FOR_WORKERS=1"]  # '-g', '-O0']
-ela = ['-std=c++11', ]
+
+if is_win:
+    eca = ['/std:c++17', "/Ox", "/DTBB_PREVIEW_WAITING_FOR_WORKERS=1", ]  # "/Zi", "/Od", "/DEBUG:FULL"]
+    ela = []  # '/DEBUG:FULL', ]
+else:
+    eca = ['-std=c++17', "-O3", "-DTBB_PREVIEW_WAITING_FOR_WORKERS=1", ]  # '-g', '-O0']
```
AndresGuzman-Ballen, line 119: If you make the following change, it should work:
```
else:
    eca = ['-std=c++17', "-O3", "-DTBB_PREVIEW_WAITING_FOR_WORKERS=1", "-fno-aligned-allocation"]  # '-g', '-O0']
```
If this is only a problem with macOS, it'd be a good idea to do something like this:
```
else:
    eca = ['-std=c++17', "-O3", "-DTBB_PREVIEW_WAITING_FOR_WORKERS=1", ]  # '-g', '-O0']

    # On macOS, c++17 flag is ignored unless this flag is also passed to distutils
    if is_osx:
        eca += ["-fno-aligned-allocation"]
```

kozlov-alexey: @AndresGuzman-Ballen Thank you Andres, it helped!

AndresGuzman-Ballen: Woohoo!

PR #971, opened 2021-04-12
Moving to numba=0.53.1

kozlov-alexey:
Motivation: keep up with latest Numba release.

Note: commit 2017e6c28d is actually just a workaround for numba=0.53.1 regressions and probably should be reverted once referred issues are fixed.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set NUMBA_VERSION = "==0.52.0" %}
+{% set NUMBA_VERSION = "==0.53.0" %}
 {% set PANDAS_VERSION = "==1.2.0" %}
 {% set PYARROW_VERSION = "==2.0.0" %}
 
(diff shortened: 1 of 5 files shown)


kozlov-alexey:
Regarding possible performance impact for numpy_like.astype() from numeric arrays to `StringArrayType`. There's actually no impact at all, since for some reason iterating over StringArrayType with prange doesn't scale (should be investigated), so testing conversion from 5 * 10 ** 6 of int64 to string, SDC code is 1.5x times faster than pandas, but doesn't scale:

n_threads | 1 | 2 | 4 | 8 | 16
-- | -- | -- | -- | -- | --
tested | 1.769644 | 1.77148 | 1.768676 | 1.768381 | 1.769647
reference | 2.677486 | 2.624062 | 2.624959 | 2.622924 | 2.625067
ratio | 1.513008 | 1.481282 | 1.484138 | 1.483235 | 1.483385

AlexanderKalistratov (approved):

PR #972, opened 2021-04-12
Initial version of ConcurrentDict container via TBB hashmap

kozlov-alexey:
Motivation: SDC relies on typed.Dict implementation in many core
pandas algorithms, and it doesn't support concurrent read/writes.
To fill this gap we add ConcurrentDict type which will be used if
threading layer is TBB.

### DIFF
--- a/sdc/extensions/sdc_hashmap_ext.py
+++ b/sdc/extensions/sdc_hashmap_ext.py
@@ -1,0 +1,1113 @@
+# *****************************************************************************
+# Copyright (c) 2019-2021, Intel Corporation All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     Redistributions of source code must retain the above copyright notice,
+#     this list of conditions and the following disclaimer.
+#
+#     Redistributions in binary form must reproduce the above copyright notice,
+#     this list of conditions and the following disclaimer in the documentation
+#     and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+# THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+# OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+# OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+# EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+import llvmlite.binding as ll
+import llvmlite.llvmpy.core as lc
+import numba
+import numpy as np
+import operator
+import sdc
+
+from sdc import hstr_ext
+from glob import glob
+from llvmlite import ir as lir
+from numba import types, cfunc
+from numba.core import cgutils
+from numba.extending import (typeof_impl, type_callable, models, register_model, NativeValue,
+                             lower_builtin, box, unbox, lower_getattr, intrinsic,
+                             overload_method, overload, overload_attribute)
+from numba.cpython.hashing import _Py_hash_t
+from numba.core.imputils import (impl_ret_new_ref, impl_ret_borrowed, iternext_impl, RefType)
+from numba.cpython.listobj import ListInstance
+from numba.core.typing.templates import (infer_global, AbstractTemplate, infer,
+                                         signature, AttributeTemplate, infer_getattr, bound_function)
+from numba import prange
+
+from sdc.str_ext import string_type
+from sdc.str_arr_type import (StringArray, string_array_type, StringArrayType,
+                              StringArrayPayloadType, str_arr_payload_type, StringArrayIterator,
+                              is_str_arr_typ, offset_typ, data_ctypes_type, offset_ctypes_type)
+from sdc.utilities.sdc_typing_utils import check_is_array_of_dtype
+
+from numba.typed.typedobjectutils import _as_bytes
+from sdc import hconc_dict
+from sdc.extensions.sdc_hashmap_type import (ConcurrentDict, ConcurrentDictType,
+                                             ConcDictKeysIterableType, ConcDictIteratorType,
+                                             ConcDictItemsIterableType, ConcDictValuesIterableType)
+from numba.extending import register_jitable
+
+from sdc.extensions.sdc_hashmap_type import SdcTypeRef
+from sdc.utilities.sdc_typing_utils import TypingError, TypeChecker, check_types_comparable
+from itertools import product
+
+from numba.typed.dictobject import _cast
+
+
+def gen_func_suffixes():
+    key_suffixes = ['int32_t', 'int64_t', 'voidptr']
+    val_suffixes = ['int32_t', 'int64_t', 'float', 'double', 'voidptr']
+    return map(lambda x: f'{x[0]}_to_{x[1]}',
+               product(key_suffixes, val_suffixes))
+
+
+def load_native_func(fname, module, skip_check=None):
+    for suffix in gen_func_suffixes():
+        if skip_check and skip_check(suffix):
+            continue
+        full_func_name = f'{fname}_{suffix}'
+        ll.add_symbol(full_func_name,
+                      getattr(module, full_func_name))
+
+
+load_native_func('hashmap_create', hconc_dict)
+load_native_func('hashmap_size', hconc_dict)
+load_native_func('hashmap_set', hconc_dict)
+load_native_func('hashmap_contains', hconc_dict)
+load_native_func('hashmap_lookup', hconc_dict)
+load_native_func('hashmap_clear', hconc_dict)
+load_native_func('hashmap_pop', hconc_dict)
+load_native_func('hashmap_update', hconc_dict)
+load_native_func('hashmap_dump', hconc_dict)
+load_native_func('hashmap_create_from_data', hconc_dict, lambda x: 'voidptr' in x)
+load_native_func('hashmap_getiter', hconc_dict)
+load_native_func('hashmap_iternext', hconc_dict)
+
+
+supported_numeric_key_types = [
+    types.int32,
+    types.uint32,
+    types.int64,
+    types.uint64
+]
+
+
+supported_numeric_value_types = [
+    types.int32,
+    types.uint32,
+    types.int64,
+    types.uint64,
+    types.float32,
+    types.float64,
+]
+
+
+# to avoid over-specialization native hashmap structs always use signed integers
+# and function arguments and return values are converted when needed
+reduced_type_map = {
+    types.int32: types.int32,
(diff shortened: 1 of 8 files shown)


sdc/native/conc_dict_module.cpp
```diff
@@ -0,0 +1,659 @@
...
+declare_hashmap_iternext(void*, void*, voidptr_to_voidptr);
+
+
+#define REGISTER(func) PyObject_SetAttrString(m, #func, PyLong_FromVoidPtr((void*)(&func)));
+
+PyMODINIT_FUNC PyInit_hconc_dict()
+{
+    static struct PyModuleDef moduledef = {
+        PyModuleDef_HEAD_INIT,
+        "htbb_hashmap",
```
Hardcode84, line 386: htbb?

sdc/native/hashmap.hpp
```diff
@@ -0,0 +1,907 @@
...
+template<typename Key,
+         typename Val,
+         typename Hasher=std::hash<Key>,
+         typename Equality=std::equal_to<Key>
+>
+class NumericHashmapType {
+public:
+    using map_type = typename tbb::concurrent_unordered_map<Key, Val, Hasher, Equality>;
+    using iterator_type = typename map_type::iterator;
+    std::unique_ptr<map_type> p_map;
```
Hardcode84, line 133: Why to store it as pointer, not as value?

sdc/native/hashmap.hpp
```diff
@@ -0,0 +1,907 @@
...
+
+    return found;
+}
+
+
+void GenericHashmapType<void*, void*, VoidPtrHashCompare>::set(void* key, void* val)
+
+{
+    auto ksize = this->key_info.size;
+    void* _key = malloc(ksize);
```
Hardcode84, line 679: Why not new? Can we wrap these explicit mallocs/frees into `unique_ptr`?

sdc/native/hashmap.hpp
```diff
@@ -0,0 +1,907 @@
...
+         typename HashCompare
+>
+int8_t GenericHashmapType<Key, void*, HashCompare>::pop(Key key, void* res) {
+    bool found = false;
+    {
+        typename map_type::const_accessor result;
+        found = this->p_map->find(result, key);
+        if (found)
+        {
+            memcpy(res, result->second, this->val_info.size);
```
Hardcode84, line 607: Clokwork will aggro on `memcpy`

sdc/extensions/sdc_hashmap_ext.py
```diff
@@ -0,0 +1,1123 @@
...
+    return map(lambda x: f'{x[0]}_to_{x[1]}',
+               product(key_suffixes, val_suffixes))
+
+
+def load_native_func(fname, module, skip_check=None):
+    for suffix in gen_func_suffixes():
+        if skip_check and skip_check(suffix):
+            continue
+        full_func_name = f'{fname}_{suffix}'
+        ll.add_symbol(full_func_name,
```
AlexanderKalistratov, line 81: Is there a reason, we are using `ll.add_symbol` and not ctypes binding, so we can directly call functions from jit code?

Like it is done here:
https://github.com/IntelPython/sdc/blob/master/sdc/functions/sort.py#L55

sdc/extensions/sdc_hashmap_ext.py
```diff
@@ -0,0 +1,1123 @@
...
+        # Note: this function raises exception so expect no scaling if you use it in prange
+        if not found:
+            raise KeyError("ConcurrentDict key not found")
+        return res
+
+    return concurrent_dict_lookup_impl
+
+
+@intrinsic
+def hashmap_clear(typingctx, dict_type):
```
AlexanderKalistratov, line 584: why do we need this to be as intrinsic and not as an overload?

```python
@overload(hashmap_clear)
def hashmap_clear_ovld(conc_dict):
    ty_key, ty_val = dict_type.key_type, dict_type.value_type
    key_type_postfix, value_type_postfix = _get_types_postfixes(ty_key, ty_val)

    hasmap_clear_func = hashmap_clear_map[(key_type_postfix, value_type_postfix)]
    
    def hashmap_clear_impl(conc_dict):
        return hashmap_clear_func(conc_dict._data_ptr)

    return hashmap_clear_impl
```

Am I missing something?

ctypes binding is required for this

sdc/native/hashmap.hpp
```diff
@@ -0,0 +1,907 @@
...
+    uint64_t size() {
+        return this->p_map->size();
+    }
+
+    int8_t contains(Key key) {
+        bool found = false;
+        {
+            typename map_type::const_accessor result;
+            found = this->p_map->find(result, key);
+            result.release();
```
AlexanderKalistratov, line 228: Do we really need both - inner scope and call to `release`?

sdc/native/hashmap.hpp
kozlov-alexey: Done.

sdc/native/hashmap.hpp
kozlov-alexey: It's problematic. I've tried a bit, but using `unique_ptr<char[ksize]>` is not possible since ksize is not compile-time const. Also the fact that TBB hashmap has to store the raw pointer (since HashCompare is actually numba functions that expect key/value to be pointers to Numba structs) kind of decrease the value of it (i.e. the ownership of allocated memory should go to hashmap itself and this function frees this memory only if insertion failed).

sdc/native/conc_dict_module.cpp
```diff
@@ -0,0 +1,659 @@
...
+{ \
+    return hashmap_iternext<key_type, val_type>(p_iter_state, ret_key, ret_val); \
+} \
+
+
+extern "C"
+{
+
+// empty hashmap creation functions
+declare_hashmap_create(int32_t, int32_t, int32_t_to_int32_t);
```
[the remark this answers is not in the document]
kozlov-alexey: Fixed.

sdc/native/conc_dict_module.cpp
```diff
@@ -0,0 +1,659 @@
...
+        NULL,
+    };
+    PyObject* m = PyModule_Create(&moduledef);
+    if (m == NULL)
+    {
+        return NULL;
+    }
+
+    // hashmap creation functions
+    REGISTER(hashmap_create_int32_t_to_int32_t)
```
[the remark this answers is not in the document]
kozlov-alexey: Ditto.

sdc/native/hashmap.hpp
kozlov-alexey: I double checked and it appears it doesn't (see report below - all three issues are known and should be fixed by #967). Not sure if `memcpy_s` will enhance something, but I can still change if you think it's better.

```
Klocwork 20.1.0.97
-----
May 4, 2021 2:01:17 PM CDT: Running build preparation stage...
May 4, 2021 2:01:17 PM CDT: Build preparation stage completed
May 4, 2021 2:01:18 PM CDT: Running compilation stage...
1> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/_distributed.cpp
2> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/_hiframes.cpp
9> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/native/sort.cpp
8> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/transport/hpat_transport_single_process.cpp
7> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/io/_io.cpp
6> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/_datetime_ext.cpp
4> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/_str_ext.cpp
3> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/_set_ext.cpp
Warning: /localdisk/work/akozlov/miniconda3/kwrun_6/.kwlp/buildspec.txt:24: 'compile' line for object file /localdisk/work/akozlov/miniconda3/hpat/build/temp.linux-x86_64-3.8/sdc/native/utils.o is already defined (ignored)
15> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/native/utils.cpp
14> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/native/module.cpp
11> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/native/stable_sort.cpp
5> CLASSIC: Compiling /localdisk/work/akozlov/miniconda3/hpat/sdc/native/conc_dict_module.cpp
May 4, 2021 2:01:29 PM CDT: Compilation stage completed
May 4, 2021 2:01:29 PM CDT: Running C/C++ defects detection stage...
1> Analyzing _distributed.cpp_0.o
2> Analyzing _hiframes.cpp_1.o
3> Analyzing _set_ext.cpp_2.o
4> Analyzing _str_ext.cpp_3.o
5> Analyzing _datetime_ext.cpp_4.o
6> Analyzing _io.cpp_5.o
7> Analyzing hpat_transport_single_process.cpp_6.o
8> Analyzing utils.cpp_10.o
9> Analyzing sort.cpp_7.o
10> Analyzing stable_sort.cpp_8.o
11> Analyzing module.cpp_9.o
12> Analyzing conc_dict_module.cpp_12.o
May 4, 2021 2:02:21 PM CDT: C/C++ defects detection stage completed
May 4, 2021 2:02:21 PM CDT: Running linking stage...
May 4, 2021 2:02:22 PM CDT: Linking stage completed
1 (Local) /localdisk/work/akozlov/miniconda3/hpat/sdc/native/sort.cpp:151 VOIDRET (2:Error) Analyze
void function returns value

2 (Local) /localdisk/work/akozlov/miniconda3/hpat/sdc/native/utils.cpp:89 NPD.CHECK.MUST (1:Critical) Analyze
Pointer 'this->arena._M_t' checked for NULL at line 86 will be dereferenced at line 89.

3 (Local) /localdisk/work/akozlov/miniconda3/hpat/sdc/native/stable_sort.cpp:340 VOIDRET (2:Error) Analyze
void function returns value

Summary: 3 Local
3 Total Issue(s)
```

sdc/native/hashmap.hpp
kozlov-alexey: No of course, but it's (I guess) usual practice to release immediately after you don't need the lock (i.e. I copy-pasted it from TBB examples :)).

sdc/extensions/sdc_hashmap_ext.py
kozlov-alexey: No, there's no particular reason for this (except it was easier to write LLVM codegen that way - the function call inserted there directly refers to the function name being loaded, so there's 1:1 mapping and it's easier to navigate the code). I think we may refactor this and re-use common ctypes/bind approach later.

sdc/extensions/sdc_hashmap_ext.py
kozlov-alexey: Thanks! I didn't consider this way of doing things. However it may be more difficult to read for methods like `set` and `pop` which make key transformations during codegen or allocate the result (e.g. https://github.com/IntelPython/sdc/blob/11774955fb3ce3237d89bc59c3522b99d6c61e55/sdc/extensions/sdc_hashmap_ext.py#L434), since the logic of generated llvm code will be split across functions (and also it would be much harder to debug this if you need to make some changed in the codegen). So if you don't mind I will keep it initial way for now.

sdc/native/hashmap.hpp
AlexanderKalistratov: You can use unique_ptr with dynamically-sized arrays. Like this :
```C++
auto _key = std::unique_ptr<char[]>(new char[ksize])
```
Or like this:
```C++
auto _key = std::make_unique<char[]>(ksize); 
```

sdc/native/hashmap.hpp
```diff
@@ -0,0 +1,862 @@
...
+{
+    auto p_self_hash_map_spec = reinterpet_hashmap_ptr<key_type, val_type>(p_self_hash_map);
+    auto p_arg_hash_map_spec = reinterpet_hashmap_ptr<key_type, val_type>(p_arg_hash_map);
+    p_self_hash_map_spec->update(*p_arg_hash_map_spec);
+    return;
+}
+
+
+template<typename key_type, typename val_type>
+void hashmap_dump(void* p_hash_map)
```
Hardcode84, line 819: Can we make  `hashmap_dump` and corresponding `#include <iostream>` optional and wrap in some ifdefs? I would really prefer release libraries without iostream dependency?

Hardcode84 (approved):

sdc/native/hashmap.hpp
kozlov-alexey: Done.

PR #974, opened 2021-05-05
Fixes pack of failed tests

kozlov-alexey:
Motivation: as title.

### DIFF
--- a/sdc/datatypes/categorical/functions.py
+++ b/sdc/datatypes/categorical/functions.py
@@ -24,9 +24,11 @@
 # EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 # *****************************************************************************
 
-from sdc.utilities.utils import sdc_overload_attribute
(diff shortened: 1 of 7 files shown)


Hardcode84:
> Skipping floordiv test due to diff behavior in Stock vs Intel numpy

This is probably a real bug either on our side or on intel Numpy/mkl side. Better to report this to mkl team

AlexanderKalistratov (approved):

kozlov-alexey:
> > Skipping floordiv test due to diff behavior in Stock vs Intel numpy
> 
> This is probably a real bug either on our side or on intel Numpy/mkl side. Better to report this to mkl team

I created SAT-4111 for tracking it. Also it seems like these unexpected  successes of this test in azure pipeline were caused by move from numpy=1.19 to numpy=1.20, where (see https://github.com/numpy/numpy/pull/16161) floordiv behavior was changed.

kozlov-alexey:
Merging this to unblock CI (fails in python 3.8 checks are investigated by Infra team).

PR #975, opened 2021-05-13
upd numba 0.53.1

ekomarova:


kozlov-alexey:
@ekomarova Thanks for the PR, but these changes are not enough to make SDC work with numba=0.53.1 (see #971 which is already opened). Also in general we do not need PRs to gold/2021 directly, since we test and merge to gold/2021 only what has been tested and integrated into master branch (@xaleryb can give you more details on this).

PR #977, opened 2021-05-17
Dependency for numpy 1.17

xaleryb:
This PR is to follow Anaconda BKM to build using the oldest numpy with a stable API, which is 1.16. Intel's oldest numpy supporting python 3.7 is numpy 1.17, so we use that only at build time so ensure broadest compatibility with newer numpy versions.

Relates to https://github.com/tensorflow/tensorflow/issues/47691  (TensorFlow does not support numpy 1.20.x)

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - python
     - setuptools
     - numba   {{ NUMBA_VERSION }}
-    - numpy
+    - numpy 1.17
     - pandas  {{ PANDAS_VERSION }}
     - pyarrow {{ PYARROW_VERSION }}
     - wheel
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -25,7 +25,8 @@ requirements:
     - python
     - setuptools
     - numba   {{ NUMBA_VERSION }}
-    - numpy 1.17
+    - numpy                           # [win or osx or py==38]
+    - numpy 1.17                      # [linux and py==37]
     - pandas  {{ PANDAS_VERSION }}
     - pyarrow {{ PYARROW_VERSION }}
     - wheel
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set NUMBA_VERSION = "==0.53.1" %}
-{% set PANDAS_VERSION = "==1.2.0" %}
+{% set PANDAS_VERSION = "==1.2.4" %}
 {% set PYARROW_VERSION = "==2.0.0" %}
 
 package:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set NUMBA_VERSION = "==0.53.1" %}
-{% set PANDAS_VERSION = "==1.2.4" %}
+{% set PANDAS_VERSION = "==1.2.0" %}
 {% set PYARROW_VERSION = "==2.0.0" %}
 
 package:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -34,7 +34,7 @@ requirements:
 
   run:
     - python
-    - numpy
+    - {{ pin_compatible('numpy') }}
     - pandas  {{ PANDAS_VERSION }}
     - pyarrow {{ PYARROW_VERSION }}
     - numba   {{ NUMBA_VERSION }}


kozlov-alexey:
Hi, @xaleryb, is this only for testing purposes? What is the rationale for this? I think we should not restrict numpy version for SDC at all, since by itself it doesn't need specific numpy.

xaleryb:
hi @kozlov-alexey  this is not for testing only we plan to merge and include to 2021.3 release. This is not a restriction because we do not set run dependency for numpy 1.17 (this is inly for building) and we expect sdc build (based on numpy 1.17) working with latest numpy which will be a part of 2021.3. This is for compatibility on AIKIt level

kozlov-alexey:
> this is not for testing only we plan to merge and include to 2021.3 release. 
Then this has to merged to development branch first, as per our agreements.

But:
> ... we expect sdc build (based on numpy 1.17) working with latest numpy which will be a part of 2021.3. This is for compatibility on AIKIt level
this has impact not only on AIKit but anyone who will build SDC with conda. There should probably be other way to restrict numpy version that is used when building SDC for AIKit, shouldn't it?

conda-recipe/meta.yaml
```diff
@@ -25,7 +25,8 @@ requirements:
     - python
     - setuptools
     - numba   {{ NUMBA_VERSION }}
-    - numpy
+    - numpy                           # [win or osx or py==38]
```
kozlov-alexey, line 28: @tomashek, @xaleryb I would prefer not to submit this as there's no requirement for SDC package itself to be built with specific numpy version. Also there are no description in the commit that explains why this change is needed and when it can be reverted (if it's some kind of workaround, at least some GitHub reference is needed). Also there should generally be more proper ways of fixing numpy version during specific conda builds (e.g. CONDA_NPY env variable may be set).

kozlov-alexey (changes requested):
@tomashek, @xaleryb I would prefer not to submit this as there's no requirement for SDC package itself to be built with specific numpy version. Also there are no description in the commit that explains why this change is needed and when it can be reverted (if it's some kind of workaround, at least some GitHub reference is needed). Also there should generally be more proper ways of fixing numpy version during specific conda builds (e.g. CONDA_NPY env variable may be set).

conda-recipe/meta.yaml
```diff
@@ -25,7 +25,8 @@ requirements:
...
+    - numpy 1.17                      # [linux and py==37]
```
kozlov-alexey, line 29: This adds constraints on the runtime numpy version too, so generally it's needed to add restrictions in run section as well (see https://numpy.org/devdocs/user/depending_on_numpy.html)

kozlov-alexey:
> _No description provided._

Please add description with motivation for this change and add the reference to related Github/Jira issues.

conda-recipe/meta.yaml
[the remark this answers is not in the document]
xaleryb: The initial problem we are trying to resolve by this change is applicable only for linux.

conda-recipe/meta.yaml
xaleryb: I don't see any recommendation to set restrictions in run area if we need to restrict something only for build time. Please point me to right sentence in provided article. More over if we'll set restrictions in run area as well this package will not work in IDP because will need to use numpy 1.17 which was not planned to be a part of IDP 2021.3. This information is a part of Todd's answer in email.

xaleryb:
> > _No description provided._
> 
> Please add description with motivation for this change and add the reference to related Github/Jira issues.

@tomashek could you please share this information if any. 
We need to merge this change ASAP because a lot of work we need to perform right after this to check our release is good to go.

tomashek:
> > _No description provided._
> 
> Please add description with motivation for this change and add the reference to related Github/Jira issues.

@kozlov-alexey , I do not think there is an Issue in this github to reference.

conda-recipe/meta.yaml
kozlov-alexey: No, not == 1.17 in the run section, but conda should know that it has to use compatible version at runrime, see the following sentence: "For conda-forge it’s a little less complicated: there’s dedicated handling for NumPy in build-time and runtime dependencies, so typically this is enough (see here for docs)" and example afterwards.

kozlov-alexey: i.e. change:
``` diff
@@ -34,7 +34,7 @@ requirements:
 
   run:
     - python
-    - numpy
+    - {{ pin_compatible('numpy') }}
     - pandas  {{ PANDAS_VERSION }}
     - pyarrow {{ PYARROW_VERSION }}
     - numba   {{ NUMBA_VERSION }}

```

tomashek: @kozlov-alexey , that should work, as I believe building with 1.17.5 will result in a runtime dependency of:
numpy >=1.17.5,<2.0a0

PR #982, opened 2021-07-28
Migrating to pyarrow=4.0.1

kozlov-alexey:
Motivation: keep up with the latest versions of dependencies

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,7 +85,7 @@ Building on Linux with setuptools
 
     export PYVER=<3.6 or 3.7>
     export NUMPYVER=<1.16 or 1.17>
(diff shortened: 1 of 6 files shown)


kozlov-alexey:
The change in `pandas_read_csv` is adaptation for https://github.com/apache/arrow/pull/9203 that changed date-inferred type from TimestampType to DateType and caused `sdc.tests.test_io.TestCSV.test_pyarrow (test='csv_date1')` to fail since `to_pandas` converts it back to strings (unless `date_as_object=False` is used).

AlexanderKalistratov (approved):

PR #986, opened 2021-10-14
Merge from gold/2021 to dev branch (excluding recipe changes)

kozlov-alexey:
As title.


kozlov-alexey:
Merging despite issues with internal CI on Linux (should be resolved afterwards).

PR #991, opened 2021-11-02
add numpy 1.19

ekomarova:
Motivation: compatibility with external packages.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -25,8 +25,7 @@ requirements:
     - python
     - setuptools
     - numba   {{ NUMBA_VERSION }}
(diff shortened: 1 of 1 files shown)


ekomarova:
@xaleryb, @kozlov-alexey please review (status checks are not reported to this PR, but they were successful for all pythons)

kozlov-alexey (commented):
@ekomarova, @xaleryb I see following numpy is used during building SDC in 'passed' checks, e.g. on Linux + python 3.8:
numpy:                                1.21.2-py39hec4e512_1      http://satpkgserver.an.intel.com:8081
Similarly as on dev branch where CI checks also seems to pass, so a question: do we need this update at all?

kozlov-alexey (approved):

PR #998, opened 2021-12-03
Allow global values to be used in read_csv params (names, usecols)

kozlov-alexey:
Motivation: fixes a bug in RewriteReadCsv rewrite that fails compilation
when exception is raised during the match phase because variable
defining value of read_csv parameter is not ir.Expr node but ir.Global
(and thus has no op attribute).


PokhodenkoSA (commented):
@kozlov-alexey
Please, do not add me to review. I didn't develop anything for SDC for a long time now.

AlexanderKalistratov (approved):

